PR #30, opened 2017-12-30
update readme and do some usability tweaks

jeffmcaffer:


README.md
```diff
@@ -10,18 +14,134 @@ production data.
...
+```
+
+You can also get the curation for a particular component revision using one of the following requests. Both return the full curation for the given component. The first (without the `pr` segment), gets the current curation that is in effect -- the content of the `master` branch. The second gets the curation proposed in the given pull request.
+
+```
+GET http://localhost:4000/curations/npm/npmjs/-/redie/0.3.0
+GET http://localhost:4000/curations/npm/npmjs/-/redie/0.3.0/pr/23
+```
+
+Both of these whill
```
mynameistechno: Incomplete sentence?

PR #31, opened 2018-01-01
First stab at replacing Travis build with custom status check

mynameistechno:
This adds a Github webhook route that replaces and improves upon the curation Travis build. It listens to PR "pushes" (opened PRs, reopened PRs, and commits pushed to PRs) and updates the PR status with curation validation results.  Currently it just lints them by loading them with `js-yaml`. Next we can validate with a schema, etc.

Failed checks will look like this:
image

Successful checks will look like:
image
(the icon that appears is the avatar of the user that updates the status via the API). 

To install:
* In the `curated-data` repo settings, add a Webhook:
  * Payload URL should be something like `https://clearlydefined.io/webhook`
  * Content type: `application/json`
  * Select "Individual events" and select "Pull request"

To do:
* Add tests
* Validate yaml via schema
* Add better logging
* Add webhook secret
* Update URL that check links to

LMK if you think this needs it's own provider - for now I just added the Github client calls to `providers/curation/github.js`.


app.js
```diff
@@ -42,6 +44,8 @@ app.use(requestId());
 app.use(logger('dev'));
 app.use(configMiddleware);
 app.use('/api-docs', swaggerUi.serve, swaggerUi.setup(swaggerDoc));
+app.use(bodyParser.json());
```
jeffmcaffer: This _should_ be ok. The `bodyParser` was after the `/harvest` endpoint as that endpoint was dealing with raw data. That has since changed but that change itself needs to be validated with others from a design point of view.

routes/webhook.js
```diff
@@ -0,0 +1,75 @@
...
+
+function fatal(request, response, error) {
+  logger().error(error);
+  response.status(500);
+  response.setHeader('content-type', 'text/plain');
+  response.end(JSON.stringify(error));
+}
+
+// @todo add real logger
+function logger() {
```
jeffmcaffer: Yea, the logger should be injected in setup. We'll propbably use something like Winston as it has pluggable backends.

routes/webhook.js
```diff
@@ -0,0 +1,75 @@
...
+  return console;
+}
+
+// @todo add secret
```
jeffmcaffer: Is this still a TODO?

routes/webhook.js
```diff
@@ -0,0 +1,75 @@
...
+
+  const curationResults = await Promise.all(
+    curationFilenames.map(
+      path => curationService.getContent({ref, path})
+        .then(curationService.isValidCuration)
+    )
+  );
+  const invalidCurations = [];
+  curationResults.forEach((result, index) => {
+    if (!result) {
```
jeffmcaffer: FWIW, I have been NOT using {} for single like expressions like this. I'm open to group consensus but lets be consistent until a change is decided on

routes/webhook.js
```diff
@@ -0,0 +1,75 @@
...
+
+  const {pull_request: pr, action: prAction, number} = request.body;
+  const {sha, ref} = pr.head;
+  const isValidPullRequest = pr && validPrActions.includes(prAction);
+  if (!isValidPullRequest) {
+    return fatal(request, response, 'Not a valid Pull Request event');
+  }
+
+  const prFiles = await curationService.getPrFiles({number});
+  const curationFilenames = prFiles.map(
```
jeffmcaffer: Perhaps format like

```javascript
const curationFilenames = prFiles
  .map(x => x.filename)
  .filter(curationService.isCurationFile);
```

routes/webhook.js
```diff
@@ -0,0 +1,75 @@
...
+router.post('/', async (request, response, next) => {
+  const isGithubEvent = request.headers['x-github-event'];
+  if (!isGithubEvent) {
+    return fatal(request, response, 'Not a Github event');
+  }
+
+  // @todo secure webhook, see https://github.com/Microsoft/ghcrawler/blob/develop/routes/webhook.js#L28
+
+  const {pull_request: pr, action: prAction, number} = request.body;
+  const {sha, ref} = pr.head;
```
jeffmcaffer: formatting.  spaces around the {} so `{ sha, ref }` rather than `{sha, ref}`.  Here and elsewhere

We should setup linting rules. Actually we have .eslint setup but a) not sure when it is run and b) it does not appear to have all the rules.  Lets iterate on that.

providers/curation/github.js
```diff
@@ -131,6 +132,53 @@ class GitHubCurationService {
...
   }
 
+  async getContent({ref, path}) {
+    const {owner, repo} = this.options;
+    const github = Github.getClient(this.options);
+    try {
+      const response = await github.repos.getContent({owner, repo, ref, path});
+      return base64.decode(response.data.content);
+    } catch (error) {
+      // @todo add logger
```
jeffmcaffer: Great call out. Need to figure the right error handling strategy across the board. Interestingly for some of these REST requests, they only throw and error if there is an error in the operation (e.g., the network is down). Getting a 4** or 5** is still success from the request point of view. So there needs to be "error" handling in the non-error cases as well. Again, lets figure a general strategy and apply.

.nvmrc
```diff
@@ -0,0 +1 @@
+9.3.0
```
jeffmcaffer: Dont think we can go this high on the version. The dockerfile specs node 8 and 8.9.3 is LTS. Is there something we need from 9.*?

providers/curation/github.js
```diff
@@ -131,6 +132,53 @@ class GitHubCurationService {
...
+    const {owner, repo} = this.options;
+    const github = Github.getClient(this.options);
+    let files = [];
+    try {
+      const response = await github.pullRequests.getFiles({owner, repo, number});
+      files = response.data;
+    } catch (error) {
+      // @todo add logger
+    }
+    return files;
```
jeffmcaffer: we'd return files even in the event of an error?  Perhaps

```javascript
  try {
    return await github.pullRequests.getFiles({owner, repo, number}).data;
  } catch (error) {
    // @todo add logger
    throw error;
  }
```

Depending on the calling scenario, we may not even want to catch here. Rather catch and log up high.

.nvmrc
mynameistechno: I just upgraded to latest after it complained. I will try LTS.

app.js
mynameistechno: Ok, I'll leave as is for now, we can change later if needed.

providers/curation/github.js
mynameistechno: Yeah, let's keep it consistent across the app. Not sure what you guys have used in the past for this. Once we sort that out I'll update.

providers/curation/github.js
mynameistechno: Sounds good.

routes/webhook.js
mynameistechno: Yeah, I had Prettier enabled in my editor on-save, I ended up disabling it. We just started using Prettier as an eslint plugin at work to eliminate style/formatting nitpicks in PRs. We rather like it, but it's not for everyone. https://github.com/prettier/prettier

routes/webhook.js
mynameistechno: Yeah Prettier kicking in, but will update.

routes/webhook.js
mynameistechno: Yep, w/e everyone agrees to. I personally always use `{}`, but we should shoot for consistency 👍

routes/webhook.js
mynameistechno: Yeah, but will add it shortly.

mynameistechno:
> Overall this is great. Exactly what we need. I put a number of comments in. Several are formatting/style nits. Some are observations about future things. There are a couple that likely need consideration but overall as a first cut, this should be good to go.

Thanks for the feedback, I'll address the comments today.

providers/curation/github.js
mynameistechno: It didn't like returning .data directly from the async function. I had to first assign it to a var, then return `.data` after. But I made it throw the error 👍

PR #44, opened 2018-01-06
Secure the webhook

mynameistechno:
This secures the webhook. You need to set the env var `CURATION_GITHUB_WEBHOOK_SECRET` to the same value you set the `secret` in the webhook config in GH. 

In this PR I also update the status of the PR check to 'pending' prior to validation, similar to other GH integrations like Travis CI, etc. It displays as "Validation in progress".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,4 @@ typings/
 
 # dotenv environment variables file
 .env
+env.json
--- a/app.js
+++ b/app.js
@@ -34,7 +34,7 @@ const curations = require('./routes/curations')(curationService);
 const packages = require('./routes/packages')(harvestStore, summaryService, aggregatorService, curationService);
 
 const appLogger = console; // @todo add real logger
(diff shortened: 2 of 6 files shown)


app.js
```diff
@@ -56,6 +55,7 @@ app.use(basicAuth({
 
 app.use('/', index);
 app.use('/harvest', harvest);
+app.use(bodyParser.json());
```
mynameistechno: @jeffmcaffer to validate the payload in the webhook I needed the raw body so I put this back where it was before I moved it.

.gitignore
```diff
@@ -56,3 +56,4 @@ typings/
 
 # dotenv environment variables file
 .env
+env.json
```
jeffmcaffer: good call. for safety I always put the env.json file one level up.  that also enables sharing of the file between our different services.

app.js
jeffmcaffer: makes sense

PR #47, opened 2018-01-09
add curation, harvest and packages search endpoints

jeffmcaffer:
to make data more discoverable, we need to be able to search for curations and harvested data. This PR adds curation prefix accessing in a really simplistic way.

* The service keeps a clone (on demand) of the entire curation repo. This obviously does not scale but will do for a while. There is a configurable TTL for the clone after which a `git pull` is done before accessing.
* The file system is then used as the search space and all `.yaml` files under the given prefix are found, and their revisions teased out.

Eventually we will need something more than this. Perhaps a level based query, perhaps a real database, ... who knows. Oh, and this does not do pagination.

Having said all that, it works and we have precisely one curation in the dev repo right now :-)

For the harvest endpoint we already have `list` so just check that it is coherent with the new curation endpoint.

Then in `packages` we need to union the two results.

### DIFF
--- a/providers/curation/github.js
+++ b/providers/curation/github.js
@@ -206,7 +206,7 @@ class GitHubCurationService {
    * @param {string} searchPattern - a partial path that describes the sort of curation to look for. 
    * @returns {[URL]} - Array of URLs describing he available curations
    */
-  async search(searchPattern) {
+  async list(searchPattern) {
     await this.ensureCurations();
     const root = `${this.tempLocation.name}/${this.options.repo}/${this._getSearchRoot(searchPattern)}`;
     if (!fs.existsSync(root))
--- a/providers/curation/github.js
+++ b/providers/curation/github.js
@@ -4,9 +4,15 @@
 const _ = require('underscore');
 const base64 = require('base-64');
 const extend = require('extend');
+const { exec } = require('child_process');
+const fs = require('fs');
 const moment = require('moment');
+const readdirp = require('readdirp');
 const yaml = require('js-yaml');
 const Github = require('../../lib/github');
+const utils = require('../../lib/utils');
+const tmp = require('tmp');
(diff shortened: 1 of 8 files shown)


providers/curation/github.js
```diff
@@ -185,6 +201,50 @@ class GitHubCurationService {
...
+  /**
+   * Given a partial spec, return the list of full spec urls for each curated version of the spec'd components 
+   * @param {string} searchPattern - a partial path that describes the sort of curation to look for. 
+   * @returns {[URL]} - Array of URLs describing he available curations
+   */
+  async search(searchPattern) {
+    await this.ensureCurations();
+    const root = `${this.tempLocation.name}/${this.options.repo}/${this._getSearchRoot(searchPattern)}`;
+    if (!fs.existsSync(root))
+      return [];
```
geneh: Since it's an Express app, any sync functions should not be used  here because it would freeze the Node loop. fs.exists is deprecated, so something else should be used instead...
I would also add curly braces here to avoid any possible bugs in the future when another line is added etc.

providers/curation/github.js
```diff
@@ -185,6 +201,50 @@ class GitHubCurationService {
...
+      const result = [];
+      readdirp({ root, fileFilter: '*.yaml' })
+        .on('data', entry => result.push(...this.handleEntry(entry)))
+        .on('end', () => resolve(result))
+        .on('error', reject)
+    });
+  }
+
+  handleEntry(entry) {
+    const curation = yaml.safeLoad(fs.readFileSync(entry.fullPath.replace(/\\/g, '/')));
```
geneh: Same story here about readFileSync vs readFile

jeffmcaffer:
Ack on the sync reads. I was just being lazy. I'll fix before merging.

PR #52, opened 2018-01-18
First stab at validating curation against schema

mynameistechno:
Made some decent progress: 
* Turns out `js-yaml` doesn't validate yaml files against a JSON schema spec file out of the box, so I had to research a bit on best way to do this. Came across `ajv` which works well in conjunction with `js-yaml`.
* Setup a couple tests and fixture files (one for a valid curation, another for an invalid curation) so I had a nice feedback loop as I updated the curation schema. LMK if you guys prefer a different directory structure for the `test` dir.
* Started building up the curation schema using @jeffmcaffer's initial schema as a guide
* Played a bit with `revisions` and `patternProperties` (thanks @iamwillbar) and that seems to work well. Next we need to define a `revision` object in greater detail (i.e. add `described`, `licensed`, etc).

Although this is a WIP, it's mergeable - the tests pass. I'll try to add to the schema definition tonight or on Friday (I have some time blocked out for ClearlyDefined on Fridays). Once the schema is a bit more fleshed out, we can break it up into multiple files so that @iamwillbar can leverage them with Swagger.

### DIFF
--- a/lib/curation.js
+++ b/lib/curation.js
@@ -1,0 +1,44 @@
+// Copyright (c) 2017, The Linux Foundation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+const yaml = require('js-yaml');
+const Ajv = require('ajv');
+const ajv = new Ajv();
+ajv.addMetaSchema(require('ajv/lib/refs/json-schema-draft-06.json'));
+const curationSchema = require('../schemas/curation');
+
+class Curation {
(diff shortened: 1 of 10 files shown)


mynameistechno:
I fleshed out the schema some more. I also created a `Curation` class to encapsulate the logic and dara associated with loading and validating yaml content. This cleans up the calling code some and also makes it easier to write tests for 😛

PR #72, opened 2018-02-09
make team perms configurable and remove some checks

jeffmcaffer:
* make the perms teams configurable externally (env vars) to allow different environments (dev, prod, ...) to use different teams
* remove a couple of the permissions checks to make the system a little more open. Others will be added in as we add more function

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -69,8 +69,8 @@ module.exports = {
         team: 10 * 60, // 10 mins
       },
       permissions: {
-        harvest: ['harvesters'],
-        curate: ['curators'],
+        harvest: [config.get('AUTH_HARVEST_TEAM')].filter(e => e),
+        curate: [config.get('AUTH_CURATION_TEAM')].filter(e => e)
       }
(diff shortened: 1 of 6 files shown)


jeffmcaffer:
I'm going to proactively merge this but @jpeddicord, it would be great if you can take a quick look and see what you think.

lib/config.js
```diff
@@ -69,8 +69,8 @@ module.exports = {
         team: 10 * 60, // 10 mins
       },
       permissions: {
-        harvest: ['harvesters'],
-        curate: ['curators'],
+        harvest: [config.get('AUTH_HARVEST_TEAM')].filter(e => e),
```
jpeddicord: What's the filter accomplish here?

jeffmcaffer: removes nulls. if the env var is not set then `get` returns null and we end up with `[ null ]`. Elsewhere in the code we use length === 0 to indicate that no permissions checking is needed.

PR #73, opened 2018-02-09
support anonymous users with no token

jeffmcaffer:
This proposal separates anonymous, logged in and team users.
* anonymous -- they get a client on the req but no token. their requests will be ratelimited by GH. If we want, we can add our own token there...
* logged in -- users that have a valid GH token but are not members of any team. Their req will have a token but there will be certain operations they cannot do.
* team members -- valid GH token and they are member of a CD team. They can do various special operations.

### DIFF
--- a/middleware/github.js
+++ b/middleware/github.js
@@ -20,73 +20,42 @@ const options = {
  */
 module.exports = asyncMiddleware(async (req, res, next) => {
   const authHeader = req.get('authorization');
-  if (!authHeader || !authHeader.startsWith('Bearer ')) {
+  // if there is an auth header, it had better be Bearer
+  if (authHeader && !authHeader.startsWith('Bearer ')) {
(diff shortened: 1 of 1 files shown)


middleware/github.js
```diff
@@ -20,73 +20,42 @@ const options = {
...
   const authHeader = req.get('authorization');
-  if (!authHeader || !authHeader.startsWith('Bearer ')) {
+  // if there is an auth header, it had better be Bearer
+  if (authHeader && !authHeader.startsWith('Bearer ')) {
     res.status(401).end();
     return;
   }
-  const token = authHeader.split(' ')[1];
 
+  const token = authHeader ? authHeader.split(' ')[1] : null;
```
jpeddicord: May want to stick in a `req.hostname === 'localhost' || req.hostname === '127.0.0.1'`

PR #74, opened 2018-02-10
Add a /badges endpoint for displaying on your project site

tesmart:
This is a beginning. Perhaps later we can update the calculation logic to be a little more sophisticated.

Thanks to @jeffmcaffer @mynameistechno @cohenmax9

### DIFF
--- a/business/badgeCalculator.js
+++ b/business/badgeCalculator.js
@@ -5,12 +5,12 @@
 // For now we are just going to go with the simple:
 // If license/copyright there it gets a 3. Only one of those it gets a 2. None, it gets a 1
 
-const _ = require('lodash');
+const _ = require("lodash");
 
 const scoreToUrl = {
-  1: 'https://img.shields.io/badge/Clearly%20Defined%20Score-1-red.svg',
-  2: 'https://img.shields.io/badge/Clearly%20Defined%20Score-2-yellow.svg',
-  3: 'https://img.shields.io/badge/Clearly%20Defined%20Score-3-brightgreen.svg'
+  0: "https://img.shields.io/badge/ClearlyDefined%20Score-0-red.svg",
+  1: "https://img.shields.io/badge/ClearlyDefined%20Score-1-yellow.svg",
+  2: "https://img.shields.io/badge/ClearlyDefined%20Score-2-brightgreen.svg"
 };
 class BadgeCalculator {
   constructor(definition) {
@@ -22,13 +22,21 @@ class BadgeCalculator {
   }
 
   // @todo we need to flesh this out
-  // For now it just checks that a license and copyright statements are present
+  // For now it just checks that a license and copyright holders are present
   calculate() {
-    const hasLicense = _.get(this.definition, 'licensed.license', false);
-    const hasCopyright = _.get(this.definition, 'licensed.copyright.statements[0]', false);
-    if (hasLicense && hasCopyright) {return 3;}
-    if (hasLicense || hasCopyright) {return 2;}
-    return 1;
+    const hasLicense = _.get(this.definition, "licensed.license", false);
+    const hasCopyright = _.get(
+      this.definition,
+      "licensed.copyright.holders[0]",
+      false
+    );
+    if (hasLicense && hasCopyright) {
+      return 2;
+    }
+    if (hasLicense || hasCopyright) {
+      return 1;
+    }
+    return 0;
   }
 }
 
--- a/business/badgeCalculator.js
+++ b/business/badgeCalculator.js
@@ -5,12 +5,12 @@
 // For now we are just going to go with the simple:
 // If license/copyright there it gets a 3. Only one of those it gets a 2. None, it gets a 1
 
-const _ = require("lodash");
+const _ = require('lodash');
 
 const scoreToUrl = {
-  0: "https://img.shields.io/badge/ClearlyDefined%20Score-0-red.svg",
-  1: "https://img.shields.io/badge/ClearlyDefined%20Score-1-yellow.svg",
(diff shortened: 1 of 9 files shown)


business/badgeCalculator.js
```diff
@@ -0,0 +1,35 @@
...
+// SPDX-License-Identifier: MIT
+
+// Responsible for calculating what badge a component/definition has.
+// For now we are just going to go with the simple:
+// If license/copyright there it gets a 3. Only one of those it gets a 2. None, it gets a 1
+
+const _ = require('lodash');
+
+const scoreToUrl = {
+  1: 'https://img.shields.io/badge/Clearly%20Defined%20Score-1-red.svg',
```
jeffmcaffer: ClearlyDefined is one word

business/badgeCalculator.js
```diff
@@ -0,0 +1,35 @@
...
+  getBadgeUrl() {
+    return scoreToUrl[this.calculate()];
+  }
+
+  // @todo we need to flesh this out
+  // For now it just checks that a license and copyright statements are present
+  calculate() {
+    const hasLicense = _.get(this.definition, 'licensed.license', false);
+    const hasCopyright = _.get(this.definition, 'licensed.copyright.statements[0]', false);
+    if (hasLicense && hasCopyright) {return 3;}
```
jeffmcaffer: formatting. Convention would have this be 
```javascript
   if (hasLicense && hasCopyright) 
     return 3;
```

business/badgeCalculator.js
jeffmcaffer: Feels odd to give component a positive score when nothing is defined.  Propose 0  when they have nothing

test/business/badgeCalculator.js
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2018, The Linux Foundation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+const {expect} = require('chai');
```
jeffmcaffer: formatting nit: space around '{' '}'

test/business/badgeCalculator.js
```diff
@@ -0,0 +1,48 @@
...
+    };
+    const badge = BadgeCalculator(definition);
+    expect(badge.calculate()).to.eq(2);
+  });
+
+  it('gives level 2 if you have just copyright', () => {
+    const definition = {
+      licensed: {
+        copyright: {
+          statements: ['My copyright']
```
jeffmcaffer: You wouldn't know this from the spec etc but we are focused on copyright `holders`. we'll likely be removing statements and authors in the not too distant future.

routes/badges.js
```diff
@@ -0,0 +1,26 @@
...
+
+const asyncMiddleware = require('../middleware/asyncMiddleware');
+const express = require('express');
+const router = express.Router();
+const utils = require('../lib/utils');
+const badgeCalculator = require('../business/badgeCalculator');
+
+router.get('/:type/:provider/:namespace/:name/:revision', asyncMiddleware(getComponentBadge));
+
+async function getComponentBadge(request, result) {
```
jeffmcaffer: For the most part we have been putting the route function directly inline in the `get` parameters. One or two places where we need to share and implementation between multiple routes the function gets factored out. Suggest here merging the named function up into the last `get` param. Mostly for code consistency.

business/badgeCalculator.js
tesmart: Good to know.

business/badgeCalculator.js
tesmart: Can we use something like prettier? It eliminates formatting issues and let's people focus on the logic/business use of the code.

business/badgeCalculator.js
tesmart: I like this idea. It would be pretty jarring to get a score of 0.

test/business/badgeCalculator.js
tesmart: OK I can change what this looks up.

routes/badges.js
tesmart: Keeping it a separate function makes for easier testing (not that I've added a test for this 😆 ). For testing purposes I'm not a big fan of longer, anonymous functions.  This reminds me that I should add a test for this.

business/badgeCalculator.js
tesmart: James long gives a great prettier talk here (8 mins) https://www.youtube.com/watch?v=hkfBvpEfWdA

jeffmcaffer: totally fine with prettier or any other formatter as long as there is an agreed to format. Looking for consistency. I've been using VS Code and the default formatting (very simplistic) has been fine. My only two formatting criteria are
* no artificially short lines (80 chars is very 80s.
* minimize vertical space in general (e.g., don't put every property on its own line in object literals unless the line would be way too long)

I have a personal preference for no ';' but will not mind either way.

Will start a thread on the implementation channel...

routes/badges.js
jeffmcaffer: fair enough. I'm fine changing the way the others are done.

tesmart:
Hi @jeffmcaffer; I added tests for the router function which caused me to refactor it a little bit to make it more modular/testable. As far as style, once this gets merged, the prettier PR can take care of that.

tesmart:
Oh, I should probably add documentation, too. I'll look at that tonight.

tesmart:
I added some documentation to the README. Note that I couldn't get the image to display in markdown locally because I still have to pass an authorization header, i.e. `-H "Authorization: Bearer TOKEN"`.

I need to find a way around this locally.

PR #75, opened 2018-02-11
try prettier

jeffmcaffer:
Just as a trial this sets up prettier with line width 120 and semi-colons on. Otherwise the default options. All the code in the main service source tree was then reformatted with those settings. 

For the most part the output is fine IMHO. There are a few notable exceptions. The main (only real) issue I have is managing breakpoints in the face of aggressive collapsing of lines.  For example

```javascript
for (let i = 0; i < 100; i++)
  if (a === b)
    return 42;  
```
vs

```javascript
for (let i = 0; i < 100; i++)
  if (a === b) return 42;  
```

If you want to a breakpoint where the `return 42` happens, it is challenging. Dev tools like VS Code and Chrome support line breakpoints, not expression breaks. The same scenario happens with lambdas. Consider the same example but with an array `forEach`.

A note on line length. My concern about line length has to do with its impact on name selection. Naming is hard. IMHO names should be semantically useful and all but the most obvious and universally accepted shortenings should be avoided. This means that property and method names may be a little longer. That makes lines longer. If the line length limit is too short, well-named yet structurally simple single lines become sparse vertical structures. This in turn may subtlely nudge devs into choosing shorter, less valuable names.

In some cases it would be possible to restructure the code (e.g., use temporary variables rather than short inline expressions). In some cases that would be `a good thing`, in others, somewhat artificial.

In this example, 120 was not overly disruptive to the code. Going to 100 introduced several uncomfortable scenarios. 80 led to quite some _veritcalization_.

So, let's have a quick round of comments and then pick and go.


providers/curation/github.js
```diff
@@ -193,7 +193,13 @@ class GitHubCurationService {
     const target_url = `https://dev.clearlydefined.io/curation/${pr.title}/pr/${pr.number}`;
     try {
       return github.repos.createStatus({
-        owner, repo, sha, state, description, target_url, context: 'ClearlyDefined'
+        owner,
```
jeffmcaffer: this is a good example of the line length topic.  this expression is about as simple as they come yet it is formatted over 9 lines rather than 1 (or 2-3). 

Not that in this case prettier split the lines because the original had the object split over two lines. If the original is all on one line, prettier leaves it.  But, if one of the property names was even one char longer, we get the same many line format out.

In this case we can get it to 103 chars by declaring `const context = 'ClearlyDefined'` and then using that in the last element of the literal.

providers/curation/github.js
```diff
@@ -227,17 +232,15 @@ class GitHubCurationService {
   }
 
   async ensureCurations() {
-    if (this.curationUpdateTime && (Date.now - this.curationUpdateTime < this.options.curationFreshness))
-      return;
+    if (this.curationUpdateTime && Date.now - this.curationUpdateTime < this.options.curationFreshness) return;
```
jeffmcaffer: getting a little long. One could argue that in the interest of clarity it should really be 

```javascript
  const cacheStillValid = this.curationUpdateTime && Date.now - this.curationUpdateTime < this.options.curationFreshness;
  if (cacheStillValid) return;
```
or
```javascript
    const age = this.curationUpdateTime ? Date.now - this.curationUpdateTime : 0;
    if (age < this.options.curationFreshness) return;
```
and that that code would actually be clearer. I could go with that.

providers/stores/file.js
```diff
@@ -95,14 +91,19 @@ class FileStore extends AbstractStore {
...
-      return new Promise((resolve, reject) =>
-        fs.readFile(file, (error, data) =>
-          error ? reject(error) : resolve({ name: file, content: JSON.parse(data) })));
-    }));
+    const contents = await Promise.all(
+      files.map(file => {
+        return new Promise((resolve, reject) =>
+          fs.readFile(
+            file,
+            (error, data) => (error ? reject(error) : resolve({ name: file, content: JSON.parse(data) }))
```
jeffmcaffer: not very happy about this but it could handled by declaring the callback function before the `readFile` and then just passing that in. changes what looks like a scary nested structure into simple, declare callback, pass callback.

routes/curations.js
```diff
@@ -7,38 +7,48 @@ const router = express.Router();
...
-router.get('/:type/:provider/:namespace/:name/:revision/pr/:pr', asyncMiddleware(async (request, response) => {
-  const coordinates = utils.toEntityCoordinatesFromRequest(request);
-  return curationService.get(coordinates, request.params.pr).then(result => {
-    if (result)
-      return response.status(200).send(result);
-    response.sendStatus(404);
-  });
-}));
+router.get(
+  '/:type/:provider/:namespace/:name/:revision/pr/:pr',
```
jeffmcaffer: Probably should ignore most of the router reformatting. They will almost all be changed because of the anonymous function being passed in as the last arg. As separate discussion has raised the point about making named functions for these to aid testing. Probably a good idea.

tesmart: I'd be very cautious about adding anything to this file, but it looks like there's a prettier ignore file if we want to ignore 1 or 2 files here or there https://github.com/prettier/prettier/issues/1428

providers/curation/github.js
tesmart: Yes, it's not perfect and I agree with you that this should be on one line. Perhaps this will be improved in later versions of prettier. There appears to be some discussion on the issue here https://github.com/prettier/prettier/issues/2716

Might be good to weigh in there.

tesmart:
This looks great! When I've used it in the past, no one could agree on every little item but we could all agree that the benefit of "not having to think about it" far outweighed the cost. 

Thanks!

middleware/github.js
```diff
@@ -44,11 +44,13 @@ function setupClient(req, token = null) {
 // get the user's teams (from GitHub or the cache) and attach them to the request
 async function setupTeams(req, token, client = null) {
   // anonymous users are not members of any team
-  if (!token)
-    return req.app.locals.user.github.teams = [];
+  if (!token) return (req.app.locals.user.github.teams = []);
 
   // check cache for team data; hash the token so we're not storing them raw
-  const hashedToken = await crypto.createHash('sha256').update(token).digest('hex');
+  const hashedToken = await crypto
```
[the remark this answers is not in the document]
mynameistechno: It may do this after a certain number of chained calls, regardless of line length.

middleware/github.js
```diff
@@ -67,15 +69,15 @@ async function setupTeams(req, token, client = null) {
 async function getTeams(client, org) {
   try {
     const resp = await client.users.getTeams();
-    return resp.data
-      .filter(entry => entry.organization.login === org)
-      .map(entry => entry.name);
+    return resp.data.filter(entry => entry.organization.login === org).map(entry => entry.name);
```
mynameistechno: Here it actually put it all on one line for us 👍

providers/curation/github.js
mynameistechno: > Not that in this case prettier split the lines because the original had the object split over two lines. If the original is all on one line, prettier leaves it.

Yeah I've noticed at times it'll leave it in it's original form.

routes/curations.js
mynameistechno: +1 on named functions for testability and composability (when called for, e.g. large-ish functions).

providers/curation/github.js
mynameistechno: +1 for readability. When someone reads that code for the first time, i.e. `if (cacheStillValid) return;` they immediately know what is going on.

mynameistechno:
> for (let i = 0; i < 100; i++)
>   if (a === b)
>     return 42;

Personally I always use curly braces 😝 So I would've just written it as 
```js
for (let i = 0; i < 100; i++) {
  if (a === b) {
    return 42;
  }
}

```
I think this is mostly out of habit/consistency. It's just more explicit. There is no chance of someone in the future adding lines to that `if` and creating a bug. E.g.
```js
  if (a === b)
    b = -1; 
    return b;
```
Although nowadays with sufficient tests, and the use of eslint, prettier, and PR reviews I suspect this is a non-issue.

middleware/github.js
jeffmcaffer: yeah, Would love to understand the inconsistency.  On the one hand, the first one above was structurally simpler yet ended up on multiple lines. This one has two lambdas yet is all on one line...

routes/curations.js
jeffmcaffer: we should just "fix" the structure IMHO rather than ignore the files. We'd have to ignore all the routes to get around the issues.

middleware/github.js
tesmart: i think there may be a 3-chain then break rule, I'll have to look it up tomorrow. May just be a suggestion someone made.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
tesmart: This file should not be in git

jeffmcaffer: IMHO it should. this is a "workspace setting" that means anyone working on this repo in VS Code will have format on save so that we don't have formatting issues in PRs etc.

middleware/github.js
tesmart: Ah yes it was merged here https://github.com/prettier/prettier/issues/1565

And via the comments it seems like this should be broken up b/c it's a 3-chain call. So I played with it locally and it *is* broken up if you limit the chars to 80, in fact it's the same code as the original. But at 120 it stays on 1 line, so it seems that this rule only applies after the char limit is reached.

.vscode/settings.json
tesmart: I see your point and I'm afraid I was unnecessarily terse in my comment above :) It's a good idea to enforce the formatting. Could we use https://github.com/prettier/eslint-plugin-prettier to make CI fail if your code is not formatted properly? This way we could leave all editor settings out of the repo and have people arrive at a clean build however they want.

jeffmcaffer:
fine trial.  we are going to move ahead with these settings plus semi: false.

PR #109, opened 2018-02-26
Rate-limit authenticated routes

jpeddicord:
Disabled by default, to enable set:

* `RATE_LIMIT_WINDOW` to the number of seconds you want to monitor a given IP for
* `RATE_LIMIT_MAX` to the maximum number of requests from an IP during the window

Note the `'trust-proxy'` setting -- I added this with the assumption that CloudFlare is fronting the request. If this isn't set, rate limits will be too aggressive as it'll think that CloudFlare's internal IP is the requester IP. This is derived from `X-Forwarded-For`; see https://support.cloudflare.com/hc/en-us/articles/200170986-How-does-Cloudflare-handle-HTTP-Request-headers-

### DIFF
--- a/app.js
+++ b/app.js
@@ -4,6 +4,7 @@ const express = require('express');
 const logger = require('morgan');
 const bodyParser = require('body-parser');
 const cors = require('cors');
+const RateLimit = require('express-rate-limit');
 const helmet = require('helmet');
 const serializeError = require('serialize-error');
 const requestId = require('request-id/express');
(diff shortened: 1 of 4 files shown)


jeffmcaffer:
Interesting. Thanks. Is this all that is needed? any suggestions for values? For now we can likely be generous as we are really only protecting against errant code from people learning the API and such.

jpeddicord:
@jeffmcaffer Perhaps something on the order of 100 requests/min? Or 500/5min (window 300, max 500) to allow for something more bursty.

Once we scale up, we'll want to add the Redis backend to share limit state, but it's a drop-in component here.

jeffmcaffer:
sure. The bursty thing makes sense.  

An interesting question will be how does the WebUI behave if it gets rate limited. It's totally possible that the current code is not very fault tolerant... anyway, lets get something in and go from there.

jpeddicord:
Yeah, the UI basically does nothing on failure from my testing. It'll still function again if you wait, but there's no indication it broke.

OK to merge? Will the new dependency get picked up on deployment?

PR #111, opened 2018-02-27
Allow localhost to access dev-api

jpeddicord:
This is accomplished by checking if the referrer is localhost, and then
setting a cookie with the real localhost origin. This cookie is read on
finalize and substituted for the real UI in postMessage.


mynameistechno:
@cohenmax9 check this out. I know you were having issues getting website and service to work together. With this you should be able to use the dev-api.

PR #114, opened 2018-02-27
get permissions on authentications

jeffmcaffer:
users authenticating should get back from the service a list of permissions they have. this enables the UI to properly enable/disable or hide/show elements.

### DIFF
--- a/routes/auth.js
+++ b/routes/auth.js
@@ -95,7 +95,7 @@ async function getPermissions(token, org) {
       .filter(entry => entry.organization.login === org)
       .map(entry => entry.name)
       .map(findPermissions)
-      .filter(e => e)
+      .filter(e => e);
   } catch (error) {
     if (error.code === 404) {
       console.error('GitHub returned a 404 when trying to read team data. ' +
--- a/routes/auth.js
+++ b/routes/auth.js
@@ -63,7 +63,7 @@ router.get('/github/finalize', passportOrPat(),
 
     // allow for sending auth responses to localhost on dev site; see /github
     // route above. real origin is stored in cookie.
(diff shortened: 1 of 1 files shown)


routes/auth.js
```diff
@@ -71,15 +73,54 @@ router.get('/github/finalize', passportOrPat(),
...
 
+/**
+ * Fetch a list of ClearlyDefined permissions related to the given token in the identified org.
+ * @param {string} token - GitHubApi token
+ * @param {string} org - org name to filter teams
+ * @returns {Promise<Array<string>>} - list of permission names
+ */
+async function getPermissions(token, org) {
+  const options = { headers: { 'user-agent': 'clearlydefined.io' } };
+  const client = new GitHubApi(options);
```
jpeddicord: low-pri, sometime we'll want to consolidate the client creations. partially my fault. :)

routes/auth.js
```diff
@@ -56,12 +56,14 @@ router.get('/github/start', passportOrPat(), (req, res) => {
...
-    const safeToken = encodeURIComponent(req.user.githubAccessToken);
-    let origin = config.endpoints.website;
+  async (req, res) => {
+    const token = req.user.githubAccessToken;
+    const permissions = await getPermissions(token, config.auth.github.org);
+    const result = JSON.stringify({ type: 'github-token', token, permissions });
 
     // allow for sending auth responses to localhost on dev site; see /github
     // route above. real origin is stored in cookie.
+    const origin = config.endpoints.website;
```
jpeddicord: This needs to be `let` still (possible re-assignment in following statement)

PR #123, opened 2018-03-06
Add proactive definitions caching

geneh:


business/cacheRefresher.js
```diff
@@ -0,0 +1,25 @@
...
+
+const DefinitionsCache = require('./definitionsCache')
+const utils = require('../lib/utils')
+const _ = require('lodash')
+
+class CacheRefresher {
+  constructor(harvest, curation) {
+    this.harvestService = harvest
+    this.curationService = curation
+    setInterval(this.refreshDefinitions.bind(this), 4 * 60 * 1000)
```
iamwillbar: If `refreshDefinitions` fails for any reason then it will take down the process, I'd wrap it in another error safe function that logs any errors but suppresses them,

business/cacheRefresher.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  async refreshDefinitions() {
+    const coordinates = utils.toEntityCoordinatesFromRequest({ params: {} })
+    const curated = await this.curationService.list(coordinates)
+    const harvest = await this.harvestService.list(coordinates)
+    const stringHarvest = harvest.map(c => c.toString())
+    const result = _.union(stringHarvest, curated)
+    DefinitionsCache.set(coordinates, result, 5 * 60 * 1000)
```
iamwillbar: I don't think you've allowed enough time between your refresh interval and cache duration, we're at 40+ seconds today and you only have a 1 minute difference. Any slowness or increase in growth will kill that 20 second buffer. I'd go double for now. We can do smarter cache population and invalidation in the future.

PR #126, opened 2018-03-08
Improve versions comparison in getLatestVersion

geneh:
This change is necessary as a preparation for supporting scancode 2.9.0b1. See issue https://github.com/clearlydefined/crawler/issues/81.


jeffmcaffer:
no need to coerce. We are making sure the tool version is always a semver

PR #133, opened 2018-05-10
some initial support for multiple changes per PR

jeffmcaffer:
DO NOT MERGE

this is some WIP of enabling one contribution to update multiple components.

### DIFF
--- a/business/definitionService.js
+++ b/business/definitionService.js
@@ -149,7 +149,7 @@ class DefinitionService {
     const recompute = mode === 'definitions'
     const baseList = coordinatesList || (await this.list(new EntityCoordinates(), recompute))
     const list = baseList.map(entry => EntityCoordinates.fromString(entry))
-    await Promise.all(
+    return await Promise.all(
       list.map(
         throat(10, async coordinates => {
           const definition = await this.get(coordinates, null, recompute)
@@ -159,8 +159,6 @@ class DefinitionService {
         })
       )
     )
-    // don't forget to store anything left over
-    return this.search.store(indexes)
   }
 
   /**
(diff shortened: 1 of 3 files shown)


providers/curation/github.js
```diff
@@ -34,6 +34,69 @@ class GitHubCurationService {
...
+    const { owner, repo, branch } = this.options
+
+    // Get the sha for the current master and create a branch for the PR
+    const master = await github.repos.getBranch({ owner, repo, branch: `refs/heads/${branch}` })
+    const sha = master.data.commit.sha
+    const prBranch = this._getBranchName(coordinates)
+    await github.gitdata.createReference({ owner, repo, ref: `refs/heads/${prBranch}`, sha })
+
+    // write all the patch files
+    await Promise.all(
```
jpeddicord: If I'm reading this right, we'll end up with 1 commit per change. Obviously this can be squashed when merging PRs, but it does make for large histories. Is this fine?

business/definitionService.js
```diff
@@ -159,8 +159,6 @@ class DefinitionService {
         })
       )
     )
-    // don't forget to store anything left over
-    return this.search.store(indexes)
```
jpeddicord: I've been trying to diagnose what's going on here but haven't had much luck. `indexes` isn't declared, so this didn't work anyway. I had hacked around that today by storing `list` instead, which seemed to work. Removing this entirely leaves me with an empty index (at least for in-memory + file stores) and a bunch of errors looking for yaml files that don't exist, e.g. `'/tmp/cd-5G4p2p/curated-data-dev/curations/npm/npmjs/-/colors/.yaml'`

PR #136, opened 2018-05-24
Handled new request object for multi component curation

Teju-Manchenella:


providers/curation/github.js
```diff
@@ -34,86 +37,72 @@ class GitHubCurationService {
...
+  }
+
+  async addOrUpdate(userGithub, serviceGithub, patch) {
+    const { owner, repo, branch } = this.options
+    const master = await serviceGithub.repos.getBranch({ owner, repo, branch: `refs/heads/${branch}` })
+    const sha = master.data.commit.sha
+    const prBranch = await this._getBranchName(userGithub, serviceGithub)
+    await serviceGithub.gitdata.createReference({ owner, repo, ref: `refs/heads/${prBranch}`, sha })
+    await Promise.all(patch.patches.map(throat(1, component => this._writePatch(userGithub, serviceGithub, patch.description, component, prBranch))))
+    // Create the PR using service github object
```
jeffmcaffer: we want to create the PR using `userGitHub` so it is attributed to the right person

providers/curation/github.js
```diff
@@ -304,19 +327,21 @@ class GitHubCurationService {
     return coordinates.toString()
   }
 
-  _getBranchName(coordinates) {
-    const c = coordinates
-    return `${c.type.toLowerCase()}_${c.name}_${c.revision}_${moment().format('YYMMDD_HHmmss.SSS')}`
+  async _getBranchName(userGithub, serviceGithub) {
+    const committer = userGithub ? await userGithub.users.get({}) : await serviceGithub.users.get({})
```
jeffmcaffer: Probably should get this info once and stash it. We can do it in another PR but too many API calls to GitHub make them sad.

providers/curation/github.js
```diff
@@ -304,19 +327,21 @@ class GitHubCurationService {
...
+    return `${committer.data.name.toLowerCase().replace(/ /g, "_")}_${moment().format('YYMMDD_HHmmss.SSS')}`
```
jeffmcaffer: Interesting problem. Not sure that all chars in a GitHub user name are valid as a branch name. Should think about this strategy some. The same is technically true of package names (the previous approach) but seemingly less likely. Names are all over the place with people using unicode and emojis and ...

providers/curation/github.js
```diff
@@ -304,19 +327,21 @@ class GitHubCurationService {
...
   }
 
+
```
jeffmcaffer: each extra blank line causes pain to cute kittens

routes/curations.js
```diff
@@ -44,8 +45,23 @@ router.patch(
   '/:type/:provider/:namespace/:name/:revision',
   asyncMiddleware(async (request, response) => {
     const coordinates = utils.toEntityCoordinatesFromRequest(request)
+    const token = request.app.locals.config.curation.store.github.token
+    const serviceGithub = Github.getClient({ token })
```
jeffmcaffer: seems like we should be setting up the serviceGitHub client somewhere near `request.app.locals.user.github.client` on the request using the same middleware as for the user client. Then we are just using it here.

routes/swagger.yaml
```diff
@@ -218,20 +218,25 @@ paths:
             application/json:
               schema:
                 $ref: '#components/bodies/definitionList'
-
-  /definitions/{type}/{provider}/{namespace}/{name}:
     get:
```
jeffmcaffer: There are other updates to do here as we changed the PATCH /definition contract to the stuff that is written on your whiteboard. I can help you through that

providers/curation/github.js
Teju-Manchenella: I was able to cache userGithub.users.get({}); however, I am still unclear when to cache the serviceGithub.users.get({})

routes/curations.js
Teju-Manchenella: This is not being called anymore. This was the old patch function.

routes/swagger.yaml
Teju-Manchenella: I was unsure about this comment and forgot to ask about it today

providers/curation/github.js
```diff
@@ -34,86 +37,71 @@ class GitHubCurationService {
...
+    const result = {
+      coordinates: {
+        type: coordinates.type,
+        provider: coordinates.provider,
+        namespace: coordinates.namespace === '-' ? null : coordinates.namespace,
+        name: coordinates.name
+      }
+    }
+    result.revisions = assign(get(currentContent, 'revisions') || {})
+    forIn(newContent, function (value, key) {
```
jeffmcaffer: arrow function here.

providers/curation/github.js
```diff
@@ -34,86 +37,71 @@ class GitHubCurationService {
...
+      result.revisions[key] = merge(result.revisions[key] || {}, value)
+    })
+    return yaml.safeDump(result, { sortKeys: true })
+  }
+
+
+  async _writePatch(userGithub, serviceGithub, info, description, patch, branch) {
+    const { owner, repo } = this.options
+    const coordinates = new EntityCoordinates(patch.coordinates.type, patch.coordinates.provider, patch.coordinates.namespace, patch.coordinates.name)
```
jeffmcaffer: is `patch.coordinates` already an EntityCoordinates? if so, use `asRevisionless` (or whatever). If not, consider making a static helper that takes `patch.coordinates` and gives you back an EntityCoordinates that you can call `asRevisionless` on.

There is some code above where this could be useful as well.

providers/curation/github.js
```diff
@@ -34,86 +37,71 @@ class GitHubCurationService {
...
-        delete parsedContent.revisions
-        parsedContent.revisions = revisions || {}
-
-        // extract the file's SHA1 to enable updating the file
-        if (parsedContent.origin) {
-          curationPathSha = parsedContent.origin.sha
-        }
+    const message = `Update ${path}`
+    if (currentContent && currentContent._origin)
+      return serviceGithub.repos.updateFile(pickBy({
```
jeffmcaffer: why `pickBy` here and in the false case?

providers/curation/github.js
```diff
@@ -34,86 +37,71 @@ class GitHubCurationService {
...
-          patch.patch
-        )
+  async addOrUpdate(userGithub, serviceGithub, info, patch) {
+    const { owner, repo, branch } = this.options
+    const masterBranch = await serviceGithub.repos.getBranch({ owner, repo, branch: `refs/heads/${branch}` })
+    const sha = masterBranch.data.commit.sha
+    const prBranch = await this._getBranchName(userGithub, serviceGithub, info)
+    await serviceGithub.gitdata.createReference({ owner, repo, ref: `refs/heads/${prBranch}`, sha })
+    await Promise.all(patch.patches.map(throat(1, component => this._writePatch(userGithub, serviceGithub, info, patch.description, component, prBranch))))
+    const master = userGithub ? userGithub : serviceGithub
```
jeffmcaffer: simplify to 

```javascript
return (userGitHub || serviceGitHub).pullRequests.create({
...
```

providers/curation/github.js
```diff
@@ -304,19 +326,21 @@ class GitHubCurationService {
     return coordinates.toString()
   }
 
-  _getBranchName(coordinates) {
-    const c = coordinates
-    return `${c.type.toLowerCase()}_${c.name}_${c.revision}_${moment().format('YYMMDD_HHmmss.SSS')}`
+  async _getBranchName(userGithub, serviceGithub, info) {
+    const committer = userGithub ? info : await serviceGithub.users.get({})
```
jeffmcaffer: should not have to do this now that we have the userInfo cached in the request. Might have to pass in the userInfo rather than the two clients.

routes/curations.js
jeffmcaffer: this should be deleted. if it no longer used

test/providers/caching/redis.js
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+
+const sinon = require('sinon')
+const sandbox = sinon.createSandbox()
+const redis = require('redis')
+const assert = require('assert')
+const redisCache = require('../../../providers/caching/redis')
+
+describe('get a tool result ', () => {
```
jeffmcaffer: need to update the descriptive strings here and below to something that makes sense.

Likely need some additional tests as well.

providers/curation/github.js
```diff
@@ -62,26 +62,10 @@ class GitHubCurationService {
...
-    return serviceGithub.repos.createFile({
-      owner,
-      repo,
-      path,
-      message,
-      content,
-      branch,
-      committer: `{ "name": "${info.name || info.login}", "email": "${info.email}" }`,
-    })
+      return serviceGithub.repos.updateFile(set(fileBody, sha, currentContent._origin.sha))
```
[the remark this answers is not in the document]
Teju-Manchenella: I am still trying to figure out when to use lodash commands and when to simply use javascript. Also I am assuming you meant 
```javascript
    if (get(currentContent, '_origin.sha'))
        fileBody.sha = currentContent._origin.sha
        return serviceGithub.repos.updateFile(fileBody)
    return serviceGithub.repos.createFile(fileBody)
```

PR #137, opened 2018-05-31
update the definition schema to include per file info

jeffmcaffer:
PROPOSAL: DO NOT MERGE

In an effort to support more usecases, this PR updates the schema for definitions and curations to be based on a data for a list of files rather than aggregates of that data. So rather than having a list of facets each of which as aggregated list of license and copyrights for their files, the data is now expressed on individual files. That per-file data can then be aggregated into per-facet data.

Also allow for curation of the data on a per file basis.

Note that as a consequence of this approach, we are now able to have ClearlyDefined retain the actual content of specific files (e.g., LICENSE) for use in notice file generation and the like.

Once the schema changes are agreed to, the summarization code will be updated and likely added to this PR.

### DIFF
--- a/schemas/curation.json
+++ b/schemas/curation.json
@@ -21,6 +21,7 @@
       "enum": ["npmjs", "github", "mavencentral", "mavencentralsource"]
     },
     "coordinates": {
+      "type": "object",
       "required": ["type", "provider", "namespace", "name"],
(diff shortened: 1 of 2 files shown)


jeffmcaffer:
This is has been superceded by other work. closing.

PR #138, opened 2018-06-05
Definition schema refactor

Teju-Manchenella:
Refactored the creation of definitions so declared licenses not saved on a per file basis

### DIFF
--- a/providers/summary/scancode.js
+++ b/providers/summary/scancode.js
@@ -28,14 +28,23 @@ class ScanCodeSummarizer {
     const result = {}
     this.addDescribedInfo(result, coordinates, harvested)
     const buckets = this.computeFileBuckets(harvested.content.files, facets)
-    set(result, 'licensed.facets', {})
-    const facetsObject = get(result, 'licensed.facets')
-    const declaredLicenses = new Set()
-    for (const key in buckets) {
-      facetsObject[key] = this.summarizeLicenseInfo(buckets[key])
-      this.summarizeDeclaredLisense(buckets[key], declaredLicenses)
+    let declaredLicenses = new Set()
+    let packageLicenses = new Set()
+    for (const fileIdx in harvested.content.files) {
+      const file = harvested.content.files[fileIdx]
+      if (!file.path.toLowerCase().includes('/')) {
+        const asserted = get(file, 'packages[0].asserted_licenses')
+        if (asserted && packageLicenses.size === 0) this._addArrayToSet(asserted, packageLicenses, license => license.license || license.spdx_license_key)
+        declaredLicenses = this._summarizeLicenseFile(file)
+        if (declaredLicenses.size > 0) break
+      }
     }
-    set(result, 'licensed.declared', this._toExpression(declaredLicenses))
+    set(result, 'licensed.declared', this._toExpression(declaredLicenses || packageLicenses))
+    result.files = []
+    for (const key in buckets) {
+      set(result, `licensed.facets.${key}`, this.summarizeLicenseInfo(buckets[key]))
+      this.summarizeFileInfo(buckets[key], result.files, key)
+    }
     return result
   }
 
@@ -43,7 +52,6 @@ class ScanCodeSummarizer {
     const facetList = Object.getOwnPropertyNames(facets)
     remove(facetList, 'core')
     if (facetList.length === 0) return { core: files }
-
     const result = {}
     for (const facet in facetList) {
       const facetKey = facetList[facet]
@@ -63,10 +71,8 @@ class ScanCodeSummarizer {
   summarizeLicenseInfo(files) {
     const copyrightHolders = new Set()
     const licenseExpressions = new Set()
-    const declaredLicenses = new Set()
     let unknownParties = 0
     let unknownLicenses = 0
-
     for (let file of files) {
       this._addArrayToSet(file.licenses, licenseExpressions, license => license.spdx_license_key)
       const asserted = get(file, 'packages[0].asserted_licenses')
@@ -76,7 +82,6 @@ class ScanCodeSummarizer {
         !hasHolders && unknownParties++
       }
     }
-
     return {
       attribution: {
         parties: this._setToArray(copyrightHolders),
@@ -90,24 +95,24 @@ class ScanCodeSummarizer {
     }
   }
 
-  summarizeDeclaredLisense(files, declaredLicenses) {
+  summarizeFileInfo(files, currentFiles, facet) {
     files.forEach(file => {
-      const asserted = get(file, 'packages[0].asserted_licenses')
-      if (asserted) this._addArrayToSet(asserted, declaredLicenses, license => license.license || license.spdx_license_key)
-      this._addLicenseFiles(file, declaredLicenses)
+      const licenses = new Set()
+      this._addArrayToSet(file.licenses, licenses, license => license.license || license.spdx_license_key)
+      const licenseExpression = this._toExpression(licenses)
+      const fileObject = { path: file.path, license: licenseExpression, attributions: file.copyrights, facet: facet }
+      currentFiles.push(fileObject)
     })
   }
 
-  _addLicenseFiles(file, declaredLicenses) {
+  _summarizeLicenseFile(file) {
     // Look for license files at the root of the scanned code
     // TODO enhance in the future to cover more license management strategies.
-    const license = ['license', 'license.txt', 'license.md', 'license.html'].reduce((current, prefix) => {
-      if (!current) return file.path.toLowerCase().includes(prefix)
-      return current
-    }, false)
-    if (!license) return
-    if (!file.licenses) return
+    const declaredLicenses = new Set()
+    const isLicense = ['license', 'license.txt', 'license.md', 'license.html'].includes(file.path.toLowerCase())
+    if (!isLicense || !file.licenses) return declaredLicenses
     file.licenses.forEach(license => declaredLicenses.add(license.spdx_license_key))
+    return declaredLicenses
   }
 
   _toExpression(licenses) {
--- a/providers/summary/scancode.js
+++ b/providers/summary/scancode.js
@@ -30,9 +30,12 @@ class ScanCodeSummarizer {
     const buckets = this.computeFileBuckets(harvested.content.files, facets)
     set(result, 'licensed.facets', {})
     const facetsObject = get(result, 'licensed.facets')
+    const declaredLicenses = new Set()
     for (const key in buckets) {
       facetsObject[key] = this.summarizeLicenseInfo(buckets[key])
+      this.summarizeDeclaredLisense(buckets[key], declaredLicenses)
     }
+    set(result, 'licensed.declared', this._toExpression(declaredLicenses))
     return result
   }
 
@@ -72,8 +75,6 @@ class ScanCodeSummarizer {
         const hasHolders = this._normalizeCopyrights(file.copyrights, copyrightHolders)
         !hasHolders && unknownParties++
       }
-      this._addArrayToSet(asserted, declaredLicenses, license => license.license || license.spdx_license_key)
-      this._addLicenseFiles(file, declaredLicenses)
     }
 
     return {
@@ -81,7 +82,6 @@ class ScanCodeSummarizer {
         parties: this._setToArray(copyrightHolders),
         unknown: unknownParties
       },
-      declared: this._toExpression(declaredLicenses),
       discovered: {
         expressions: this._setToArray(licenseExpressions),
         unknown: unknownLicenses
@@ -90,10 +90,22 @@ class ScanCodeSummarizer {
     }
   }
 
+  summarizeDeclaredLisense(files, declaredLicenses) {
+    files.forEach(file => {
+      const asserted = get(file, 'packages[0].asserted_licenses')
+      if (asserted) this._addArrayToSet(asserted, declaredLicenses, license => license.license || license.spdx_license_key)
+      this._addLicenseFiles(file, declaredLicenses)
+    })
+  }
+
   _addLicenseFiles(file, declaredLicenses) {
     // Look for license files at the root of the scanned code
     // TODO enhance in the future to cover more license management strategies.
-    if (!['license', 'license.txt', 'license.md', 'license.html'].includes(file.path.toLowerCase())) return
+    const license = ['license', 'license.txt', 'license.md', 'license.html'].reduce((current, prefix) => {
+      if (!current) return file.path.toLowerCase().includes(prefix)
+      return current
+    }, false)
+    if (!license) return
     if (!file.licenses) return
     file.licenses.forEach(license => declaredLicenses.add(license.spdx_license_key))
   }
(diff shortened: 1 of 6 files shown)


providers/summary/scancode.js
```diff
@@ -90,10 +90,22 @@ class ScanCodeSummarizer {
     }
   }
 
+  summarizeDeclaredLisense(files, declaredLicenses) {
```
jeffmcaffer: spelling of license

providers/summary/scancode.js
```diff
@@ -90,10 +90,22 @@ class ScanCodeSummarizer {
...
+    files.forEach(file => {
+      const asserted = get(file, 'packages[0].asserted_licenses')
+      if (asserted) this._addArrayToSet(asserted, declaredLicenses, license => license.license || license.spdx_license_key)
+      this._addLicenseFiles(file, declaredLicenses)
+    })
+  }
+
   _addLicenseFiles(file, declaredLicenses) {
```
jeffmcaffer: this should be addLicenseFile (singular)

providers/summary/scancode.js
```diff
@@ -90,10 +90,22 @@ class ScanCodeSummarizer {
...
     // Look for license files at the root of the scanned code
     // TODO enhance in the future to cover more license management strategies.
-    if (!['license', 'license.txt', 'license.md', 'license.html'].includes(file.path.toLowerCase())) return
+    const license = ['license', 'license.txt', 'license.md', 'license.html'].reduce((current, prefix) => {
```
jeffmcaffer: There must be a lodash function to find the first element in an array that matches a with some function.

jeffmcaffer: suggest calling this `isLicense`. `is`, `has` prefix is a good convention for boolean variables, props and functoins

providers/summary/scancode.js
```diff
@@ -72,16 +75,13 @@ class ScanCodeSummarizer {
         const hasHolders = this._normalizeCopyrights(file.copyrights, copyrightHolders)
         !hasHolders && unknownParties++
       }
-      this._addArrayToSet(asserted, declaredLicenses, license => license.license || license.spdx_license_key)
```
jeffmcaffer: should remove the const `declaredLicenses` above

routes/curations.js
```diff
@@ -47,7 +47,7 @@ router.patch(
     const info = request.app.locals.user.github.info
     return curationService
       .addOrUpdate(userGithub, serviceGithub, info, request.body)
-      .then(() => response.sendStatus(200))
+      .then(result => { response.status(200).send({ prNumber: result.data.number }) })
```
jeffmcaffer: don't think you need the outer `{}`

providers/summary/scancode.js
```diff
@@ -90,10 +90,22 @@ class ScanCodeSummarizer {
...
+
   _addLicenseFiles(file, declaredLicenses) {
     // Look for license files at the root of the scanned code
     // TODO enhance in the future to cover more license management strategies.
-    if (!['license', 'license.txt', 'license.md', 'license.html'].includes(file.path.toLowerCase())) return
+    const license = ['license', 'license.txt', 'license.md', 'license.html'].reduce((current, prefix) => {
+      if (!current) return file.path.toLowerCase().includes(prefix)
+      return current
+    }, false)
+    if (!license) return
```
jeffmcaffer: combine into `if (!isLicense || !file.licenses) return

providers/summary/scancode.js
```diff
@@ -30,9 +30,12 @@ class ScanCodeSummarizer {
     const buckets = this.computeFileBuckets(harvested.content.files, facets)
     set(result, 'licensed.facets', {})
     const facetsObject = get(result, 'licensed.facets')
+    const declaredLicenses = new Set()
     for (const key in buckets) {
       facetsObject[key] = this.summarizeLicenseInfo(buckets[key])
+      this.summarizeDeclaredLisense(buckets[key], declaredLicenses)
```
jeffmcaffer: I was thinking of looping over the files themselves, rather than the buckets. Several reasons. 

1. the bucketization is not relevant. the summarize just loops over the files anyway
1. files can theoretically be in multiple buckets. would be less efficient but should still be correct.

providers/summary/scancode.js
```diff
@@ -30,9 +30,12 @@ class ScanCodeSummarizer {
     const buckets = this.computeFileBuckets(harvested.content.files, facets)
     set(result, 'licensed.facets', {})
     const facetsObject = get(result, 'licensed.facets')
```
jeffmcaffer: still feels odd to `set` and then immediately `get`

providers/summary/scancode.js
```diff
@@ -28,22 +28,30 @@ class ScanCodeSummarizer {
...
     this.addDescribedInfo(result, coordinates, harvested)
     const buckets = this.computeFileBuckets(harvested.content.files, facets)
-    set(result, 'licensed.facets', {})
-    const facetsObject = get(result, 'licensed.facets')
-    const declaredLicenses = new Set()
+    let declaredLicenses = new Set()
+    let packageLicenses = new Set()
+    for (const fileIdx in harvested.content.files) {
+      const file = harvested.content.files[fileIdx]
+      if (!file.path.toLowerCase().includes('/')) {
```
Teju-Manchenella: Only look at files that are in the root folder. File paths in the root folder will not have a slash in them

jeffmcaffer: OK. we should not really try to fight whole communities like that. So we need to adapt. there will be other cases as well I'm sure. How do you propose to accommodate the npm scenario?

providers/summary/scancode.js
```diff
@@ -28,22 +28,30 @@ class ScanCodeSummarizer {
...
+        const asserted = get(file, 'packages[0].asserted_licenses')
+        if (asserted && packageLicenses.size === 0) this._addArrayToSet(asserted, packageLicenses, license => license.license || license.spdx_license_key)
+        declaredLicenses = this._summarizeLicenseFile(file)
+        if (declaredLicenses.size > 0) break
```
jeffmcaffer: This (`break`) is a bit of smell. Indicates that the code might be doing too much in one function. Suggest here that the declared/package license code be broken out into something like `_summarizeDeclaredInfo`. The other loop might be good to factor out into say `_summarizeDiscoveredInfo`

providers/summary/scancode.js
```diff
@@ -28,22 +28,30 @@ class ScanCodeSummarizer {
...
+      }
+    }
+    set(result, 'licensed.declared', this._toExpression(declaredLicenses || packageLicenses))
+    result.files = []
     for (const key in buckets) {
-      facetsObject[key] = this.summarizeLicenseInfo(buckets[key])
-      this.summarizeDeclaredLisense(buckets[key], declaredLicenses)
+      set(result, `licensed.facets.${key}`, this.summarizeLicenseInfo(buckets[key]))
+      this.summarizeFileInfo(buckets[key], result.files, key)
```
jeffmcaffer: Should not have to pass `result.files` here. `files` should be set by the other loop that is going over all the files. That is independent of the facetization (if any).

schemas/curation.json
```diff
@@ -2,7 +2,10 @@
   "$id": "https://api.clearlydefined.io/schemas/curation.json#",
   "$schema": "http://json-schema.org/draft-06/schema#",
   "title": "curation",
-  "required": ["coordinates", "revisions"],
+  "required": [
+    "coordinates",
```
jeffmcaffer: this reformating is incorrect wrt the project style. Install `prettier` on VS Code and let it format based on the config in the repo.

schemas/definition.json
```diff
@@ -44,11 +68,54 @@
...
+        "required": [
+          "path"
+        ],
+        "additionalProperties": false,
+        "properties": {
+          "path": {
+            "type": "string",
+            "description": "The absolute path of the file as found in the component."
+          },
+          "facet": {
```
jeffmcaffer: `facet` and `facets`?  why both?  Seems like just facets. It might have been me who added both. either way it is wrong.

providers/summary/scancode.js
Teju-Manchenella: I decided to look at the last component of each file path instead. This allows us to find all the license/package files even though they are not in their proper location in the root folder. Its tough to expect everyone to follow the open source guidelines because it seems like most everyone is doing it wrong right now anyway. I do think we should take into account the fact that the license/package files are not in the root folder in the clearlyDefined score.

providers/summary/scancode.js
Teju-Manchenella: I have broken it up it into 4 helper functions (_summarizeDeclaredLicenseInfo, _summarizePackageInfo, _summarizeFileInfo, and _summarizeDiscoveredLicenseInfo). Please let me know if that is too verbose. I have always been bad at writing modular code, so any comments would be helpful.

providers/summary/scancode.js
Teju-Manchenella: I think there was a misunderstanding in my last commit ("Look at last component of path for license/package, made code more modular", e8035f3c48722aa75389b612df2e857345d685b2). It seems that each file needs to be tagged that facets it belongs to. It seems that I have to compute per file info in this loop since I will know exactlu which facet I am looping over. If I summarize per file info in its own loop, then how will each file know what facet it belongs to. I think I am not understanding how you know which facet a file belongs to.

providers/summary/scancode.js
```diff
@@ -90,12 +85,49 @@ class ScanCodeSummarizer {
...
-    // TODO enhance in the future to cover more license management strategies.
-    if (!['license', 'license.txt', 'license.md', 'license.html'].includes(file.path.toLowerCase())) return
-    if (!file.licenses) return
-    file.licenses.forEach(license => declaredLicenses.add(license.spdx_license_key))
+  _summarizeDeclaredLicenseInfo(files) {
+    const declaredLicenses = new Set()
+    for (const fileIdx in files) {
+      const file = files[fileIdx]
+      const pathArray = file.path.split('/')
+      const baseName = last(pathArray)
```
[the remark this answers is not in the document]
Teju-Manchenella: If we only check the root we won't be able to get any declared licenses for npm packages. It seems like all the npm packages have their license file inside the package directory. Should we make this specific exception for npm? Along with that, in the future would we have to make these exceptions for each type of package manager depending on how they structure their projects?

providers/summary/scancode.js
```diff
@@ -27,11 +27,13 @@ class ScanCodeSummarizer {
...
+    const declaredLicenses = this._summarizeDeclaredLicenseInfo(harvested.content.files)
+    const packageLicenses = this._summarizePackageInfo(harvested.content.files)
+    set(result, 'licensed.declared', this._toExpression(declaredLicenses || packageLicenses))
+    result.files = this._summarizeFileInfo(harvested.content.files)
     const buckets = this.computeFileBuckets(harvested.content.files, facets)
-    set(result, 'licensed.facets', {})
-    const facetsObject = get(result, 'licensed.facets')
     for (const key in buckets) {
-      facetsObject[key] = this.summarizeLicenseInfo(buckets[key])
+      set(result, `licensed.facets.${key}`, this._summarizeDiscoveredLicenseInfo(buckets[key]))
```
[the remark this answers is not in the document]
Teju-Manchenella: That makes sense I can definitely take care of that

providers/summary/scancode.js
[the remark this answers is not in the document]
jeffmcaffer: If it is an NPM that we are scanning then the License file (if any) is of secondary importance. Since NPM (and typical other package managers) have explicit metadata to call out the license (e.g, entry in the package.json file), we should be using that information as the declared license. So this method returning nothing is fine as a first step. `summarizePackageInfo` should pick up the license in the metadata.

We may need to revise this approach but lets try the simple thing and see how it goes.

providers/summary/scancode.js
```diff
@@ -90,12 +85,49 @@ class ScanCodeSummarizer {
...
+        return declaredLicenses
+      }
+    }
+    // Did not find license file so returning empty set (no declared license)
+    return declaredLicenses
+  }
+
+  _summarizePackageInfo(files) {
+    const packageLicenses = new Set()
+    for (const fileIdx in files) {
```
jeffmcaffer: suggest `forEach` for looping over arrays (here and elsewhere) when you do not need the index.

providers/summary/scancode.js
```diff
@@ -90,12 +85,49 @@ class ScanCodeSummarizer {
...
+    const declaredLicenses = new Set()
+    for (const fileIdx in files) {
+      const file = files[fileIdx]
+      const pathArray = file.path.split('/')
+      const baseName = last(pathArray)
+      const isLicense = ['license', 'license.txt', 'license.md', 'license.html'].includes(baseName.toLowerCase())
+      if (isLicense && file.licenses) {
+        // Find first license file and return
+        file.licenses.forEach(license => declaredLicenses.add(license.spdx_license_key))
+        return declaredLicenses
```
jeffmcaffer: should return a license expression I believe.

providers/summary/scancode.js
```diff
@@ -90,12 +85,49 @@ class ScanCodeSummarizer {
...
+  _summarizeDeclaredLicenseInfo(files) {
+    const declaredLicenses = new Set()
+    for (const fileIdx in files) {
+      const file = files[fileIdx]
+      const pathArray = file.path.split('/')
+      const baseName = last(pathArray)
+      const isLicense = ['license', 'license.txt', 'license.md', 'license.html'].includes(baseName.toLowerCase())
+      if (isLicense && file.licenses) {
+        // Find first license file and return
+        file.licenses.forEach(license => declaredLicenses.add(license.spdx_license_key))
```
jeffmcaffer: consider 

```javascript
return this._toExpression(file.licenses.map(licence => license.spdx_license_key))
```
and then returning null in the final case where nothing is found.

providers/summary/scancode.js
```diff
@@ -90,12 +85,49 @@ class ScanCodeSummarizer {
...
+
+  _summarizePackageInfo(files) {
+    const packageLicenses = new Set()
+    for (const fileIdx in files) {
+      const file = files[fileIdx]
+      const asserted = get(file, 'packages[0].asserted_licenses')
+      if (asserted) {
+        // File first package file and return
+        this._addArrayToSet(asserted, packageLicenses, license => license.license || license.spdx_license_key)
+        return packageLicenses
```
jeffmcaffer: Similarly here, we should return the license expression and null if none is found

providers/summary/scancode.js
```diff
@@ -90,12 +85,49 @@ class ScanCodeSummarizer {
...
+      }
+    }
+    // Did not find package file
+    return packageLicenses
+  }
+
+  _summarizeFileInfo(files) {
+    const fileArray = []
```
jeffmcaffer: use `map`

```javascript
return files.map(file => {
   ...
   return { path: file.path, license: licenseExpression, attributions: file.copyrights }
}
```

PR #140, opened 2018-06-12
Yaml line width

Teju-Manchenella:
Changed yaml.safeDump lineWidth attribute so each yaml key and value fits on one line


jeffmcaffer:
change included in other commit

PR #154, opened 2018-07-09
Nuget support

Teju-Manchenella:
add nuget api calls to search for packages and versions. extracted date for nuget packages in summarizer.


app.js
```diff
@@ -109,6 +109,7 @@ app.use('/', index)
 app.use('/origins/github', require('./routes/originGitHub')())
 app.use('/origins/npm', require('./routes/originNpm')())
 app.use('/origins/maven', require('./routes/originMaven')())
+app.use('/origins/nuget', require('./routes/originNuget')())
```
jeffmcaffer: nit: NuGet is NuGet, not Nuget. Goofy but well, lets be consistent.

Please do a search and fix other instances as well.

PR #156, opened 2018-07-12
Pypi support

Teju-Manchenella:
Added package/version search and updated schema to accept pypi


routes/originPyPi.js
```diff
@@ -0,0 +1,35 @@
...
+const router = require('express').Router()
+const requestPromise = require('request-promise-native')
+
+router.get(
+  '/:namespace?/:name/revisions',
+  asyncMiddleware(async (request, response) => {
+    const { name } = request.params
+    const url = `https://pypi.python.org/pypi/${name}/json`
+    const answer = await requestPromise({ url, method: 'GET', json: true })
+    const result = answer && answer.releases ? Object.keys(answer.releases) : []
```
jeffmcaffer: are there any issues with volume here? for example, if there are 1000 releases, does pypi page the results?

routes/originPyPi.js
```diff
@@ -0,0 +1,35 @@
...
+    result.reverse()
+    return response.status(200).send(result)
+  })
+)
+
+router.get(
+  '/:namespace/:name?',
+  asyncMiddleware(async (request, response) => {
```
jeffmcaffer: Where is name used here? It should be the partial value being searched for.

routes/originPyPi.js
```diff
@@ -0,0 +1,35 @@
...
+// SPDX-License-Identifier: MIT
+
+const asyncMiddleware = require('../middleware/asyncMiddleware')
+const router = require('express').Router()
+const requestPromise = require('request-promise-native')
+
+router.get(
+  '/:namespace?/:name/revisions',
+  asyncMiddleware(async (request, response) => {
+    const { name } = request.params
```
jeffmcaffer: Does pypi have namespaces? Seems like it might based on the other route below that uses `namespace`. If so, how does the namespace factor in here?

routes/originPyPi.js
Teju-Manchenella: From my knowledge no this is not a problem. The request returns an object with every single release associated with it. I am yet to run into this case and there is little to no api documentation.

craigez: I checked a few packages and it looks like it returns all the releases, no pagination.

routes/originPyPi.js
craigez: I think there is no namespace, just a name. For example a valid URL is: https://pypi.org/pypi/billiard/json

routes/originPyPi.js
craigez: I don't think it has namespaces.

routes/originPyPi.js
jeffmcaffer: OK. Interestingly confusing they do have something called `namespace packages` but that appears to be more a way of sharding packages and joining them into one at runtime. 
So @Teju-Manchenella if there are no namespaces, lets remove the namespace code from the pypi support.

Issue #161, opened 2018-07-15
Change to use copyright statement, not holder


jeffmcaffer:
Initially we used the copyright holder (as defined by ScanCode) as the Attributions value in the ClearlyDefined definition. To better support scenarios like NOTICE file generation, we want to change to surface the copyright **statement** as discovered by the various tools. This should be a relatively simple change to `_collectAttributions` in `scancode.js`.

Should also look to see if there is anywhere else that summarizes attributions.

Having done this we will need to flush all existing definitions and either agressively recompute or let them lazily recompute.

fossygirl:
Closing as #166 was merged.

PR #166, opened 2018-07-16
Use copyright statements when getting attributions

cco3:
Rather than using the copyright holders as the attributions value, this
change uses the copyright statements in order to better support
scenarios like NOTICE file generation.

#161


cco3:
@jeffmcaffer

cco3:
Sorry...still need to fix the tests.

cco3:
OK, this should be ready now.

PR #169, opened 2018-07-16
Validation

Teju-Manchenella:
Updated the curation schema. Each time the curation endpoint is hit, each patch within the request body is checked against the curation.json schema. If there are no errors then the curation is processed and a pull request is created on GitHub. If there are errors, a 400 error is thrown and the invalid fields of the curation are returned.

I think the curation schema should be an array of curations, rather than a single curation object. Currently the curation endpoint gets a request with a patches field. The patches field is an array where each element in the array is a curation on a specific package. Each element in the patches field array is individually checked against the curation.json schema and if any element fails, then a 400 error is thrown and no curation occurs.

Added some testing :)


jeffmcaffer:
Regarding the schema being an array of curations... Having it be a single curation allows us to check just one curation as necessary. We could have another schema for the POST body that is the array of curation schema objects if that makes sense.

Issue #163, opened 2018-07-15
Add the ClearlyDefined score to definitions


jeffmcaffer:
Currently the ClearlyDefined score computation (for what it is) is done in the UI. This is not the right place as the business logic is not shareable and the value is recomputed over and over. This logic should be moved to the service and computed after all other summarization etc is completed (e.g., towards the end of `definitionService.compute`)

Note that the score should be computed and recorded for each facet separately. This enables scenarios where users want to consider only certain facets and ignore the others. The observer can aggregate (and indeed weight) the facet scores as they please. By default we will talk about the `core` facet score as the component's overall score.

The `score` property will need to be added to the definition schema but not the `curation` schema (the score is computed and is not editable)

Once this work is done we need to queue up some work in the website to stop computing the score there

jeffmcaffer:
done

PR #170, opened 2018-07-16
Adding top level score to definition

gauntface:
After discussing how to get badges working on the site (avoiding the clearly defined redirect and using shields.io directly on the site), it made sense to add a score to the definition, which this PR does.

Things to note with this PR:
- This PR will be required for the [website PR to work](https://github.com/clearlydefined/website/pull/217).
- A negative of this implementation is that the logic for the shield to use is duplicated on server and client side.

Starts: https://github.com/clearlydefined/service/issues/163
Partly Addresses: https://github.com/clearlydefined/website/issues/138


business/badgeCalculator.js
```diff
@@ -12,23 +12,14 @@ const scoreToUrl = {
   1: 'https://img.shields.io/badge/ClearlyDefined-1-yellow.svg',
   2: 'https://img.shields.io/badge/ClearlyDefined-2-brightgreen.svg'
 }
+
```
jeffmcaffer: Can this be removed completely now? the new browser code just computes this. the only reason we were doing it here was to enable the browser to ask the service. That is no longer the case.

Note that also means we can remove the tests and the route

test/business/badgeCalculator.js
```diff
@@ -1,48 +0,0 @@
-// Copyright (c) The Linux Foundation and others. Licensed under the MIT license.
-// SPDX-License-Identifier: MIT
-
```
jeffmcaffer: This test is still relevant. it just becomes a test of `definitionService.computeScore`

test/routes/badges.js
```diff
@@ -15,8 +15,11 @@ describe('Badge Route', () => {
         revision: '351396f971280ab79faddcf9782ea50f4e88358d'
       }
     }
```
jeffmcaffer: This whole test can go away if we remove the badge calculator

Issue #162, opened 2018-07-15
Clean up copyrights


jeffmcaffer:
Many of the copyright tools basically identify copyright lines in various files and then include them verbatim in their output. As a result we end up with lots of "special chars" and obvious goofiness. For example

* &lt;jmcaffer@microsoft.com&gt; (or some variation) -- at the very least we should convert the &** to the appropriate real char. We should also consider the merits of disclosing email addresses (though if they were in the code...)
* /r/n/...  These quoted chars often end up in strings. We could replace with spaces, truncate the string there (assuming they are not leading), other...
* strings with complete gibberish. these could just be dropped

So during summarization (pragmatically in `scancode.js` `_collectAttributions` we should invoke a generic string clean up mechanism. The mechanism can be improved over time as we see more and more variations.

jeffmcaffer:
There is always more work we could do here but #171 is a good start. We will see more as more contributors get involved and give feedback

PR #171, opened 2018-07-16
Handle special characters in copyright attribution

cco3:
This change handles html entities and literal instances of `\n` that
appear in the copyright attribution by replacing them with more
appropriate strings.

Issue: #162 
Dependency: #166

### DIFF
--- a/providers/summary/scancode.js
+++ b/providers/summary/scancode.js
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
 // SPDX-License-Identifier: MIT
 
+const he = require('he')
 const { get, remove, set, first, pullAllWith, isEqual } = require('lodash')
 const minimatch = require('minimatch')
 const { extractDate } = require('../../lib/utils')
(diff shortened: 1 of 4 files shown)


cco3:
@jeffmcaffer

providers/summary/scancode.js
```diff
@@ -153,9 +154,16 @@ class ScanCodeSummarizer {
...
     if (!copyrights || !copyrights.length) return null
-    return this._setToArray(copyrights.reduce((result, entry) => this._addArrayToSet(entry.holders, result), new Set()))
+    const set = new Set()
+    for (let copyright of copyrights) {
+      if (!copyright.statements) continue
+      const statements = copyright.statements.map(statement => he.decode(statement)
+        .replace(/\\[nr]/, '\n'))
+      this._addArrayToSet(statements, set)
+    }
+    return this._setToArray(set)
```
jeffmcaffer: Good for now. Going forward there are some additional cleanups we can do like normalizing punctuation at the end of the line, removing duplicate white spaces, trimming both ends of the string, ...

providers/summary/scancode.js
```diff
@@ -153,9 +154,16 @@ class ScanCodeSummarizer {
...
 
-  _collectAttributions(copyrights, holders) {
+  _collectAttributions(copyrights) {
     if (!copyrights || !copyrights.length) return null
-    return this._setToArray(copyrights.reduce((result, entry) => this._addArrayToSet(entry.holders, result), new Set()))
+    const set = new Set()
+    for (let copyright of copyrights) {
+      if (!copyright.statements) continue
+      const statements = copyright.statements.map(statement => he.decode(statement)
+        .replace(/\\[nr]/, '\n'))
```
jeffmcaffer: suspect we should just remove these. that is, each statement should be one line. Actually, in that case, we should also remove real CR and LF instances so the statement is on one line. Perhaps replace with a space and then overall remove multiple whiltespace.

providers/summary/scancode.js
jeffmcaffer: Does the call need to change as we are now returning the array rather than modifying the `holders` array that was passed in?

cco3: Acknowledged.

providers/summary/scancode.js
cco3: Done.

providers/summary/scancode.js
cco3: I think it should still be the same (we still want to build a deduped list of statements and return an array)

PR #173, opened 2018-07-17
View pr

daniellandau:
Related to https://github.com/clearlydefined/website/pull/219

### DIFF
--- a/providers/curation/github.js
+++ b/providers/curation/github.js
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
 // SPDX-License-Identifier: MIT
 
-const { concat, get, forIn, merge, set } = require('lodash')
+const { concat, get, forIn, merge, set, isEqual, uniq } = require('lodash')
 const base64 = require('base-64')
 const extend = require('extend')
 const { exec } = require('child_process')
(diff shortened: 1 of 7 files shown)


routes/curations.js
```diff
@@ -6,6 +6,12 @@ const express = require('express')
 const router = express.Router()
 const utils = require('../lib/utils')
 
+router.get('/base_url', (request, response) => {
+  const repo = request.app.locals.config.curation.store.github.repo
```
jeffmcaffer: suggest this be removed in favor of returning this value in the `/pr` route response. See other comments

routes/curations.js
```diff
@@ -18,6 +24,19 @@ router.get(
   })
 )
 
+// Get list of components included in a pr
+router.get(
+  '/pr/:pr',
+  asyncMiddleware(async (request, response) => {
+    return curationService.getPrComponents(request.params.pr).then(result => {
```
jeffmcaffer: Here add in the URL for the PR so we can avoid people constructing URLs. So the response might look like 

```javascript
{
  pr: 107,
  url: 'https://github.com/....',
  changes: [
    npm/npmjs/-/redie/0.3.0,
    git/github/microsoft/redie/082303205723abc324bd3e
  ]
}
```

That makes the response a bit more stand alone/complete

providers/curation/github.js
```diff
@@ -320,6 +328,25 @@ class GitHubCurationService {
     }
   }
 
+  async getPrComponents(number) {
```
jeffmcaffer: Nice. Some naming comments (naming is hard)

* getPrComponents => getChangedDefinitions
* result => files
* prComponents => changedCoordinates
* fileObj => remove and just files[i] in the one place
* prVersion => prDefinitions
* masterVersion => currentDefinitions
* relevantRevisions => changedRevisions

providers/curation/github.js
```diff
@@ -93,14 +93,23 @@ class GitHubCurationService {
         )
       )
     )
-    return (userGithub || serviceGithub).pullRequests.create({
+    const retVal = await (userGithub || serviceGithub).pullRequests.create({
```
jeffmcaffer: general naming comment. we tend to use whole words and semantic naming. So `retVal` => `result` and `commentObj` => `comment`

routes/curations.js
daniellandau: I think that depends on whether or not the base URL is useful elsewhere as well. In the currently disabled `PageCurate` that same URL is hard coded and could be replaced with an API call, but the `/curate` page only shows one component so it's not exactly the same and thus doesn't need the same full dataset either.

providers/curation/github.js
daniellandau: Naming updated.

routes/curations.js
jeffmcaffer: @daniellandau thoughts on this? I commented similarly in the corresponding website PR.

providers/curation/github.js
```diff
@@ -93,14 +93,23 @@ class GitHubCurationService {
...
       body: patch.description,
       head: `refs/heads/${prBranch}`,
       base: branch
     })
+    const number = result.data.number
+    const comment = {
+      owner,
+      repo,
+      number,
+      body: `You can review the change introduced to the full definiton at [ClearlyDefined](https://clearlydefined.io/view_pr/${number}).`
```
jeffmcaffer: this url changed. It is now 'contributions'

also, type in 'definition' (missing the last i)

jeffmcaffer: Note that I merged this PR and fixed the above issue

PR #177, opened 2018-07-29
Add FOSSology summarization

larainema:
This commit to add FOSSology summarization, to decode and parse nomos
output.

### DIFF
--- a/providers/summary/fossology.js
+++ b/providers/summary/fossology.js
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 const { set } = require('lodash')
+const { setIfValue } = require('../../lib/utils')
 const base64 = require('base-64')
 
 class FOSSologySummarizer {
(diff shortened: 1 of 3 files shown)


providers/summary/fossology.js
```diff
@@ -0,0 +1,48 @@
...
+   * @param {EntitySpec} coordinates - The entity for which we are summarizing
+   * @param {*} harvested - the set of raw tool ouptuts related to the idenified entity
+   * @param {Facets} facets - an object detailing the facets to group by.
+   * @returns {Definition} - a summary of the given raw information
+   */
+  summarize(coordinates, harvested, facets = {}) {
+    if (!harvested || !harvested.nomos || !harvested.nomos.version)
+      throw new Error('Not valid FOSSology data')
+
+    const result = {}
```
jeffmcaffer:
```javascript
return { nomos: this._summarizeNomosLicenseInfo(harvested.nomos.output.content) }
```

providers/summary/fossology.js
```diff
@@ -0,0 +1,48 @@
...
+  _setIfValue(target, path, value) {
+    if (!value) return
+    set(target, path, value)
+  }
+
+  _summarizeNomosLicenseInfo(content) {
+    const nomosOutput = base64.decode(content)
+    const files = nomosOutput.split("\n")
+    return files.map(file => {
+      const path = file.replace(/File/g, ' ').replace(/contains.+/g, ' ').trim()
```
jeffmcaffer: can these regex be simplified? Seems like we are really just removing sections of the string. Alternatively, the regex could extract the interesting information using `match`

jeffmcaffer:
This is not quite right but it points out some opportunities for simplification in the general case. I started a small refactor to move some code up from the ScanCode summarizer so it can be reused here. Once that is done, this can restructured a little to match. Should be soon and a small amount of change here.

jeffmcaffer:
Thanks @larainema

Issue #178, opened 2018-07-31
Fix addNpmData() in summary/clearlyDefined.js


Teju-Manchenella:
According to npmjs documentation (https://docs.npmjs.com/files/package.json#license) old packages used to store licenses as an array of license objects. This style has now been deprecated; however,  some packages might still have this deprecated style of licenses. A check should be placed in addNpmData() i to check the type of the license field in the registry data and get the correct SPDX license from it.

Teju-Manchenella:
Closed PR #179

PR #179, opened 2018-07-31
npm license type fix

Teju-Manchenella:
Addressing issue #178. Based on https://docs.npmjs.com/files/package.json#license there can be a license object where the spdx license is stored in the type field.


providers/summary/clearlydefined.js
```diff
@@ -65,9 +65,15 @@ class ClearlyDescribedSummarizer {
     result.described.projectWebsite = data.registryData.manifest.homepage
     result.described.issueTracker = data.registryData.manifest.bugs
     result.described.releaseDate = extractDate(data.registryData.releaseDate)
-    result.licensed = {
-      declared: data.registryData.manifest.license
-    }
+    const license = data.registryData.manifest.license
```
jeffmcaffer: Suggest a simplification like 
```javascript
    if (license) result.licensed = { declared: typeof license === 'string' ? license : license.type }
```

Issue #157, opened 2018-07-12
case sensitivity


jeffmcaffer:
One thing that @geneh and I noticed:  `EntityCoordinates` has a (bogusly named) list of providers and parts their coordinates that can/should be normalized. See https://github.com/clearlydefined/service/blob/master/lib/entityCoordinates.js#L9. When adding a new package manager, we should update that. The idea here is that we need to know what parts are case-sensitive and which are not.

Note: In reality the values in the object should indicate 

1. case sensitive, 
2. case insensitive but case preserving
3, case insensitive and non-preserving

Then we can use that to handle the incoming and outgoing paths and coordinates appropriately.

geneh:
In order to support this requirement, the following needs to change:

### Crawler:
In https://github.com/Microsoft/ghcrawler/blob/jm/apicrawler/providers/storage/storageDocStore.js: Add normalizeNames parameter in both _getBlobNameFromUrl and _getBlobPathFromUrn which would be set to false by default. If set to false, toLowerCase() should not be called, otherwise it should as in the current behavior. The parameter should be configurable through the passed in options into the constructor.
After this change we'll have to be very careful about URL's namespace, name and revision casing, especially when queueing components up manually.

### Service:
https://github.com/clearlydefined/service/blob/master/lib/entityCoordinates.js: Change normalize() as described above or better always preserve namespace, name and revision casing regardless of case sensitivity by removing normalize method.
The output and not the input should always be returned as is.

Since Azure search is case insensitive, no code changes should be required for browsing but all the data in blob storage and Azure search will have to be deleted and then re-crawled (top 10,000 components for each supported type). Memory search should be updated accordingly in https://github.com/clearlydefined/service/blob/master/providers/search/memory.js.

### Website:
Update https://github.com/clearlydefined/website/blob/master/src/utils/entitySpec.js similarly to the one in the service.

#### Testing notes:
Test against both file and Azure blob stores as well as in-memory and Azure search.
Make sure Browse, Inspect, and Harvest work as expected by preserving the casing in search results.

PR #180, opened 2018-08-01
Case sensitivity

Teju-Manchenella:
New toLowerCase map across website, service, and crawler (ignore changes in clearlydefined.js it is part of another PR in an other branch).  Addresses issue #[237](https://github.com/clearlydefined/service/issues/157)

### DIFF
--- a/lib/entityCoordinates.js
+++ b/lib/entityCoordinates.js
@@ -7,10 +7,13 @@ const REVISION = 0x1
 const NONE = 0
 
 const toLowerCaseMap = {
-  github: NAMESPACE | NAME,
-  npmjs: NONE,
-  mavencentral: NONE,
-  mavencentralsource: NONE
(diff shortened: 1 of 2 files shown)


jeffmcaffer:
Note that this changes some pretty fundamental settings for how identity is managed. We likely need to either clear out all the data and reharvest to get the right paths in storage or figure out a way to update any incorrect paths. For example, all the NuGet paths are likely wrong. Perhaps we only do the ones that have been affected?

Teju-Manchenella:
I have opened a new issue [#123](https://github.com/clearlydefined/crawler/issues/123) in the crawler that addresses renaming blob storage. I discussed this with Gene this morning and he wants to also look at GHCrawler before attempting to rename the storage paths.

jeffmcaffer:
subsumed by #198

PR #182, opened 2018-08-07
refactor facet computation

jeffmcaffer:
The previous notion of facets and grouping data into facets was largely coded into the scancode summarizer. With a recent redesign that introduced per-file data  such as license and copyright, the summarizers can be simplified and the facet grouping moved up a level or two and thus shared. 

This refactoring moves all the file grouping code up to the definition service and rationalizes the related tests.

Note that this does not change the definition schema so should be mergable without any data migrations.

### DIFF
--- a/business/definitionService.js
+++ b/business/definitionService.js
@@ -3,9 +3,11 @@
 
 const Readable = require('stream').Readable
 const throat = require('throat')
-const { get, union, remove } = require('lodash')
+const { get, union, remove, pullAllWith, isEqual } = require('lodash')
 const EntityCoordinates = require('../lib/entityCoordinates')
 const { setIfValue, setToArray, addArrayToSet } = require('../lib/utils')
+const minimatch = require('minimatch')
+const he = require('he')
 
 class DefinitionService {
   constructor(harvest, summary, aggregator, curation, store, search) {
@@ -130,7 +132,7 @@ class DefinitionService {
     this._ensureFacets(coordinates, definition)
     this._ensureCurationInfo(definition, curation)
     this._ensureSourceLocation(coordinates, definition)
-    this._ensureCoordinates(coordinates, definition)
+    this._ensureCoordinates(definition)
     definition.score = this.computeScore(definition)
     return definition
   }
@@ -178,7 +180,7 @@ class DefinitionService {
   }
 
   // ensure all the right facet information has been computed and added to the given definition
-  _ensureFacets(coordinates, definition) {
+  _ensureFacets(definition) {
     if (!definition.files) return
     const facetFiles = this._computeFacetFiles([...definition.files], definition.described.facets)
     for (const facet in facetFiles)
@@ -212,7 +214,8 @@ class DefinitionService {
     // accummulate all the licenses and attributions, and count anything that's missing
     for (let file of facetFiles) {
       file.license ? licenseExpressions.add(file.license) : unknownLicenses++
-      file.attributions ? addArrayToSet(file.attributions, attributions) : unknownParties++
+      const statements = this._simplifyAttributions(file.attributions)
+      statements ? addArrayToSet(statements, attributions) : unknownParties++
       if (facet !== 'core') {
         // tag the file with the current facet if not core
         file.facets = file.facets || []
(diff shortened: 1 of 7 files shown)


business/definitionService.js
```diff
@@ -123,13 +126,13 @@ class DefinitionService {
...
-    const facets = await this._computeFacets(coordinates, raw, curation)
-    const summarized = await this.summaryService.summarizeAll(coordinates, raw, facets || {})
+    const summarized = await this.summaryService.summarizeAll(coordinates, raw)
     const aggregated = await this.aggregationService.process(coordinates, summarized)
     const definition = await this.curationService.apply(coordinates, curation, aggregated)
+    this._ensureFacets(coordinates, definition)
     this._ensureCurationInfo(definition, curation)
     this._ensureSourceLocation(coordinates, definition)
-    this._ensureCoordinates(coordinates, definition)
+    this._ensureCoordinates(definition)
```
dabutvin: does this function need to be updated below for the signature change?

```
_ensureCoordinates(coordinates, definition) {
    definition.coordinates = EntityCoordinates.fromObject(coordinates)
  }
```

business/definitionService.js
```diff
@@ -123,13 +126,13 @@ class DefinitionService {
   async compute(coordinates, curationSpec) {
     const curation = await this.curationService.get(coordinates, curationSpec)
     const raw = await this.harvestService.getAll(coordinates)
-    const facets = await this._computeFacets(coordinates, raw, curation)
-    const summarized = await this.summaryService.summarizeAll(coordinates, raw, facets || {})
+    const summarized = await this.summaryService.summarizeAll(coordinates, raw)
     const aggregated = await this.aggregationService.process(coordinates, summarized)
     const definition = await this.curationService.apply(coordinates, curation, aggregated)
+    this._ensureFacets(coordinates, definition)
```
dabutvin: `_ensureFacets` only accepts `definition` as a parameter

business/definitionService.js
```diff
@@ -176,22 +179,76 @@ class DefinitionService {
...
+    const attributions = new Set()
+    const licenseExpressions = new Set()
+    let unknownParties = 0
+    let unknownLicenses = 0
+    // accummulate all the licenses and attributions, and count anything that's missing
+    for (let file of facetFiles) {
+      file.license ? licenseExpressions.add(file.license) : unknownLicenses++
+      const statements = this._simplifyAttributions(file.attributions)
+      statements ? addArrayToSet(statements, attributions) : unknownParties++
+      if (facet !== 'core') {
```
dabutvin: should we bother checking for 'core' here even though we removed 'core' in `_computeFacetFiles` just before this method?

business/definitionService.js
```diff
@@ -176,22 +179,76 @@ class DefinitionService {
...
-    const summarized = await this.summaryService.summarizeFacets(coordinates, raw)
-    const aggregated = await this.aggregationService.process(coordinates, summarized)
-    const definition = await this.curationService.apply(coordinates, curation, aggregated)
-    return get(definition, 'described.facets')
+  // ensure all the right facet information has been computed and added to the given definition
+  _ensureFacets(definition) {
+    if (!definition.files) return
+    const facetFiles = this._computeFacetFiles([...definition.files], definition.described.facets)
+    for (const facet in facetFiles)
+      setIfValue(definition, `licensed.facets.${facet}`, this._summarizeFacetInfo(facet, facetFiles[facet]))
```
dabutvin: I wonder if it would simplify to call `_summarizeFacetInfo` within the loop in `_computeFacetFiles` since it modifies the facets in place

so it would look something like

```
_computeFacetFiles(files, facets = {}) {
    const facetList = Object.getOwnPropertyNames(facets)
    remove(facetList, 'core')
    if (facetList.length === 0) return { core: files }
    const result = { core: [...files] }
    for (const facet in facetList) {
      const facetKey = facetList[facet]
      const filters = facets[facetKey]
      if (!filters || filters.length === 0) break
      const facetFile = files.filter(file => filters.some(filter => minimatch(file.path, filter)));
      result[facetKey] = this._summarizeFacetInfo(facetKey, facetFile)
      pullAllWith(result.core, result[facetKey], isEqual)
    }
    return result
  }
```

business/definitionService.js
jeffmcaffer: D'oh figured it out. the tests are all calling `_ensureFacets` directly. That was a bit of a hack but I didn't want to setup and run through the whole service. lazy. I will update the tests to use more mocks and run the whole service.

test/business/definitionServiceTest.js
```diff
@@ -241,3 +232,16 @@ function buildFile(path, license, holders) {
   setIfValue(result, 'attributions', holders ? holders.map(entry => `Copyright ${entry}`) : null)
   return result
 }
+
+function createService(definition, curation) {
```
dabutvin: 👍

PR #193, opened 2018-08-12
Handle and expose interesting file content

jeffmcaffer:
The crawler now captures interesting files (e.g., LICENSE.md, ...) and stores it with an opaque token (sha256 hash of the raw content). This PR adds the following to support this new data.

* API for getting the content given a token
* extraction of the relevant harvested data during summarization 
* proper merging of the file tokens during aggregation and curation
* More tests

Note that the store access technique is replicated from how we are handling the definition and harvest stores. That is, we directly access the underlying storage shared with the crawler. This may prove to be an issue in the future as we are sharing implementation details. As it is, the storage structure is pretty stable. 

Note as well that this includes a subtle change in a little used part of the schema. The `files[n].content` property changed to be called `files[n].token`. The term content was misleading as the value is a pointer to the content not the content itself.

### DIFF
--- a/app.js
+++ b/app.js
@@ -46,9 +46,6 @@ const curationService = require(`./providers/curation/${curationProvider}`)(
 )
 const curations = require('./routes/curations')(curationService)
 
-const contentStoreProvider = config.content.store.provider
-const contentStore = require(`./providers/stores/${contentStoreProvider}`)(config.content.store[contentStoreProvider])
-
 const definitionStoreProvider = config.definition.store.provider
 const definitionStore = require(`./providers/stores/${definitionStoreProvider}`)(
   config.definition.store[definitionStoreProvider]
@@ -71,7 +68,7 @@ const definitionService = require('./business/definitionService')(
 // Circular dependency. Reach in and set the curationService's definitionService. Sigh.
 curationService.definitionService = definitionService
 
-const content = require('./routes/content')(contentStore)
+const attachments = require('./routes/attachments')(harvestStore)
 const definitions = require('./routes/definitions')(harvestStore, curationService, definitionService)
 
 const appLogger = console // @todo add real logger
@@ -125,7 +122,7 @@ app.use('/harvest', harvest)
 app.use(bodyParser.json())
 app.use('/curations', curations)
 app.use('/definitions', definitions)
-app.use('/content', content)
+app.use('/attachments', attachments)
 
 // catch 404 and forward to error handler
 const requestHandler = (req, res, next) => {
--- a/app.js
+++ b/app.js
@@ -1,5 +1,11 @@
 // Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
 // SPDX-License-Identifier: MIT
+
+// TODO consider putting this in for real
+// process.on('unhandledRejection', (reason, p) => {
+//   throw reason
+// })
+
 const express = require('express')
 const logger = require('morgan')
 const bodyParser = require('body-parser')
@@ -40,6 +46,9 @@ const curationService = require(`./providers/curation/${curationProvider}`)(
 )
 const curations = require('./routes/curations')(curationService)
 
+const contentStoreProvider = config.content.store.provider
+const contentStore = require(`./providers/stores/${contentStoreProvider}`)(config.content.store[contentStoreProvider])
+
 const definitionStoreProvider = config.definition.store.provider
 const definitionStore = require(`./providers/stores/${definitionStoreProvider}`)(
   config.definition.store[definitionStoreProvider]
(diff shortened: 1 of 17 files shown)


jeffmcaffer:
Ready to merge. @dabutvin you have any comments?

app.js
```diff
@@ -1,5 +1,11 @@
 // Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
 // SPDX-License-Identifier: MIT
+
+// TODO consider putting this in for real
```
dabutvin: 👍

package.json
```diff
@@ -29,6 +29,7 @@
     "express-init": "^1.1.0",
     "express-rate-limit": "^2.11.0",
     "extend": "^3.0.1",
+    "glob": "^7.1.2",
```
dabutvin: is this required?

lib/utils.js
```diff
@@ -67,6 +68,21 @@ function addArrayToSet(array, set, valueExtractor) {
   return set
 }
 
+// merge the given definition onto the base definition. Be careful to handle various arrays
+// correctly by finding matching entried and merging them as appropriate.
+function merge(base, newDefinition) {
```
dabutvin: we may want to update the name of this or make the implementation more generic.

Maybe we could leave it called `merge()` and update to check all the props for `IsArray()` and merge everything

or maybe we could leave the implementation and rename it to `mergeFiles()` or something like that?

routes/attachments.js
```diff
@@ -0,0 +1,28 @@
...
+// Get a proposed patch for a specific revision of a component
+router.get('/:token', asyncMiddleware(getAttachment))
+
+async function getAttachment(request, response) {
+  // TODO this is a hack. The coordinates are not intended to be sparse. Here make sure the
+  // token is the "namespace" to avoid the '-' showing up. Fix is likely to have an explicit
+  // AttachmentCoordinates class that has the right shape and generalize the stores to just take
+  // "coordinates" and not worry about how they are structured.
+  const coordinates = new ResultCoordinates('attachment', null, request.params.token)
+  const result = await harvestStore.get(coordinates)
```
dabutvin: we should probably pass a stream to the store as a 2nd parameter.  Otherwise all attachments will get parsed as JSON

azblob looks like this:

```
    if (stream)
      return new Promise((resolve, reject) => {
        this.blobService.getBlobToStream(this.containerName, name, stream, responseOrError(resolve, reject))
      })
    return new Promise((resolve, reject) => {
      this.blobService.getBlobToText(this.containerName, name, resultOrError(resolve, reject))
    }).then(
      result => JSON.parse(result),
      error => {
        if (error.statusCode === 404) return null
        throw error
      }
    )

```

dabutvin: ohhhh i see - the attachment is a property of the JSON blob. Should we be worried about contentType at all? Or is this only for text attachments?

routes/swagger.yaml
```diff
@@ -285,21 +285,21 @@ paths:
...
           name: token
           schema:
             type: string
           required: true
-          description: The token identifying the content to access
+          description: The token identifying the attachment to access
       responses:
         200:
-          description: The raw content corresponding to the given token.
+          description: The raw contne of the attachment corresponding to the given token.
```
iamwillbar: Type: content

routes/attachments.js
```diff
@@ -7,22 +7,22 @@ const router = express.Router()
...
   // token is the "namespace" to avoid the '-' showing up. Fix is likely to have an explicit
-  // ContentCoordinates class that has the right shape and generalize the stores to just take
+  // AttachmentCoordinates class that has the right shape and generalize the stores to just take
   // "coordinates" and not worry about how they are structured.
-  const coordinates = new ResultCoordinates('content', null, request.params.contentToken)
-  const result = await contentStore.get(coordinates)
-  if (result) return response.status(200).send(result.content)
+  const coordinates = new ResultCoordinates('attachment', null, request.params.token)
+  const result = await harvestStore.get(coordinates)
+  if (result) return response.status(200).send(result.attachment)
```
iamwillbar: Small style question, I normally try and keep the happy path on the "root" of the method and have the guard causes bail out (because often there's more bail out conditions then success conditions), so I would have written this as:

```javascript
if (!result) return response.sendStatus(404)
return response.status(200).send(result.attachment)
```

lib/utils.js
jeffmcaffer: renamed to mergeDefinitions

package.json
jeffmcaffer: Huh. Interestingly no. Good catch. Must be left over from some iteration of the code.

routes/attachments.js
jeffmcaffer: Don't expect complex files here. simple text. md and html at best

jeffmcaffer:
Thanks for the review feedback @iamwillbar and @dabutvin . I've addressed all of it so think we are good to go. Can;t push merge right now but will soon

PR #197, opened 2018-08-14
compute implicit and curated scores

jeffmcaffer:
The current scoring tells users the score of the entire definition and includes any curations. However, 
* Tome users will only be interested in (or emphasize) particular dimensions of the definition (e.g., licensed). For example, they may not care that the source location is not known.
* Project teams should be interested in how inherently ClearlyDefined their project is. The community should be able to tell the difference between a project has no data whatsoever but has been lovingly curated by an attentive user community, and one which is loved and clarified by the project team.

To address these scenarios, this PR splits the score computation into two (using just the harvested data and using the final, curated definition) and then does this for each dimension (described and licensed) separately. The resultant scores are stored in their respective dimension as `toolScore` and `score`.

The top level score is removed in this model. UI's presenting definitions and users assessing definitions are free to interpret and reveal these scores however most makes sense to their user base.

### DIFF
--- a/business/definitionService.js
+++ b/business/definitionService.js
@@ -3,11 +3,12 @@
 
 const Readable = require('stream').Readable
 const throat = require('throat')
-const { get, union, remove, pullAllWith, isEqual } = require('lodash')
+const { get, set, union, remove, pullAllWith, isEqual } = require('lodash')
 const EntityCoordinates = require('../lib/entityCoordinates')
 const { setIfValue, setToArray, addArrayToSet } = require('../lib/utils')
 const minimatch = require('minimatch')
 const he = require('he')
+const extend = require('extend')
 
 class DefinitionService {
   constructor(harvest, summary, aggregator, curation, store, search) {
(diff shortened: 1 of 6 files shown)


business/definitionService.js
```diff
@@ -135,29 +136,64 @@ class DefinitionService {
...
     // For now it just checks that a license and copyright holders are present
-    const hasLicense = get(definition, 'licensed.declared')
-    const hasAttributionParties = get(definition, 'licensed.attribution.parties[0]')
-    if (hasLicense && hasAttributionParties) return 2
-    if (hasLicense || hasAttributionParties) return 1
-    return 0
+    let result = 0
+    result += !!get(definition, 'licensed.declared')
+    result += !!get(definition, 'licensed.facets.core.attribution.parties[0]')
+    return result
```
dabutvin: 👍 this is nice

should we divide by the max to give a percentage and/or grade letter?

business/definitionService.js
```diff
@@ -135,29 +136,64 @@ class DefinitionService {
...
+    }
+  }
+
+  // Given a definition, calculate and return the score for the described dimension
+  _computeDescribedScore(definition) {
+    // @todo we need to flesh this out
+    // For now it just checks that a few props are present
+    let result = 0
+    result += !!get(definition, 'described.releaseDate')
+    result += !!get(definition, 'described.sourceLocation')
```
dabutvin: might want to update to `result += !!get(definition, 'described.sourceLocation.url')`

no sense giving points if we can't actually get to the source code 😝

app.js
```diff
@@ -61,8 +67,9 @@ const definitionService = require('./business/definitionService')(
 )
 // Circular dependency. Reach in and set the curationService's definitionService. Sigh.
 curationService.definitionService = definitionService
+const definitions = require('./routes/definitions')(definitionService)
 
-const definitions = require('./routes/definitions')(harvestStore, curationService, definitionService)
+const attachments = require('./routes/attachments')(harvestStore)
```
dabutvin: It looks like this leaked in here from your other PR

PR #198, opened 2018-08-14
add metadata to store for preserving case

dabutvin:
pending updates before merge

- [x] push updated tests for this change
- [x] update to the crawler to preserve case on self link during harvesting (https://github.com/clearlydefined/crawler/pull/132)
- [x] update to the crawler to lowercase self link when determining storage path (https://github.com/Microsoft/ghcrawler/pull/143)
- [ ] update the store to add properly case all the existing blob metadata and selflinks
- [ ] review existing curations to ensure they have the right casing
- [x] `npm update ghcrawler` after merging ghcrawler
- [ ] reload the whole search index

### DIFF
--- a/providers/stores/azblob.js
+++ b/providers/stores/azblob.js
@@ -4,6 +4,7 @@
 const azure = require('azure-storage')
 const AbstractStore = require('./abstractStore')
 const EntityCoordinates = require('../../lib/entityCoordinates')
+const ResultCoordinates = require('../../lib/resultCoordinates')
 
 const resultOrError = (resolve, reject) => (error, result) => (error ? reject(error) : resolve(result))
 const responseOrError = (resolve, reject) => (error, result, response) => (error ? reject(error) : resolve(response))
@@ -77,13 +78,7 @@ class AzBlobStore extends AbstractStore {
       })
       result.entries.forEach(entry => {
         const urn = entry.metadata.urn
-        if (urn) {
-          const value = this._toPreservedCoordinatesFromResultsStoragePath(
-            entry.name,
-            EntityCoordinates.fromUrn(urn).toString()
-          )
-          if (value) list.add(value)
-        }
+        if (urn) list.add(ResultCoordinates.fromUrn(urn).toString())
       })
       continuation = result.continuationToken
     } while (continuation)
@@ -156,7 +151,7 @@ class AzBlobStore extends AbstractStore {
     })
   }
 
-  async store(coordinates, stream) {
+  store(coordinates, stream) {
     const name = this._toStoragePathFromCoordinates(coordinates) + '.json'
     return new Promise((resolve, reject) => {
       stream.pipe(
--- a/providers/stores/azblob.js
+++ b/providers/stores/azblob.js
@@ -4,7 +4,6 @@
 const azure = require('azure-storage')
 const AbstractStore = require('./abstractStore')
 const EntityCoordinates = require('../../lib/entityCoordinates')
-const ResultCoordinates = require('../../lib/resultCoordinates')
 
 const resultOrError = (resolve, reject) => (error, result) => (error ? reject(error) : resolve(result))
 const responseOrError = (resolve, reject) => (error, result, response) => (error ? reject(error) : resolve(response))
@@ -35,7 +34,7 @@ class AzBlobStore extends AbstractStore {
   async list(coordinates) {
     const list = new Set()
     let continuation = null
-    while (true) {
+    do {
       const result = await new Promise((resolve, reject) => {
         const name = this._toStoragePathFromCoordinates(coordinates)
         this.blobService.listBlobsSegmentedWithPrefix(
@@ -50,9 +49,9 @@ class AzBlobStore extends AbstractStore {
         const value = entry.metadata.coordinates
         if (value) list.add(value)
       })
-      if (!result.continuationToken) return Array.from(list).sort()
       continuation = result.continuationToken
-    }
+    } while (continuation)
+    return Array.from(list).sort()
   }
 
   /**
@@ -65,7 +64,7 @@ class AzBlobStore extends AbstractStore {
   async listResults(coordinates) {
     const list = new Set()
     let continuation = null
-    while (true) {
+    do {
       const result = await new Promise((resolve, reject) => {
         const name = this._toStoragePathFromCoordinates(coordinates)
         this.blobService.listBlobsSegmentedWithPrefix(
@@ -80,9 +79,9 @@ class AzBlobStore extends AbstractStore {
         const value = this._toPreservedCoordinatesFromResultsStoragePath(entry.name, entry.metadata.coordinates)
         if (value) list.add(value)
       })
-      if (!result.continuationToken) return Array.from(list).sort()
       continuation = result.continuationToken
-    }
+    } while (continuation)
+    return Array.from(list).sort()
   }
 
   /**
--- a/providers/stores/azblob.js
+++ b/providers/stores/azblob.js
@@ -46,8 +46,8 @@ class AzBlobStore extends AbstractStore {
         )
       })
       result.entries.forEach(entry => {
-        const value = entry.metadata.coordinates
-        if (value) list.add(value)
+        const urn = entry.metadata.urn
+        if (urn) list.add(EntityCoordinates.fromUrn(urn).toString())
       })
       continuation = result.continuationToken
     } while (continuation)
@@ -76,8 +76,14 @@ class AzBlobStore extends AbstractStore {
         )
       })
       result.entries.forEach(entry => {
-        const value = this._toPreservedCoordinatesFromResultsStoragePath(entry.name, entry.metadata.coordinates)
-        if (value) list.add(value)
+        const urn = entry.metadata.urn
+        if (urn) {
+          const value = this._toPreservedCoordinatesFromResultsStoragePath(
+            entry.name,
+            EntityCoordinates.fromUrn(urn).toString()
+          )
+          if (value) list.add(value)
+        }
       })
       continuation = result.continuationToken
     } while (continuation)
@@ -151,18 +157,13 @@ class AzBlobStore extends AbstractStore {
   }
 
   async store(coordinates, stream) {
-    const blobName = this._toStoragePathFromCoordinates(coordinates) + '.json'
-    const preservedName = EntityCoordinates.fromObject(coordinates).toString()
+    const name = this._toStoragePathFromCoordinates(coordinates) + '.json'
     return new Promise((resolve, reject) => {
       stream.pipe(
         this.blobService.createWriteStreamToBlockBlob(
           this.containerName,
-          blobName,
-          {
-            blockIdPrefix: 'block',
-            contentSettings: { contentType: 'application/json' },
-            metadata: { coordinates: preservedName }
-          },
+          name,
+          { blockIdPrefix: 'block', contentSettings: { contentType: 'application/json' } },
           responseOrError(resolve, reject)
         )
       )
--- a/providers/stores/azblob.js
+++ b/providers/stores/azblob.js
@@ -4,6 +4,7 @@
 const azure = require('azure-storage')
 const AbstractStore = require('./abstractStore')
 const EntityCoordinates = require('../../lib/entityCoordinates')
+const ResultCoordinates = require('../../lib/resultCoordinates')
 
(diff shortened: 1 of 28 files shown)


providers/stores/azblob.js
```diff
@@ -33,11 +34,13 @@ class AzBlobStore extends AbstractStore {
...
           name,
           continuation,
+          { include: azure.BlobUtilities.BlobListingDetails.METADATA },
           resultOrError(resolve, reject)
         )
       })
       result.entries.forEach(entry => {
-        const value = this._getEntry(entry.name, type)
+        const value =
+          type === 'entity' ? this._getEntry(entry.metadata.coordinates, type) : this._getEntry(entry.name, type)
```
jeffmcaffer: Just to be clear, only the Entities (i.e., definitions) will have `coordinates` in the blob metadata? Could/should we just put it in there all the time?

Note the new code in PR adds a new type of thing being stored in blobs. It currently adds a new instance of azblob using a different container (sort of how we do it for definitions) to store the interesting file content. These could be classified as `results` since they really do go with the actual tool results, we've just chosen to store them in separate files.

providers/stores/azblob.js
```diff
@@ -113,13 +116,20 @@ class AzBlobStore extends AbstractStore {
...
       stream.pipe(
         this.blobService.createWriteStreamToBlockBlob(
           this.containerName,
-          name,
-          { blockIdPrefix: 'block', contentSettings: { contentType: 'application/json' } },
+          blobName,
+          {
+            blockIdPrefix: 'block',
+            contentSettings: { contentType: 'application/json' },
+            metadata: {
```
jeffmcaffer: nit: suggest formatting on one line as with `contentString`

providers/stores/file.js
```diff
@@ -138,14 +151,19 @@ class FileStore extends AbstractStore {
...
     const filePath = this._toStoragePathFromCoordinates(coordinates) + '.json'
+    const preservedName = EntityCoordinates.fromObject(coordinates).toString()
     const dirName = path.dirname(filePath)
     await promisify(mkdirp)(dirName)
     return new Promise((resolve, reject) => {
-      const file = fs
-        .createWriteStream(filePath)
-        .on('finish', () => resolve())
+      const fileStream = fs.createWriteStream(filePath).on('error', error => reject(error))
+      const jsonStream = JSONStream.parse()
```
jeffmcaffer: Hmmm, nice way to handle this unfortunate situation.

providers/stores/file.js
```diff
@@ -138,14 +151,19 @@ class FileStore extends AbstractStore {
...
+        .on('data', data => {
+          data._metadata.coordinates = preservedName
+          fileStream.write(JSON.stringify(data))
+        })
+        .on('end', () => resolve())
```
jeffmcaffer: should we be resolving when the fileStream `finish`es? Seems like the JSONstream will `end` when `stream` is done piping to it but we may still not have written the file?

providers/stores/file.js
```diff
@@ -21,8 +24,18 @@ class FileStore extends AbstractStore {
...
-      paths.forEach(entry => {
-        const value = this._getEntry(entry, type)
+      const files = await Promise.all(
+        paths.map(path => {
+          return new Promise((resolve, reject) =>
+            fs.readFile(path, (error, data) => (error ? reject(error) : resolve({ path, contents: JSON.parse(data) })))
+          )
+        })
+      )
+      files.forEach(file => {
```
jeffmcaffer: Interesting, here I would have combined these two loops and just done something like

```javascript
return Promise.all(paths.map(path => {
  if (type !== entity) return this._getEntry(path, type).toString()
  return new Promise((resolve, reject) =>
    fs.readFile(path, (error, data) => {
        if (error) return reject(error)
        const contents = JSON.parse(data)
        resolve(this._getEntry(contents._metadata.coordinates , type)).toString()
}).filter(e => e))
```

and uniquified at the end.

providers/stores/azblob.js
```diff
@@ -33,11 +34,13 @@ class AzBlobStore extends AbstractStore {
           this.containerName,
           name,
           continuation,
+          { include: azure.BlobUtilities.BlobListingDetails.METADATA },
```
jeffmcaffer: is there much of a cost to getting the metadata? if it is not trivial, perhaps only add this option if `type === entity`

providers/stores/azblob.js
dabutvin: I figured we should put it on all blobs and stop referring to blob names whenever we want coordinates. The following is my current plan + understanding, please correct me if I am on the wrong track:

---

when you call `list(coordinates, 'entity')` you are looking for something that looks like

```
[ 'npm/npmjs/-/co/4.6.0' ]
```

this will respect casing

---

when you call `list(coordinates, 'result')` you are looking for something that looks like

```
[ 'npm/npmjs/-/co/4.6.0/clearlydefined/1, 'npm/npmjs/-/co/4.6.0/scancode/2.2.1' ]
```

this will always be lowercased, driven from the blob name itself

---

we will put the metadata on the blobs that live here:

harvested:
 - `npm/npmjs/-/co/revision/4.6.0/tool/clearlydefined/1.json`
 - `npm/npmjs/-/co/revision/4.6.0/tool/scancode/2.2.1.json`

cached:
 - `npm/npmjs/-/co/revision/4.6.0/tool/definition/1.json`

and that metadata will be `coordinates=npm/npmjs/-/co/4.6.0`

providers/stores/file.js
dabutvin: I will double check on that

providers/stores/file.js
dabutvin: Right! Only read the file if we have to. Will update this

providers/stores/azblob.js
dabutvin: I think we're only getting a few blobs at a time, but it's definetly not free.

I'll see if it works with

```
{ include: type === 'entity' ? azure.BlobUtilities.BlobListingDetails.METADATA : '' },
```

dabutvin:
@jeffmcaffer 
I'm starting to get the feeling we should have separate methods for listResults and listEntities instead of list('entity' | 'result')

We would have to split the methods here to `listAllEntities()` and `listAllResults()` https://github.com/clearlydefined/service/blob/master/providers/stores/abstractStore.js#L18

what do you think?

providers/stores/azblob.js
jeffmcaffer: Hmmm, I think we should be consistent. If the result coords are lowered then the caller cannot reasonably/reliably convert back and forth. If we just say that coords "case preserving" then it is simple and easy for people to understand. 

Regarding putting the metadata on all blobs, I agree but it seems like the code has two different behaviors by triggering off the `type` passed to `list`. If the canonical coords were stored in the metadata and all coords were case preserving then the code and user story would be simplified.

dabutvin: If I call the list function with type=result do I want to get back blob names?

If we preserve the casing of the coordinate, it no longer will point to a blob

jeffmcaffer: you want to get back a coordinates object. if it is cased, that's ok as we'll go back to the store at some point and ask for the content related to the coordinates. The store implementation will (or perhaps not) toLowerCase the coordinate elements to form the name/path/id to look for in the store. 

Casing in the store infrastructure is a topic for the store. The inputs and outputs are coordinates. It would be much clearer/stronger if we could say that they are case preserving.

Does that work?

dabutvin: Yep makes sense :+1:

providers/stores/azblob.js
dabutvin: since we're going to preserve case if `type == entity || type == result` I will include metadata for both

providers/stores/file.js
dabutvin: The more I looked into this, I realized I was doing it wrong and ended up dropping JSONStream altogether. Now we will do the following:

1. read the input stream to the end
1. parse the JSON
1. modify the JSON
1. write the JSON to disk
1. resolve the promise

providers/stores/file.js
dabutvin: since we're going to preserve case if `type == entity || type == result` I ended up always reading the file

test/providers/store/fileScan.js
```diff
@@ -19,10 +19,10 @@ describe('list a tool result ', () => {
...
       const result = [
         '/foo/npm/npmjs/-/test/revision/0.0',
-        '/foo/npm/npmjs/-/test/revision/1.0/tool/testTool/2.0.json',
-        '/foo/npm/npmjs/-/test/revision/2.0/tool/testTool0/1.0.json',
-        '/foo/npm/npmjs/-/test/revision/2.0/tool/testTool1/2.0.json',
-        '/foo/npm/npmjs/-/test/revision/2.0/tool/testTool2/3.0.json'
+        '/foo/npm/npmjs/-/test/revision/1.0/tool/testtool/2.0.json',
+        '/foo/npm/npmjs/-/test/revision/2.0/tool/testtool0/1.0.json',
+        '/foo/npm/npmjs/-/test/revision/2.0/tool/testtool1/2.0.json',
+        '/foo/npm/npmjs/-/test/revision/2.0/tool/testtool2/3.0.json'
```
dabutvin: this is the stubbed result of recursive-dir - all the files are lowercased as they are stored now, so updated the stub to return lowercased files

test/providers/store/file.js
```diff
@@ -105,28 +105,15 @@ const data = [
   }
 ]
 
-describe('path to coordinates mapping', () => {
```
dabutvin: we are no longer mapping path to coordinates, we are reading explicit coordinate metadata

providers/stores/azblob.js
```diff
@@ -23,27 +24,70 @@ class AzBlobStore extends AbstractStore {
...
+          resultOrError(resolve, reject)
+        )
+      })
+      result.entries.forEach(entry => {
+        const urn = entry.metadata.urn
+        if (urn) {
+          const value = this._toPreservedCoordinatesFromResultsStoragePath(
+            entry.name,
+            EntityCoordinates.fromUrn(urn).toString()
```
dabutvin: I think we can just do ResultCoordinates.fromUrn(entry.metadata.urn)

providers/stores/file.js
```diff
@@ -34,9 +48,40 @@ class FileStore extends AbstractStore {
...
+        const urn = file.contents._metadata.urn
+        if (urn) {
+          const value = this._toPreservedCoordinatesFromResultsStoragePath(
+            file.path.slice(this.options.location.length + 1),
+            EntityCoordinates.fromUrn(urn).toString()
+          )
```
dabutvin: I think we can just do ResultCoordinates.fromUrn(entry.metadata.urn)

providers/stores/azblob.js
```diff
@@ -23,27 +25,64 @@ class AzBlobStore extends AbstractStore {
     return this.blobService
   }
 
-  async list(coordinates, type = 'entity') {
```
dabutvin: put these methods back together - they are basically the same again

providers/stores/file.js
```diff
@@ -17,14 +19,27 @@ class FileStore extends AbstractStore {
...
-        list.add(value.toString())
+      const files = await Promise.all(
+        paths.map(path => {
+          return new Promise((resolve, reject) =>
+            fs.readFile(path, (error, data) => (error ? reject(error) : resolve({ contents: JSON.parse(data) })))
+          )
+        })
+      )
+      files.forEach(file => {
+        const urn = file.contents._metadata.urn
```
dabutvin: this should be` _metadata.links.self.href`

providers/stores/file.js
```diff
@@ -34,9 +49,34 @@ class FileStore extends AbstractStore {
...
-    const result = super._getEntry(entry, type)
-    return ['gem', 'git', 'npm', 'maven', 'sourcearchive', 'nuget', 'pypi'].includes(result.type) ? result : null
+  /**
+   * List all of the matching tool output coordinates
+   * Accepts partial coordinates.
+   *
+   * @param {EntityCoordinates} coordinates
+   * @returns A list of matching coordinates i.e. [ 'npm/npmjs/-/JSONStream/1.3.3/clearlydefined/1', 'npm/npmjs/-/JSONStream/1.3.3/scancode/2.9.2' ]
+   */
+  async listResults(coordinates) {
```
dabutvin: combine these methods

providers/stores/file.js
dabutvin: follow up on this to combine into a single loop

PR #199, opened 2018-08-15
Take 2 at capturing content

jeffmcaffer:
#193 proposes one approach to capturing interesting file content. structures the content as separate, globally deduplicated blobs. That is kind of nice but makes the system more complex as there are more stores to configure and keep in sync. Crawler PR https://github.com/clearlydefined/crawler/pull/128 implements an alternative that has all the content in one place. 

This PR adjusts for that and presents a different route. Rather than having an explicit `content` route, this PR augments the `definitions` route to have an `expand` query parameter. If you add `?expand=files` then the response definitions include the content of any interesting files. That feels more in line with the user scenarios around notice generation etc.


jeffmcaffer:
OK after some discussions and thought, I'm tending towards the approach in #193 over this. #193 (separate blobs for different "attached" files) is a clearer separation of concerns. I'll leave this one open until the other is merged just in case we need to refer or change our minds.

jeffmcaffer:
closing in favor of #193

PR #202, opened 2018-08-20
add extractLicenseFromLicenseUrl util

dabutvin:
- hook up to the NuGet ClearlyDefined summarizer

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -137,6 +137,6 @@ module.exports = {
   setIfValue,
   setToArray,
   addArrayToSet,
-  extractLicenseFromLicenseUrl
+  extractLicenseFromLicenseUrl,
   mergeDefinitions
 }
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -4,6 +4,7 @@
 const semver = require('semver')
 const EntityCoordinates = require('./entityCoordinates')
 const ResultCoordinates = require('./resultCoordinates')
+const spdxLicenseList = require('spdx-license-list/simple')
 const moment = require('moment')
 const { set } = require('lodash')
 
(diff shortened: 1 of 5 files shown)


lib/utils.js
```diff
@@ -67,17 +67,59 @@ function addArrayToSet(array, set, valueExtractor) {
...
+  },
+  {
+    test: /^\w*http(s)?:\/\/(www.)?apache.org\/licenses\/LICENSE-2\.0(.html)?$/i,
+    license: 'Apache-2.0'
+  },
+  {
+    test: /^\w*https?:\/\/(?:www.)?gnu.org\/licenses\/(.*?)(\.html)?$/i,
+    licenseMatchGroup: 1
+  },
+  {
```
iamwillbar: This one (and the one below) are Microsoft specific so we shouldn't include them in ClearlyDefined.

lib/utils.js
```diff
@@ -67,17 +67,59 @@ function addArrayToSet(array, set, valueExtractor) {
...
 }
 
+const _licenseUrlOverrides = [
+  {
+    test: /^\w*https?:\/\/(?:www.)?opensource.org\/licenses\/mit-license(\.php)?$/i,
+    license: 'MIT'
+  },
+  {
+    test: /^\w*https?:\/\/(?:www.)?opensource.org\/licenses\/(.*?)(\.php)?$/i,
+    licenseMatchGroup: 1
```
iamwillbar: Not sure if it gets handled elsewhere but be careful that the URL isn't case-sensitive but our use of SPDX identifiers *may* be. Same problem exists for the GNU URLs below.

dabutvin: ahhh good point. I'll add a test case for this

dabutvin: @iamwillbar I brought in the [spdx-license-list](https://www.npmjs.com/package/spdx-license-list) package to validate against to get the right casing all the way through

This is the same package used in the [chive](https://github.com/amzn/tiny-attribution-generator/blob/master/package.json#L20) library

PR #204, opened 2018-08-21
add missing copyright deduplication

### LINKED ISSUE #203: Preview API returns duplicated attributions
When a curation get posted to the preview API, some files returns back with duplicated attributions.

Take as example this definition: npm/npmjs/-/react-redux/5.0.7 which has a file with 4 attributions.

If the same definition, without any changes, passes through the preview API, it came back with 7 copyrights on the same file, some of them gets duplicated, as you can see in the attached screenshots.

schermata 2018-08-20 alle 22 11 58
schermata 2018-08-20 alle 22 11 27

jeffmcaffer:
This fixes #203. There was a missing `setIfValue` in the code that computed the definition. We cleaned up all the attributions and used those values to compute the cumulative attributions for each facet but neglected to store the cleaned up attributions on the corresponding file. 

I also took the opportunity to add some simple deduping (though not full cleanup) to the scancode summarizer and add a few tests for these cases.

### DIFF
--- a/business/definitionService.js
+++ b/business/definitionService.js
@@ -259,6 +259,7 @@ class DefinitionService {
     for (let file of facetFiles) {
       file.license ? licenseExpressions.add(file.license) : unknownLicenses++
       const statements = this._simplifyAttributions(file.attributions)
+      setIfValue(file, 'attributions', statements)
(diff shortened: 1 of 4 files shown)


jeffmcaffer:
Proactively merged. Please comment as needed and I'll retroactively address comments.

PR #209, opened 2018-08-25
update fossology summarization

jeffmcaffer:
The fossology summarization was returning an incorrectly structured object `{ nomos: ...}` rather than `{ files: ... }`

There also appeared to be some issues with the regular expressions perhaps due to differences in node versions? On my windows machine with Node 8.9.4 the runtime was complaining that the expressions were invalid.

### DIFF
--- a/providers/summary/fossology.js
+++ b/providers/summary/fossology.js
@@ -18,24 +18,22 @@ class FOSSologySummarizer {
    * @returns {Definition} - a summary of the given raw information
    */
   summarize(coordinates, harvested) {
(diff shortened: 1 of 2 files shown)


jeffmcaffer:
Aggressively merging to get prod in shape.  retroactive review appreciated.

larainema:
Hi @jeffmcaffer , have one comment here, I saw you changed the fossology result in crawler to text/plain, if here not necessary to decode base64.

jeffmcaffer:
Thanks @larainema. That was covered in another PR I believe. I just did a quick check in the service and did not see any issue.

PR #211, opened 2018-08-25
fix merge bug and adapt to nomos output change

jeffmcaffer:
There was a bug in `mergeDefinition` such that new files in the supplied newDefinition were not added to the base. Fixed and added a mess of tests

Also adapted to nomos' changed to NOT base64 encode


jeffmcaffer:
Merging proactively.

PR #222, opened 2018-08-28
fetch fresh curations from GitHub for every def compute

jeffmcaffer:
The current code has the server caching a full copy of the curation repo locally and uses that when computing definitions. This works fine in most scenarios but when the user merges a curation, the service gets a webhook that immediately turns around and tries to (re)compute the definition. Unfortunately, the local copy of the repo will generally not have the "just merged" content. (The repo refresh period is 10 min but even if it was super fast, it will not be that fast)

This PR changes to hit GitHub all the time. That is not as bad as it seems. 

1) most components don't have curations
2) we can hit the GitHub raw content store to test for content without using tokens
3) we only recompute defintions when something might have changed (new blob, new curation) 

We can get better at this but for now we this is relatively simple and enables this important user flow.

### DIFF
--- a/providers/curation/github.js
+++ b/providers/curation/github.js
@@ -7,6 +7,7 @@ const { exec } = require('child_process')
 const fs = require('fs')
 const moment = require('moment')
 const readdirp = require('readdirp')
+const requestPromise = require('request-promise-native')
 const yaml = require('js-yaml')
 const throat = require('throat')
 const Github = require('../../lib/github')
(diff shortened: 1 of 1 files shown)


providers/curation/github.js
```diff
@@ -139,44 +140,28 @@ class GitHubCurationService {
...
+    const branch = await this.getBranchAndSha(pr)
+    const branchName = branch.sha || branch.ref
+    const url = `https://raw.githubusercontent.com/${owner}/${repo}/${branchName}/${path}`
+    const content = await requestPromise({ url, method: 'GET', resolveWithFullResponse: true, simple: false })
+    if (content.statusCode !== 200) return null
+    // If there is content, go and get it. This is a little wasteful (two calls) but
+    // a) the vast majority of coordinates will not have any curation
+    // b) we need the sha of the curation to form part of the final definition tool chain
+    // At some point in the future we can look at caching and etags etc
+    return this._getFullContent(coordinates, branch.ref)
```
dabutvin: should this be `return this._getFullContent(coordinates, branchName)` ?

so we try sha first?

providers/curation/github.js
```diff
@@ -139,44 +140,28 @@ class GitHubCurationService {
...
-      })
-    })
+  async getAll(coordinates, pr = null) {
+    // Check to see if there is content for the given coordinates
+    const path = this._getCurationPath(coordinates)
+    const { owner, repo } = this.options
+    const branch = await this.getBranchAndSha(pr)
+    const branchName = branch.sha || branch.ref
+    const url = `https://raw.githubusercontent.com/${owner}/${repo}/${branchName}/${path}`
+    const content = await requestPromise({ url, method: 'GET', resolveWithFullResponse: true, simple: false })
```
dabutvin: you can use a `HEAD` request instead of `GET` to only do the headers here

providers/curation/github.js
jeffmcaffer: We only use sha. By this time we either have the SHA of master or of the PR.

jeffmcaffer:
I was trying to delete my comment. GitHub was doing odd things or I just did not understand what I was doing.

PR #224, opened 2018-08-28
check route schemas

jeffmcaffer:
Here we want to check the schema of all incoming requests. In particular, the preview api for curations. This adds that check directly to the route handler to sketch out the idea. topics

* There is probably a better way do to this at the route level using joi or some such
* I can't see how to have one schema reference another schema. Have to look at this more. I created curations-ref.json which has the intent of what I want but I cannot see how to make that ref work.
* need some tests to ensure that the curation (and curations) schema is correct


dabutvin:
there are places where we require in the schemas around the app. They need to be updated with the new file names

schemas/validator.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) The Linux Foundation and others. Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+
+const Ajv = require('ajv')
+const ajv = new Ajv({ allErrors: true })
+require('ajv-errors')(ajv)
+ajv.addSchema(require('../schemas/curations-1.0'), 'curations')
+ajv.addSchema(require('../schemas/curation-1.0'), 'curation')
+ajv.addSchema(require('../schemas/definition-1.0'), 'definition')
```
dabutvin: let's drop the 1.0's for now

schemas/curations-1.0.json
```diff
@@ -0,0 +1,79 @@
...
+      "errorMessage": {
+        "required": "Coordinates object require type, provider, namespace, and name"
+      }
+    },
+    "revisions": {
+      "type": "object",
+      "additionalProperties": false,
+      "patternProperties": {
+        "^.+$": {
+          "$ref": "curation-1.0.json"
```
dabutvin: see the drop 1.0's

PR #229, opened 2018-09-08
check harvest requests

### LINKED ISSUE #206: Validate schema of request body in harvest route
at https://github.com/clearlydefined/service/blob/master/routes/harvest.js#L111

We pass the `request.body` straight through to the `harvestService`

We should use schema.json shared with the swagger doc to validate the schema for required parameters, etc

If the schema is invalid we should not call the `harvestService` and instead return a 4XX Bad request status code with information why it's a bad request

jeffmcaffer:
Fix #206

### DIFF
--- a/routes/harvest.js
+++ b/routes/harvest.js
@@ -8,34 +8,37 @@ const utils = require('../lib/utils')
 const EntityCoordinates = require('../lib/entityCoordinates')
 const bodyParser = require('body-parser')
 const { permissionCheck } = require('../middleware/permissions')
+const Ajv = require('ajv')
+const ajv = new Ajv({ allErrors: true })
+require('ajv-errors')(ajv)
+const harvestSchema = require('../schemas/harvest')
 
 // Gets a given harvested file
-router.get(
-  '/:type/:provider/:namespace/:name/:revision/:tool/:toolVersion',
-  asyncMiddleware(async (request, response) => {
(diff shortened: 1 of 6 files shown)


routes/harvest.js
```diff
@@ -54,73 +57,74 @@ function buildFilter(facets) {
...
-  asyncMiddleware(async (request, response) => {
-    await harvestService.harvest(request.body)
-    response.sendStatus(201)
-  })
-)
+router.post('/', permissionCheck('harvest'), bodyParser.json(), asyncMiddleware(queue))
+
+async function queue(request, response) {
+  const requests = Array.isArray(request.body) ? request.body : [request.body]
+  const isValid = ajv.validate(harvestSchema, requests)
```
dabutvin: get the previous pr in first and then resolve conflicts here and use the validator helper utility

routes/harvest.js
```diff
@@ -54,73 +57,74 @@ function buildFilter(facets) {
...
 let harvestStore
 let summarizeService
 
-function setup(harvester, store, summarizer) {
+function setup(harvester, store, summarizer, testFlag = false) {
   harvestService = harvester
   harvestStore = store
   summarizeService = summarizer
+  test = testFlag
+  if (test) router._queue = queue
```
dabutvin: let's see if we can test the route directly with the permissionCheck updates we made friday to be a factory

PR #232, opened 2018-09-12
setup AttachmentCoordinates (fixes #230)

### LINKED ISSUE #230: unable to retrieve attachments through attachments route
We create the coordinates using 
```js
const coordinates = new ResultCoordinates('attachment', null, request.params.token)
```
Which gives us a `type` of `attachment` and a `namespace` of the `token`
image

When we run `abstractStore._toStoragePathFromCoordinates(coordinates)` to actually go get the blob we hit this line 

```js
// if there is a provider then consider the namespace otherwise there can't be one so ignore null
const namespace = c.provider ? c.namespace || '-' : null
```

so this just returns `'attachment'` and drops the token completely because there is no provider.

Now when using the filestore, this still appears to work because of `file._getFilePath`

image

Since we have no `toolVersion` on these coordinates, we look for the latest attachment. So even though i requested an attachment for one token,  I get results for another.

When using `azblob` we are just searching for a blob named `attachment.json` no matter what token gets passed and we 404.

dabutvin:


test/lib/attachmentCoordinates.js
```diff
@@ -0,0 +1,33 @@
+const {expect} = require('chai')
```
jeffmcaffer: need copyright and license header

PR #239, opened 2018-09-26
Get commit hash instead of annotated tag hash

daniellandau:


jeffmcaffer:
Need more context on the user scenario for this change. Proactively getting all the tags for for annotated tags will likely blow our github tokens quickly.

note as well that if we are going to do this in a `map` (or any kind of loop) we'll need to use a `throat` or we will surely get throttled.

daniellandau:
The point is that we have:

```
static printCoordinates = value => (value && value.url ? `${value.url}/commit/${value.revision}` : value)
```

Which is a 404 if `value.revision` is the hash of an annotated tag. I didn't find a way to link to a tag hash on Github and linking to `/tree/${value.tagName}` (that value isn't saved, but could be) seems less real than a commit hash as upstream could just remove that tag name and push a new one with the same name.

One way we could avoid fetching them all proactively would be to not show them in `GitHubCommitPicker` while picking. Currently the format shown is `v2.4.1 (e5b8eb4ae59a954938014c06bb7ec428235e06dc)` where the hash is that of the annotated tag itself. It's not very helpful and not very useful and quite confusing.

daniellandau:
I added a throat.

routes/originGitHub.js
```diff
@@ -20,13 +21,30 @@ router.get(
...
+
+          // we know now that we have an annotated tag, get the associated commit hash
+          let response
+          try {
+            response = await github.gitdata.getTag({ owner: login, repo, sha: item.object.sha })
+          } catch (e) {
+            console.error(e)
+            return null
+          }
+          return { tag: item.ref.slice(10), sha: response.data.object.sha }
```
dabutvin: we should add a comment about this `slice(10)` something like

```
// slice /ref/tags/<tagname> to <tagname>
```

routes/originGitHub.js
```diff
@@ -20,13 +21,30 @@ router.get(
...
+      const unsorted = await Promise.all(answer.data
+        .map(throat(5, async item => {
+          if (item.object.type === 'commit')
+            return { tag: item.ref.slice(10), sha: item.object.sha }
+
+          if (item.object.type !== 'tag') {
+            return null
+          }
+
+          // we know now that we have an annotated tag, get the associated commit hash
```
dabutvin: can you point to an example of an annotated tag on a github repo? I'm having trouble testing this out

daniellandau: Sure. The tag v2.4.1 (https://github.com/chalk/chalk/tree/v2.4.1) points to commit https://github.com/chalk/chalk/commit/48ba5b0b9beadcabd9fc406ac4d9337d8fa6b36d, but the annotated tag hash is e5b8eb4ae59a954938014c06bb7ec428235e06dc (which is the one you currently see in clearlydefined in the harvest page).

routes/originGitHub.js
daniellandau: Added a comment.

routes/originGitHub.js
dabutvin: I see. Makes sense. Thanks for the explanation!

dabutvin:
Verified in dev 👍 

image

PR #241, opened 2018-09-27
set licensed.declared based on interesting files license

dabutvin:
If we find a license on a harvested interesting file, set this to the `licensed.declared` property of the definition.

This runs early, so it can be overwritten by the package manager metadata.

@jeffmcaffer does this meet the bar for declaring a license?

The scenario I am thinking about is someone ships a `LICENSE` file with their package and doesn't register this license metadata anywhere else.

### DIFF
--- a/business/aggregator.js
+++ b/business/aggregator.js
@@ -35,7 +35,7 @@ const { flattenDeep, set } = require('lodash')
 class AggregationService {
   constructor(options) {
     this.options = options
-    // we take the configured precedence exoected to be highest first
+    // we take the configured precedence expected to be highest first
     this.workingPrecedence =
       options.precedence && flattenDeep(options.precedence.map(group => [...group].reverse()).reverse())
   }
--- a/business/aggregator.js
+++ b/business/aggregator.js
@@ -35,6 +35,7 @@ const { flattenDeep, set } = require('lodash')
 class AggregationService {
   constructor(options) {
     this.options = options
+    // we take the configured precedence exoected to be highest first
     this.workingPrecedence =
       options.precedence && flattenDeep(options.precedence.map(group => [...group].reverse()).reverse())
   }
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -116,7 +116,7 @@ function buildSourceUrl(spec) {
       const fullName = `${spec.namespace}/${spec.name}`.replace(/\./g, '/')
       return `https://search.maven.org/remotecontent?filepath=${fullName}/${spec.revision}/${spec.name}-${
         spec.revision
-        }-sources.jar`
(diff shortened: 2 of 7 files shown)


providers/summary/clearlydefined.js
```diff
@@ -100,6 +101,14 @@ class ClearlyDescribedSummarizer {
     setIfValue(result, 'described.releaseDate', extractDate(data.releaseDate))
     setIfValue(result, 'licensed.declared', data.declaredLicense)
   }
+
+  _extractLicenseFromFiles(files) {
+    if (!files) return
+    for (const file of files) {
+      if (file.license && file.license != 'NOASSERTION')
+        return file.license
```
dabutvin: Is this a place where we might form an SPDX expression based on multiple files having different licenses?

lib/utils.js
```diff
@@ -141,11 +141,34 @@ function updateSourceLocation(spec) {
...
+  return false
+}
+
 function _normalizeVersion(version) {
   if (version == '1') return '1.0.0' // version '1' is not semver valid see https://github.com/clearlydefined/crawler/issues/124
   return semver.valid(version) ? version : null
 }
 
+const _licenseFileNames = ['license', 'license.txt', 'license.md', 'license.html']
+const _licenseLocations = { 'npm': ['package/'] }
```
dabutvin: know of any other prefixes we should add here off hand?

business/aggregator.js
```diff
@@ -35,6 +35,7 @@ const { flattenDeep, set } = require('lodash')
 class AggregationService {
   constructor(options) {
     this.options = options
+    // we take the configured precedence exoected to be highest first
```
jeffmcaffer: "expected"

lib/utils.js
jeffmcaffer: 'maven': 'meta-inf/'

Note as well that we should likely expand the set of "license" files. Some communities use "COPYING" for example. @pombredanne  would be a great source of a list of typical filenames containing license declarations for different package types.

providers/summary/clearlydefined.js
```diff
@@ -60,6 +62,14 @@ class ClearlyDescribedSummarizer {
     setIfValue(result, 'files', data.interestingFiles)
   }
 
+  addLicenseFromFiles(result, data, coordinates) {
+    if (!data.interestingFiles) return
+    const licenses = data.interestingFiles
+      .map(file => file.license != 'NOASSERTION' && isLicenseFile(file.path, coordinates) ? file.license : null)
```
jeffmcaffer: !==

dabutvin:
@jeffmcaffer thanks for the review, I updated the branch

PR #242, opened 2018-10-01
massive refactor of providers and add mongo support

jeffmcaffer:
This a pretty big change that was unfortunately hard to tease apart.

* Added support for mongo as a definition store
* That resulted in a review of the store api as we had previously mixed up the harvest and definitions stores (as they were both always using the same provider, file or azblob).
* That resulted in a review of the config structure and a rationalization of the how providers are plugged in
* That resulted in tweaks to the app construction
* Various bits of the above resulted in refactoring/restructuring of related tests
* net result is quite a bit of change

Still to do:
* Attachments store -- stubbed but not implemented. need to decide if that will be a fully separate store or just an api on the definition store (leaning towards the former)
* More testing of the individual providers (especially the azure based ones)


test/providers/store/fileHarvest.js
```diff
@@ -0,0 +1,79 @@
...
+    contents: { _metadata: { links: { self: { href: 'urn:npm:npmjs:-:test:revision:2.0:tool:testtool1:2.0' } } } }
+  },
+  'npm/npmjs/-/test/2.0/tool/testtool2/3.0': {
+    contents: { _metadata: { links: { self: { href: 'urn:npm:npmjs:-:test:revision:2.0:tool:testtool2:3.0' } } } }
+  }
+}
+
+describe('FileHarvestStore list tool results', () => {
+  before(() => {
+    sinon.stub(AbstractFileStore.prototype, 'list').callsFake(async (coordinates, visitor) => {
```
dabutvin: do you actually want to stub `'list` here or stub `fs.readFile`?

routes/harvest.js
```diff
@@ -92,15 +92,15 @@ router.get(
...
-  '/status',
-  bodyParser.json(),
-  asyncMiddleware(async (request, response) => {
-    const coordinatesList = request.body.map(entry => EntityCoordinates.fromString(entry))
-    const result = await harvestStore.listAll(coordinatesList, 'result')
-    response.status(200).send(result)
-  })
-)
+// router.post(
+//   '/status',
```
dabutvin: TODO :)

routes/definitions.js
```diff
@@ -17,7 +17,7 @@ async function getDefinition(request, response) {
   const pr = request.params.pr
   // if running on localhost, allow a force arg for testing without webhooks to invalidate the caches
   const force = request.hostname.includes('localhost') ? request.query.force || false : false
-  const result = await definitionService.get(coordinates, pr, force)
+  const result = await definitionService.get(EntityCoordinates.fromObject(coordinates), pr, force)
```
dabutvin: is `utils.toEntityCoordinatesFromRequest` not returning the `EntityCoordinates` the right way?

providers/stores/mongoose.js
```diff
@@ -0,0 +1,158 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+
+const mongoose = require('mongoose')
+const Schema = mongoose.Schema
+
+class MongooseStore {
+  constructor(options) {
+    throw new Error('dont use this yet. Not sure we need it')
```
dabutvin: TODO :)

providers/stores/abstractAzblobStore.js
```diff
@@ -0,0 +1,73 @@
...
+  constructor(options) {
+    this.options = options
+    this.containerName = options.containerName
+  }
+
+  async initialize() {
+    const blobService = azure
+      .createBlobService(this.options.connectionString)
+      .withFilter(new azure.LinearRetryPolicyFilter())
+    Object.defineProperty(this, 'blobService', { value: blobService, writable: false, configurable: true })
```
dabutvin: not sure if configurable should be set to true here

> true if and only if the type of this property descriptor may be changed and if the property may be deleted from the corresponding object.

providers/stores/abstractAzblobStore.js
```diff
@@ -0,0 +1,73 @@
...
+      if (error.statusCode === 404) return null
+      throw error
+    }
+  }
+
+  _toStoragePathFromCoordinates(coordinates) {
+    return AbstractFileStore.toStoragePathFromCoordinates(coordinates)
+  }
+
+  _toResultCoordinatesFromStoragePath(path) {
```
dabutvin: `_toStoragePathFromCoordinates` and `_toResultCoordinatesFromStoragePath` in this file can be removed

providers/stores/abstractFileStore.js
```diff
@@ -0,0 +1,126 @@
...
+   */
+  async list(coordinates, visitor) {
+    try {
+      const paths = await recursive(this._toStoragePathFromCoordinates(coordinates), ['.DS_Store'])
+      return await Promise.all(
+        paths.map(async path => {
+          if (!this._isValidPath(path)) return null
+          const data = await promisify(fs.readFile)(path)
+          return data ? visitor(JSON.parse(data)) : null
+        })
```
dabutvin: should we add `.filter(x => x)` after this `.map()` since we are returning null in some situations

providers/stores/abstractFileStore.js
```diff
@@ -0,0 +1,126 @@
...
+    const trimmed = path.slice(this.options.location.length + 1)
+    return AbstractFileStore.toResultCoordinatesFromStoragePath(trimmed)
+  }
+
+  async _getFilePath(coordinates) {
+    const toolPath = this._toStoragePathFromCoordinates(coordinates)
+    if (coordinates.toolVersion) return toolPath + '.json'
+    const latest = await this._findLatest(toolPath)
+    if (!latest) return null
+    return path.join(toolPath, latest) + '.json'
```
dabutvin: do you think this belongs in the AbstractFileStore? It's very tied to tool results.

It may make more sense in `fileHarvestStore.js`

providers/stores/abstractFileStore.js
```diff
@@ -0,0 +1,126 @@
...
+        const result = list.map(entry => (path.extname(entry) === '.json' ? path.basename(entry).slice(0, -5) : null))
+        resolve(utils.getLatestVersion(result.filter(e => e)))
+      })
+    })
+  }
+
+  // Static helper methods shared between path-based stores
+
+  static isInterestingCoordinates(coordinates) {
+    return ['gem', 'git', 'npm', 'maven', 'sourcearchive', 'nuget', 'pypi'].includes(coordinates.type)
```
dabutvin: it would be handy if there was a way to read this list from https://github.com/clearlydefined/service/blob/master/schemas/definition.json#L20.

should we include `deadletter`?

providers/stores/azblobDefinitionStore.js
```diff
@@ -0,0 +1,50 @@
...
+class AzBlobDefinitionStore extends AbstractAzBlobStore {
+  /**
+   * List all of the definitions for the given coordinates.
+   *
+   * @param {EntityCoordinates} coordinates - Accepts partial coordinates
+   * @returns A list of matching coordinates i.e. [ 'npm/npmjs/-/JSONStream/1.3.3' ]
+   */
+  async list(coordinates) {
+    const list = await super.list(coordinates, entry => {
+      const urn = entry.metadata.urn
```
dabutvin: I may be misunderstanding how this is setup, but I don't think there is ever blob metadata on the definitions.

I think you can just 

`return AbstractFileStore.isInterestingCoordinates(entry.name)`

providers/stores/mongo.js
```diff
@@ -0,0 +1,70 @@
...
+   * @param {EntityCoordinates} coordinates
+   * @returns A list of matching coordinates i.e. [ 'npm/npmjs/-/JSONStream/1.3.3' ]
+   */
+  async list(coordinates, type = 'entity') {
+    const list = await this.definitionModel.find({ id: this._getId(coordinates) }, 'id')
+    return list.map(entry => entry.id)
+  }
+
+  /**
+   * Get the results of running the tool specified in the coordinates on the entty specified
```
dabutvin: spelling `entty`

providers/stores/mongo.js
```diff
@@ -0,0 +1,70 @@
...
+   * in the coordinates. If a stream is given, write the content directly on the stream and close.
+   * Otherwise, return an object that represents the result.
+   *
+   * @param {ResultCoordinates} coordinates - The coordinates of the result to get
+   * @param {WriteStream} [stream] - The stream onto which the output is written, if specified
```
dabutvin: stream is unused here, we are always just returning the result of findOne

providers/stores/mongo.js
```diff
@@ -0,0 +1,70 @@
...
+   * @returns The result object if no stream is specified, otherwise the return value is unspecified.
+   */
+  get(coordinates, stream) {
+    return this.definitions.findOne({ id: this._getId(coordinates) }, { projection: { _id: 0, id: 0 } })
+  }
+
+  store(coordinates, definition) {
+    definition.id = definition.id || definition.coordinates.toString()
```
dabutvin: should we be using the `_getId` method here?

```
definition.id = definition.id || this. _getId(definition.coordinates)
```

providers/stores/mongo.js
```diff
@@ -0,0 +1,70 @@
...
+    return this.definitions.replaceOne({ id: definition.id }, definition, { upsert: true })
+  }
+
+  delete(coordinates) {
+    return this.definitions.deleteOne({ id: definition.coordinates.toString() })
+  }
+
+  _getId(coordinates) {
+    return EntityCoordinates.fromObject(coordinates).toString()
```
dabutvin: should this id field contain the definition schema version?

I'm thinking about a case where we have 2 definitions for the same coordinates side by side, one with schema=1.0.0 and the other with schema=2.0.0

providers/stores/mongo.js
```diff
@@ -0,0 +1,70 @@
...
+    return this.definitions.findOne({ id: this._getId(coordinates) }, { projection: { _id: 0, id: 0 } })
+  }
+
+  store(coordinates, definition) {
+    definition.id = definition.id || definition.coordinates.toString()
+    return this.definitions.replaceOne({ id: definition.id }, definition, { upsert: true })
+  }
+
+  delete(coordinates) {
+    return this.definitions.deleteOne({ id: definition.coordinates.toString() })
```
dabutvin: should this use `_getId` method too?

routes/attachments.js
```diff
@@ -4,21 +4,19 @@
...
 const router = express.Router()
-const AttachmentCoordinates = require('../lib/attachmentCoordinates')
 
 // Get a proposed patch for a specific revision of a component
 router.get('/:id', asyncMiddleware(getAttachment))
 
 async function getAttachment(request, response) {
-  const coordinates = new AttachmentCoordinates(request.params.id)
-  const result = await harvestStore.getAttachment(coordinates)
+  const result = await attachmentStore.get(request.params.id)
```
dabutvin: either here or in the attachmentStore, we need to add the segment `attachment/` to the blob name.
This was being done in the `AttachmentCoordinates` before.

This `id` is just the hash and then the `azblobAttachmentStore` just does `name = id + '.json'`

providers/stores/azblobConfig.js
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+const config = require('painless-config')
+
+const harvestConnectionString = config.get('HARVEST_AZBLOB_CONNECTION_STRING')
+const harvestContainerName = config.get('HARVEST_AZBLOB_CONTAINER_NAME')
+const definitionConnectionString = config.get('DEFINITION_AZBLOB_CONNECTION_STRING') || harvestConnectionString
+const definitionContainerName = config.get('DEFINITION_AZBLOB_CONTAINER_NAME') || harvestContainerName + '-definition'
+const attachmentConnectionString = config.get('ATTACHMENT_AZBLOB_CONNECTION_STRING') || harvestConnectionString
+const attachmentContainerName = config.get('ATTACHMENT_AZBLOB_CONTAINER_NAME') || harvestContainerName + 'attachments'
```
dabutvin: missing `-` before attachments for default

routes/attachments.js
dabutvin: this is now in a separate container 👍

providers/stores/mongo.js
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * List all of the matching components for the given coordinates.
+   * Accepts partial coordinates.
+   *
+   * @param {EntityCoordinates} coordinates
+   * @returns A list of matching coordinates i.e. [ 'npm/npmjs/-/JSONStream/1.3.3' ]
+   */
+  async list(coordinates, type = 'entity') {
+    const list = await this.definitionModel.find({ id: this._getId(coordinates) }, 'id')
```
dabutvin: we should check if this is doing a 'startsWith' for id

providers/stores/mongo.js
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * List all of the matching components for the given coordinates.
+   * Accepts partial coordinates.
+   *
+   * @param {EntityCoordinates} coordinates
+   * @returns A list of matching coordinates i.e. [ 'npm/npmjs/-/JSONStream/1.3.3' ]
+   */
+  async list(coordinates) {
+    const list = await this.collection.find({ id: this._getId(coordinates) }, 'id')
```
dabutvin: this needs to be update to list by prefix, something like:

```js
await this.collection.find({ id: new RegExp('^' + this._getId(coordinates) }, 'id')
```

app.js
```diff
@@ -20,154 +20,154 @@ const swaggerUi = require('swagger-ui-express')
...
+  initializers.push(async () => definitionStore.initialize())
+
+  const attachmentStore = config.attachment.store()
+  initializers.push(async () => attachmentStore.initialize())
+
+  const searchService = config.search.service()
+  initializers.push(async () => {
+    await searchService.initialize()
+    // Bit of trick for local hosting. Preload search if using an in-memory search service
+    if (searchService.constructor.name === 'MemorySearch') definitionService.reload('definitions')
```
dabutvin: what do you think about moving this into the requestHandler.init right after?

```js
console.log('Service initialized')
```

I was getting a bunch of calls to `store()` in mongo provider before the collection was initialized

providers/stores/mongo.js
```diff
@@ -0,0 +1,66 @@
...
+   * @param {Coordinates} coordinates - The coordinates of the object to get
+   * @returns The loaded object
+   */
+  get(coordinates) {
+    return this.collection.findOne({ id: this._getId(coordinates) }, { projection: { _id: 0, id: 0 } })
+  }
+
+  store(definition) {
+    definition.id = this._getId(definition.coordinates)
+    return this.collection.replaceOne({ id: definition.id }, definition, { upsert: true })
```
dabutvin: this is actually leaking the `id` property back to the definition response.

When we call `get()` we project-away the id property

app.js
jeffmcaffer: Yeah, there is a race between the mongo store being initialized and the this search reloading. Those are both in the same set of initializers. They are run in parallel so it is not guaranteed that the definition store is initialized before this is run. This works in production as the search service is not 'memory' so this is not done. Even still, seems like a good move.

On a related note, when does the mongo collection get created. in mongo.js `initialize` we have code to create the db and the collection but it is unclear if those are local client-only calls or if they talk to the service and need to be awaited or something. Do we need to explicitly create the collection like we do with the container in blob?

providers/stores/mongo.js
dabutvin: Regex does work for mongo filters. 

Good point about needing to escape I hadn't thought of that.

For the redos protection, maybe we can do something simple like limit the length of the input to 100 chars

providers/stores/mongo.js
dabutvin: Yep makes sense. We should just return the original value that is passed in

providers/stores/mongo.js
dabutvin: Not sure. This library looks legit https://github.com/substack/safe-regex/blob/master/readme.markdown

app.js
jeffmcaffer: so are you getting errors from trying to read before the collection is created? That is a different problem. We are not proactively creating the collection so until we upsert it may not be there. I'm hoping we don't have to do a bogus upsert on every start just to keep `find` happy.

Or are you saying that it is failing before `this.collection` the property is set?

dabutvin: `this.collection` is undefined when the reload happens.

We shouldn't need to worry about bogus upsert or anything I think the initializers just need to finish

dabutvin:
👍 👍 👍

PR #246, opened 2018-10-09
return from azblobConfig and fix promisify

dabutvin:


providers/stores/azblobConfig.js
```diff
@@ -10,7 +10,7 @@ const attachmentConnectionString = config.get('ATTACHMENT_AZBLOB_CONNECTION_STRI
 const attachmentContainerName = config.get('ATTACHMENT_AZBLOB_CONTAINER_NAME') || harvestContainerName
 
 function harvest(options) {
-  require('./azblobHarvestStore')(
+  return require('./azblobHarvestStore')(
```
dabutvin: these were returning undefined :(

middleware/permissions.js
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Amazon.com, Inc. or its affiliates and others. Licensed under the MIT license.
 // SPDX-License-Identifier: MIT
 
+const config = require('../bin/config')
```
dabutvin: I don't think this is aligned with the new design for configuration

dabutvin:
merging this to get the main branch back in shape

fyi @jeffmcaffer

PR #247, opened 2018-10-10
remove the defintion tool assignment

dabutvin:
- when we are getting harvest blobs this does not work
 - it ended up the last call filters out any parts that are not provided


dabutvin:
see #248

PR #248, opened 2018-10-10
fix mongo tests and azure definition coordinates

jeffmcaffer:
obsoletes #247 

* Azure blob definitions have moved to be in their own collection but we were still appending on a /tool/definition/1 to the coordinates. Not needed. Remove and fix up tests
* The Mongo definition store tests were left in a broken state. Fix them up.


dabutvin:
👍 nice

this should be the last piece we needed

PR #256, opened 2018-10-23
store curations in doc db

jeffmcaffer:
Given we now have a doc db, it would be useful to store all the curations (proposed and merged) so we can search and find interesting things. Note that this will be a secondary store, the computations around curating definitions etc should remain unchanged and GitHub is the source of truth.

Ready for review and merging


providers/curation/github.js
```diff
@@ -36,6 +35,40 @@ class GitHubCurationService {
...
+
+    // get and store the current state of the curation branch
+    const curations = await this._getContributedCurations(pr.number, pr.head.sha)
+    await this.store.updateCurations(curations)
+
+    // invalidate all affected definitions then recompute. This ensures the changed defs are cleared out
+    // even if there are errors recomputing the definitions.
+    const coordinateList = Curation.getAllCoordinates(curations)
+    await this.definitionService.invalidate(coordinateList)
+    return Promise.all(coordinateList.map(coordinates => this.definitionService.computeAndStore(coordinates)))
```
dabutvin: throttle this, maybe let them fail individually

providers/curation/github.js
```diff
@@ -271,80 +300,38 @@ ${this.formatDefinitions(patch.patches)}`
...
-    const github = Github.getClient(this.options)
     try {
-      const response = await github.pullRequests.getFiles({ owner, repo, number })
+      const response = await this.github.pullRequests.getFiles({ owner, repo, number })
       return response.data
     } catch (error) {
-      // @todo add logger
-      throw error
+      if (error.code === 404) throw new Error(`Could not find pr#${number} in ${owner}/${repo}`)
+      throw new Error(`Error calling GitHub to get pr#${number}. Code ${error.code}`)
```
dabutvin: might want to add the owner/repo in this message too

```suggestion
      throw new Error(`Error calling GitHub to get pr#${number} in ${owner}/${repo}. Code ${error.code}`)
```

providers/curation/mongoCurationStore.js
```diff
@@ -0,0 +1,87 @@
...
+   * Update the contribution entry for the given PR and record the associated curations -- the
+   * actual file content in the PR. If the curations are not provided, just the PR data is updated,
+   * any existing curation data is preserved.
+   * @param {*} pr - the PR object as provided by GitHub
+   * @param {[Curation]} curations? - The set of actual proposed changes.
+   */
+  updateContribution(pr, curations = null) {
+    if (curations) {
+      const files = {}
+      if (curations) curations.forEach(curation => (files[curation.path] = curation.data))
```
dabutvin: checking `if (curations)` twice here

providers/curation/memoryStore.js
```diff
@@ -0,0 +1,49 @@
...
+    curations.forEach(curation => {
+      const coordinates = EntityCoordinates.fromObject(curation.data.coordinates)
+      this.curations[this._getCurationId(coordinates)] = curation.data
+    })
+  }
+
+  updateContribution(pr, curations = null) {
+    if (curations) {
+      const files = {}
+      if (curations) curations.forEach(curation => (files[curation.path] = curation.data))
```
dabutvin: curations checked twice here

providers/curation/mongoCurationStore.js
```diff
@@ -0,0 +1,87 @@
...
+      return this.collection.replaceOne({ _id: pr.number }, { _id: pr.number, pr, files }, { upsert: true })
+    }
+    // TODO reconsider `upsert` here. Great for resiliency but will it result in undetected inconsistent data?
+    return this.collection.updateOne({ _id: pr.number }, { $set: { pr: pr } }, { upsert: true })
```
dabutvin: not sure we need `curations` to be optional here. We pull the set of files from GH based on PR number and commit ref from head of PR. 

The only way there are no curations is if the PR is empty. In this case, we shouldn't need to store anything and this could just always do a full replace above

Is there another case where curations is null?

providers/curation/mongoCurationStore.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  /**
+   * List all of the Curations whose coordinates match the given partial coordinates.
+   * @param {EntityCoordinates} coordinates - partial coordinates to look for
+   * @returns the array of Curations found
+   */
+  // TODO need to do something about paging
```
dabutvin: I can't find anything about the default page size in docs. Is it going to dump all of them?

providers/curation/mongoCurationStore.js
```diff
@@ -0,0 +1,87 @@
...
+   * Store the given set of curations overwriting any previous content. This effectively replicates the
+   * content of the files in the GitHub curation repo.
+   * @param {[Curation]} curations
+   */
+  async updateCurations(curations) {
+    await Promise.all(
+      curations.map(
+        throat(10, async curation => {
+          const _id = this._getCurationId(curation.data.coordinates)
+          await this.collection.replaceOne({ _id }, { _id, ...curation.data }, { upsert: true })
```
dabutvin: do you want to check `if (_id)` before we call `replaceOne`?

providers/curation/mongoCurationStore.js
```diff
@@ -0,0 +1,87 @@
...
+   * List all of the Curations whose coordinates match the given partial coordinates.
+   * @param {EntityCoordinates} coordinates - partial coordinates to look for
+   * @returns the array of Curations found
+   */
+  // TODO need to do something about paging
+  list(coordinates) {
+    if (!coordinates) throw new Error('must specify coordinates to list')
+    const pattern = this._getCurationId(coordinates.asRevisionless())
+    return this.collection
+      .find({ _id: new RegExp('^' + pattern) })
```
dabutvin: do you want to check `if (pattern)` before we call `find`?

providers/curation/github.js
jeffmcaffer: there is only one repo

PR #266, opened 2018-10-28
Limit the number of definitions per request (#192)

### LINKED ISSUE #192: Limit the number of definitions per request
Users can POST to /definitions with a body listing the coordinates for which they want definitions. The service gets/computes the defs and returns them as one body. This can be relatively costly if the defs are not already computed. As means of limiting DOS attacks we should limit the number of coords per request to something like 100. That a) bounds the amount of time the service will spend on any given request and b) forces abusive users to do repeated calls and be subject to our (e.g., CloudFlare) rate limiting.

Implementation is likely a simple check in routes/definitions,js POST /.  There check the count and do a 400 if there are too many (with a good message)

AlexWebYourmind:
Closes #192

### DIFF
--- a/routes/definitions.js
+++ b/routes/definitions.js
@@ -67,12 +67,21 @@ router.post(
   '/',
   asyncMiddleware(async (request, response) => {
     const coordinatesList = request.body.map(entry => EntityCoordinates.fromString(entry))
-    // if running on localhost, allow a force arg for testing without webhooks to invalidate the caches
-    const force = request.hostname.includes('localhost') ? request.query.force || false : false
-    const result = await definitionService.getAll(coordinatesList, force)
-    response.status(200).send(result)
+
+    // Check for possible DDOS attack and handle
+    if(coordinatesList.length > 100) {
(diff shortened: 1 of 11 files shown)


routes/definitions.js
```diff
@@ -67,13 +67,22 @@ router.post(
...
   asyncMiddleware(async (request, response) => {
     const coordinatesList = request.body.map(entry => EntityCoordinates.fromString(entry))
-    // if running on localhost, allow a force arg for testing without webhooks to invalidate the caches
-    const force = request.hostname.includes('localhost') ? request.query.force || false : false
-    const result = await definitionService.getAll(coordinatesList, force)
-    response.status(200).send(result)
+
+    // Check for possible DDOS attack and handle
+    if(coordinatesList.length > 100) {
+      console.log('Possible DDOS attack logged - request body coordinates list greater than 100 entries');
```
dabutvin: I feel like 100 is too small. Can we bump this to `> 1000` ?

PR #268, opened 2018-10-30
harvest new cordinates when defintion is fetched

dabutvin:


business/definitionService.js
```diff
@@ -115,7 +116,10 @@ class DefinitionService {
     // If no tools participated in the creation of the definition then don't bother storing.
     // Note that curation is a tool so no tools really means there the definition is effectively empty.
     const tools = get(definition, 'described.tools')
-    if (!tools || tools.length === 0) return definition
+    if (!tools || tools.length === 0) {
+      await this.harvestService.harvest({ tool: 'package', coordinates })
```
iamwillbar: I don't think we need to `await` here, we're not going to use the results so queuing it up asynchronously would be fine.

iamwillbar: Just need to make sure any exceptions/rejected promises are caught and logged so they don't take the service down.

business/definitionService.js
```diff
@@ -80,7 +81,7 @@ class DefinitionService {
   async list(coordinates, recompute = false) {
     if (!recompute) return this.definitionStore.list(coordinates)
     const curated = await this.curationService.list(coordinates)
-    const tools = await this.harvestService.list(coordinates)
+    const tools = await this.harvestStore.list(coordinates)
```
iamwillbar: Why the switch from `harvestService` to `harvestStore` for these?

business/definitionService.js
iamwillbar: In GHCrawler if the same thing was queued multiple times the subsequent queue requests were dropped, is it the same in ClearlyDefined?

jeffwilcox:
:shipit:

business/definitionService.js
dabutvin: this was using the harvest store all along, it just had a funny name

[HarvestStore](https://github.com/clearlydefined/service/blob/master/providers/stores/azblobHarvestStore.js)
[HarvestService](https://github.com/clearlydefined/service/blob/master/providers/harvest/crawler.js)

business/definitionService.js
dabutvin: I'm guessing that is the case, but I'm not sure. It's using a branch of GHCrawler
https://github.com/Microsoft/ghcrawler/tree/jm/apicrawler/providers/queuing

Do you know where that handling was done?

PR #274, opened 2018-11-07
add quotes around test path

dabutvin:
wasn't finding all the tests on `npm test` without this


dabutvin:
@jeffmcaffer do you want to test this on windows?>

PR #275, opened 2018-11-08
release to prod

dabutvin:


business/definitionService.js
```diff
@@ -80,8 +79,8 @@ class DefinitionService {
    */
   async list(coordinates, recompute = false) {
     if (!recompute) return this.definitionStore.list(coordinates)
-    const curated = await this.curationService.list(coordinates)
-    const tools = await this.harvestStore.list(coordinates)
+    const curated = (await this.curationService.list(coordinates)).map(c => c.toString())
+    const tools = await this.harvestService.list(coordinates)
```
dabutvin: this looks like a merge issue. there is no list() on harvestServiec

```suggestion
    const tools = await this.harvestStore.list(coordinates)
```

PR #277, opened 2018-11-09
add logging

jeffmcaffer:
add support for winston and app insights based logging in a pluggable manner. 

Also add a couple uses of logging just to show how the logger is accessed.

### DIFF
--- a/app.js
+++ b/app.js
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 const express = require('express')
-const logger = require('morgan')
+const morgan = require('morgan')
 const bodyParser = require('body-parser')
 const cookieParser = require('cookie-parser')
 const cors = require('cors')
@@ -15,9 +15,15 @@ const swaggerUi = require('swagger-ui-express')
 const fs = require('fs')
 const yaml = require('js-yaml')
 const swaggerDoc = yaml.safeLoad(fs.readFileSync('./routes/swagger.yaml'))
+const loggerFactory = require('./providers/logging/logger')
 
 function createApp(config) {
   const initializers = []
+
+  const logger = loggerFactory(config.logging.logger())
+  process.on('unhandledRejection', reason =>
+    logger.error('unhandledRejection', { exception: reason, name: 'unhandledRejection' })
+  )
 
   config.auth.service.permissionsSetup()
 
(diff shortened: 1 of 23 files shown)


app.js
```diff
@@ -147,22 +153,27 @@ function createApp(config) {
 
   // error handler
   // eslint-disable-next-line no-unused-vars
-  app.use((err, req, res, next) => {
+  app.use((error, request, response, next) => {
+    if (!(request && request.url && request.url.includes('robots933456.txt')))
```
moranthomas: I would add in a comment in the code directly here to explain this in case the azure feedback url mentioned ever changes e.g. 
// With Web App for Containers, Azure uses a special URL "/robots933456.txt" to check when the container has started, and expects a valid HTTP response (404 is fine).

providers/curation/github.js
```diff
@@ -12,6 +12,7 @@ const Curation = require('../../lib/curation')
 const EntityCoordinates = require('../../lib/entityCoordinates')
 const tmp = require('tmp')
 tmp.setGracefulCleanup()
+const logger = require('../logging/logger')
```
dabutvin: have to call the logger

app.js
```diff
@@ -15,25 +15,31 @@ const swaggerUi = require('swagger-ui-express')
...
 const yaml = require('js-yaml')
 const swaggerDoc = yaml.safeLoad(fs.readFileSync('./routes/swagger.yaml'))
+const loggerFactory = require('./providers/logging/logger')
 
 function createApp(config) {
   const initializers = []
 
+  const logger = loggerFactory(config.logging.logger())
+  process.on('unhandledRejection', reason =>
+    logger.error('unhandledRejection', { exception: reason, name: 'unhandledRejection' })
```
dabutvin: let's validate this

app.js
```diff
@@ -82,9 +88,9 @@ function createApp(config) {
   app.use(requestId())
   app.use(cachingMiddleware(caching()))
 
-  app.use(logger('dev'))
+  app.use(morgan('dev'))
```
moranthomas: Consider using the standard Apache log format - includes useful information in the logs such as remote IP address and the user-agent HTTP request header.
app.use(morgan('combined'));

app.js
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 const express = require('express')
-const logger = require('morgan')
+const morgan = require('morgan')
```
moranthomas: We could also consider installing 
var appRoot = require('app-root-path'); 
very useful when specifying paths to files in Node.js code.

dabutvin:
@jeffmcaffer I pushed some updates after going through the branch, feel free to merge

PR #278, opened 2018-11-09
package -> component for general harvesting

dabutvin:


dabutvin:
dependent on https://github.com/clearlydefined/crawler/pull/199

PR #285, opened 2018-11-20
update swagger spec and UI

### LINKED ISSUE #238: update swagger
The swagger for the service is now getting out of date as we updated the various schemas. Really need to figure out how to have swagger use the JSON schemas we are defining so there is one source of truth,

jeffmcaffer:
fixes #238 

turns out there were multiple issues. 
* old version of swagger-ui 
* swagger.yaml that was invalid (halfway between 2.0 and 3.0 format)
* not supplying a swagger url so swagger could not figure out base urls for referenced docs
* no route serving up the referenced files

This PR addresses all of these and gets the structure in the right format. There are still issues with the actual swagger being out of date and likely some incompatibilities with JSON schema and swagger (e.g., pattern properties). Those can be worked out in a subsequent PR.

### DIFF
--- a/app.js
+++ b/app.js
@@ -12,9 +12,6 @@ const serializeError = require('serialize-error')
 const requestId = require('request-id/express')
 const passport = require('passport')
 const swaggerUi = require('swagger-ui-express')
-const fs = require('fs')
-const yaml = require('js-yaml')
-const swaggerDoc = yaml.safeLoad(fs.readFileSync('./routes/swagger.yaml'))
 const loggerFactory = require('./providers/logging/logger')
 
 function createApp(config) {
@@ -84,9 +81,12 @@ function createApp(config) {
   app.use(helmet())
   app.use(requestId())
   app.use(cachingMiddleware(caching()))
+  app.use('/schemas', express.static('./schemas'))
 
   app.use(morgan('dev'))
-  app.use('/api-docs', swaggerUi.serve, swaggerUi.setup(swaggerDoc))
+
+  const swaggerOptions = { swaggerUrl: 'http://localhost:4000/schemas/swagger.yaml' }
+  app.use('/api-docs', swaggerUi.serve, swaggerUi.setup(null, swaggerOptions))
   app.use('/webhook', bodyParser.raw({ limit: '5mb', type: '*/*' }), webhookRoute)
 
   // OAuth app initialization; skip if not configured (middleware can cope)
--- a/app.js
+++ b/app.js
@@ -85,7 +85,7 @@ function createApp(config) {
 
   app.use(morgan('dev'))
 
-  const swaggerOptions = { swaggerUrl: 'http://localhost:4000/schemas/swagger.yaml' }
+  const swaggerOptions = { swaggerUrl: `${config.endpoints.service}/schemas/swagger.yaml` }
   app.use('/api-docs', swaggerUi.serve, swaggerUi.setup(null, swaggerOptions))
   app.use('/webhook', bodyParser.raw({ limit: '5mb', type: '*/*' }), webhookRoute)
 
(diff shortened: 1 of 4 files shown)


app.js
```diff
@@ -84,9 +81,12 @@ function createApp(config) {
   app.use(helmet())
   app.use(requestId())
   app.use(cachingMiddleware(caching()))
+  app.use('/schemas', express.static('./schemas'))
 
   app.use(morgan('dev'))
-  app.use('/api-docs', swaggerUi.serve, swaggerUi.setup(swaggerDoc))
+
+  const swaggerOptions = { swaggerUrl: 'http://localhost:4000/schemas/swagger.yaml' }
```
AlexWebYourmind: @jeffmcaffer this is probably a minor, but is the hardcoded URL necessary?  The way I generally approach this is via the swagger-express-mw, and injecting a config like:
.```
.
const config = {
  appRoot: __dirname + '/server', // required config
  swaggerFile: __dirname + '/server/api/swagger/swagger.yaml'
};
...

swagger_mw.create(config, function (err, swagger_mw) {
  if (err) {
    console.error('Unable to create swagger-express-mw');
    throw err;
  }
```

But I understand we are using the standard express middlewares and if we don't have any use of the above, no need to import it.

File: schemas/swagger.yaml [hunk 2]
```diff
@@ -54,31 +56,32 @@ paths:
...
-        - $ref: '#/components/parameters/provider'
-        - $ref: '#/components/parameters/namespace'
-        - $ref: '#/components/parameters/name'
-        - $ref: '#/components/parameters/revision'
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: /schemas/curations-1.0.json
```
AlexWebYourmind: @jeffmcaffer @moranthomas I want to spend a few words around the issues we are facing around JSON schemas, starting from Jeff's comments: 

> This PR addresses all of these and gets the structure in the right format. There are still issues with the actual swagger being out of date and likely some incompatibilities with JSON schema and swagger (e.g., pattern properties). Those can be worked out in a subsequent PR

We have actually came up to the same conclusion.  
https://swagger.io/docs/specification/data-models/keywords/  Outlines the supported JSON Schema Keywords, and the Unsupported ones.
If we check for example https://github.com/clearlydefined/service/blob/master/schemas/curations-1.0.json, we'll notice that it makes use of the following **unsopported** keywords:
 - $schema
 - patternProperties

@jeffmcaffer  @moranthomas  I would recommend that We identify all the unsupported  properties and we then initiate a discussion to decide how to address this matter (e.g. creating the corresponding OpeAPI Schema, starting from something like https://github.com/wework/json-schema-to-openapi-schema, etc.)


@moranthomas can you ensure to spend some time doing it and open a new descriptive issue to address this matter?

app.js
jeffmcaffer: right, no this will not be hard coded, that is just where I got to in the experiments. Thanks for pointing it out. Note that the real problem here is that swagger-ui needs to be configured with a `swaggerUrl` if it is going to resolve remote $refs. So what we need here is the service's base URL. That is configured into the service via the `SERVICE_ENDPOINT` config property.

schemas/swagger.yaml
jeffmcaffer: Agreed that this is not a complete solution. It does cover 90% of the cases:

* $schema is not that important afaick
* patternProperties is used in one place, curations

I'm fine heading in the direction you mentioned (see also the related article https://philsturgeon.uk/api/2018/04/13/openapi-and-json-schema-divergence-solved/). However, given we are 90% there, we could also wait for them to complete the OpenAPI 3.1 work. In the meantime we can put an example in the Swagger for curations and call it a day.

Short version: if it is a day of work to get the conversion approach to work, let's do it. If it's more, we likely have bigger fish to fry.

jeffmcaffer:
Merging to get this working for now so we can start making progress on the schemas etc. I opened #289 to capture the possible futures.

PR #286, opened 2018-11-20
Add a suggestions API

jeffmcaffer:
To enable new workflows that suggest curations to users, we need an API that finds/creates suggestions. This is the first of what is sure to be a long line of suggestion algorithms. The point of this PR is to plumb in an API and some basic suggestions to demonstrate the function.

* Only suggesting values for missing things
* Only consults other versions of the same component (i.e., it does not follow other types of relations like source location)
* Pretty rudimentary context info for the origin of each suggestion

### DIFF
--- a/app.js
+++ b/app.js
@@ -61,6 +61,9 @@ function createApp(config) {
   curationService.definitionService = definitionService
   const definitionsRoute = require('./routes/definitions')(definitionService)
 
+  const suggestionService = require('./business/suggestionService')(definitionService, definitionStore)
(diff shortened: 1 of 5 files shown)


jeffmcaffer:
@moranthomas and @storrisi let's collaborate here on the shape of the API etc. I have not yet taken into account the notes in #339

moranthomas:
We've opened up a WIP PR to take forward the development of this work #292 .
So this PR can be closed/archived

PR #288, opened 2018-11-20
check that curations actually exist

### LINKED ISSUE #227: validate components in curations actually exist
When a curation is PATCHed , the service should validate that the included components actually exist.
The service should also do this for curations submitted via the API so we can fail earlier for the user and prevent the bogus PRs from accumulating.

We should consider also doing this on PR webhook events since anyone can post a PR or change the files in a PR. In this case the service should mark the PR with an error "status" and point users to some page that guides them through the problem.

moranthomas:
fixes #227


providers/curation/github.js
```diff
@@ -109,13 +110,43 @@ class GitHubCurationService {
...
       fileBody.committer = { name: info.name || info.login, email: info.email }
+    else logger.info('Github requires name/email to set committer, info.email is currently: ', info.email)
     if (get(currentContent, '_origin.sha')) {
       fileBody.sha = currentContent._origin.sha
+      // Extract the component values
+      const { coordinates, revisions } = currentContent
+      const { name, provider, type } = coordinates
+      // Only get the latest revision
+
+      let newRevision = Object.keys(patch.revisions)[0]
```
jeffmcaffer: I'm torn on this. We can check that all versions exist or just that the component exists. Feels like in the end we should be checking that all things mentioned in the proposed patch exists.

providers/curation/github.js
```diff
@@ -109,13 +110,43 @@ class GitHubCurationService {
     // Github requires name/email to set committer
     if ((info.name || info.login) && info.email)
       fileBody.committer = { name: info.name || info.login, email: info.email }
+    else logger.info('Github requires name/email to set committer, info.email is currently: ', info.email)
```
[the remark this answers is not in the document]
moranthomas: After I deleted the cache on the browser I got logged out and didn't realise it. So it was bombing out and had no idea why so I put the check in. No worries. I can reverse it.

providers/curation/github.js
```diff
@@ -139,11 +140,42 @@ class GitHubCurationService {
...
       return serviceGithub.repos.updateFile(fileBody)
     }
     return serviceGithub.repos.createFile(fileBody)
   }
 
+  // If a definition exists then return the patch.
+  // Otherwise return an array of just the coordinates of the definitions that do not exist
+  async _validateComponentDefinitionExists(patch) {
+    const result = await Promise.all(
+      map(patch.revisions, async (_, key) => {
```
jeffmcaffer: should use `throat` (perhaps 10) as there may be many revisions

providers/curation/github.js
```diff
@@ -139,11 +140,42 @@ class GitHubCurationService {
       fileBody.committer = { name: info.name || info.login, email: info.email }
     if (get(currentContent, '_origin.sha')) {
       fileBody.sha = currentContent._origin.sha
+
+      try {
+        let validateResult = await this._validateComponentDefinitionExists(patch)
+        for (var elem of validateResult) {
+          if (elem.revision) {
+            return response.status(400).send({ errors: `component does not exist: ${elem.revision.toString()}` })
```
jeffmcaffer: where is `response` coming from? This is the service that could be called from anywhere. response and status codes should stay in the route.

jeffmcaffer: Related, perhaps validation should go in `addOrUpdate`. There we can just use `definitionService.getAll()` having created a complete list of all the definitions affected in the curation. If some of them are missing then we can report back the missing list and bail on the entire operation.

providers/curation/github.js
```diff
@@ -139,9 +140,41 @@ class GitHubCurationService {
...
+  // This function should only return an array of the coordinates of the definitions that do not exist
+  async _validateComponentDefinitionExists(patch) {
+    const result = await map(patch.revisions, async (_, key) => {
+      const coordinates = Object.assign({}, patch.coordinates, {
+        revision: key,
+        tool: 'definition',
+        toolVersion: 1
+      })
+      try {
+        const singleResult = await this.definitionService.definitionStore.get(coordinates)
```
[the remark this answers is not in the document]
jeffmcaffer: yes. call definitionSERVICE.get() so you are not exposed to the implementation details of the store. 

In fact, `definitionService._getDefinitionCoordinates` is dead method. The stores themselves decide how the definitions are keyed/stored.

providers/curation/github.js
moranthomas: from analysis it appears that if we use definitionService.getAll() , it will fail the entire PATCH if only one of the definition/revision is missing. Because of this behaviour we have implemented a solution that uses definitionService.get(coordinates) instead

providers/curation/github.js
moranthomas: agreed.

providers/curation/github.js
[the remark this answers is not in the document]
moranthomas: see comment below on reasons for moving away from getAll() and moving to the get () method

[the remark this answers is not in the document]
jeffmcaffer: As discussed in the call, I was wrong suggesting `getAll`. I should have suggested `list`. The point being that enabling the service to batch as much as possible is a good thing. `list` ends up being one call to the backing store. `get` would do N independent calls.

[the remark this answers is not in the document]
moranthomas: Sure thing. We have implemented using getList now.

dabutvin:
@moranthomas I got the tests fixed up in master, you can uncomment them out here now :)

moranthomas:
@dabutvin Thanks! Test case code uncommented out here now and all checks have passed

AlexWebYourmind:
@jeffmcaffer @dabutvin is this PR ready to go (merge to master)?

providers/curation/github.js
```diff
@@ -144,15 +145,60 @@ class GitHubCurationService {
...
+    const validResult = { errors: notFoundDefinitions }
+    if (validPatches.length === 0) {
+      validResult.errors.push(
+        'The contribution has failed because none of the supplied component definitions were found'
+      )
+      return validResult
+    }
     await Promise.all(
-      patch.patches.map(throat(1, component => this._writePatch(serviceGithub, info, component, prBranch)))
+      validPatches.map(throat(10, validPatch => this._writePatch(serviceGithub, info, validPatch, prBranch)))
```
jeffmcaffer: Believe we actually have to use 1 as GitHub balks at multiple concurrent patches being written on the same branch or some such.

providers/curation/github.js
```diff
@@ -144,15 +145,60 @@ class GitHubCurationService {
...
+
   async addOrUpdate(userGithub, serviceGithub, info, patch) {
     const { owner, repo, branch } = this.options
     const masterBranch = await serviceGithub.repos.getBranch({ owner, repo, branch: `refs/heads/${branch}` })
     const sha = masterBranch.data.commit.sha
     const prBranch = await this._getBranchName(info)
     await serviceGithub.gitdata.createReference({ owner, repo, ref: `refs/heads/${prBranch}`, sha })
+    const { validPatches, notFoundDefinitions } = await this._validateComponentDefinitionExists(patch.patches)
+    const validResult = { errors: notFoundDefinitions }
+    if (validPatches.length === 0) {
```
jeffmcaffer: Good question here is whether we should succeed at all if there are failures. My first cut at this is to say fail it all if any one of them are bad. That is part of the value of putting the check higher up in the code. We can do that check for all before writing anything

Here I would throw. In the route we can catch and return a 400

routes/curations.js
```diff
@@ -63,9 +63,11 @@ async function updateCurations(request, response) {
   })
   if (curationErrors.length > 0) return response.status(400).send({ errors: curationErrors })
   const result = await curationService.addOrUpdate(userGithub, serviceGithub, info, request.body)
+  if (!result.contribution) return response.status(400).send({ errors: result.errors })
```
jeffmcaffer: suggest throwing in addOrUpdate catching/handling here

providers/curation/github.js
AlexWebYourmind: @jeffmcaffer We set that to 10 as you originally suggested to use that setup. @moranthomas can you restore it to 1?

moranthomas: That's an easy rollback. No problem

business/definitionService.js
```diff
@@ -84,6 +84,39 @@ class DefinitionService {
...
+  /**
+   * Get a list of all the definitions that exist in the store matching the given coordinates
+   * @param {Object[]} coordinatesList
+   * @returns {Object[]} A list of all components that have definitions that are available
+   */
+  async listAll(coordinatesList) {
+    //Take the array of coordinates, strip out the revision and only return uniques
+    const coordinatesWithoutRevision = uniqBy(
+      coordinatesList.map(coordinates => {
+        const withoutRevision = omit(coordinates, ['revision'])
```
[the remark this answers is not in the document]
moranthomas: Agreed, that would be a simpler and more consistent approach

business/definitionService.js
```diff
@@ -84,6 +84,39 @@ class DefinitionService {
...
+        return withoutRevision
+      }),
+      isEqual
+    )
+
+    const promises = coordinatesWithoutRevision.map(async coordinates => {
```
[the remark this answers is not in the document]
moranthomas: Affirmative

providers/curation/github.js
storrisi: I've understand your point, but from a contributor point of view i think it couldn't be worth.
Thinking about a contribution made of hundreds of components, if only one them is not existing then the whole PR should fail?
What we've done here is to allow all the valid definitions to be contributed, and return a message for those that are not existing.
We could definitely go back and stop the entire process if something is wrong, we just want to be sure that is the expected behavior during a contribution process.

business/definitionService.js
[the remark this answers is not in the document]
jeffmcaffer: My comment was specifically to use `asRevisionless()`. That will return the right revisionless coordinates (instance of correct class, has only the right properties etc). `omit` does not return an instance of EntityCoordinates and will leave on any additional properties.

providers/curation/github.js
jeffmcaffer: as discussed in the call today, we decided to fail the entire call if there are any problems. The API caller should get a 400 and hopefully some indication as to what was wrong.

storrisi:
@jeffmcaffer we have updated the PR based on your suggestions, and modified the workflow returning a 400 error if any of the definitions doesn't exits in the store.
The implementation has been provided with related tests as well.

providers/curation/github.js
```diff
@@ -169,14 +171,59 @@ class GitHubCurationService {
...
+  }
+
   async addOrUpdate(userGithub, serviceGithub, info, patch) {
     const { owner, repo, branch } = this.options
+    const { isValid, notFoundDefinitions } = await this._validateComponentDefinitionExists(patch.patches)
+    if (!isValid) {
+      notFoundDefinitions.push(
+        'The contribution has failed because some of the supplied component definitions do not exist'
+      )
+      return { errors: notFoundDefinitions }
```
jeffmcaffer: consider throwing here. from an API point of view it seems reasonable. That is, the caller passed in some non-existent coords. Like trying to open a non-existent file. 

Also, it would avoid having to return funky structured objects.

Actually I see that i suggested that at the call site a while back

storrisi: We've changed this and throwing an error as you suggested.
To test it, given this is an async method, we introduced `chai-as-promised`

providers/curation/github.js
```diff
@@ -211,7 +258,7 @@ ${this._formatDefinitions(patch.patches)}`
       body: `You can review the change introduced to the full definition at [ClearlyDefined](https://clearlydefined.io/curations/${number}).`
     }
     await serviceGithub.issues.createComment(comment)
-    return result
+    return { contribution: result }
```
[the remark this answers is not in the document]
storrisi: That what used when returning an object with both `errors` and `contribution`. Now that we return only contributed data, we've removed it.

business/definitionService.js
```diff
@@ -88,6 +100,35 @@ class DefinitionService {
     return sortedUniq([...harvest, ...curated])
   }
 
+  /**
+   * Get a list of all the definitions that exist in the store matching the given coordinates
+   * @param {Object[]} coordinatesList
```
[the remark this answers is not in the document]
storrisi: You are right, we implemented it

routes/curations.js
storrisi: We have added a try / catch method as you suggested

providers/curation/github.js
```diff
@@ -144,15 +145,60 @@ class GitHubCurationService {
     return serviceGithub.repos.createFile(fileBody)
   }
 
+  // Return an array of valid patches that exist
+  // and a list of definitions that do not exist in the store
+  async _validateComponentDefinitionExists(patches) {
+    const coordinatesWithRevisions = flatten(
```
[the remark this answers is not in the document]
storrisi: Yes, thanks for the suggestion. I've updated a little bit in order to let it work into the implementation.

business/definitionService.js
```diff
@@ -84,6 +84,39 @@ class DefinitionService {
...
+
+    const promises = coordinatesWithoutRevision.map(async coordinates => {
+      try {
+        return await this.list(coordinates)
+      } catch (error) {
+        console.log(error)
+      }
+    })
+
+    const foundDefinitions = flatten(await Promise.all(promises))
```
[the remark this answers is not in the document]
storrisi: We used `intersectionWith` , much cleaner, thanks

business/definitionService.js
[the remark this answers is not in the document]
storrisi: Thanks for the suggestion, we've applied it to the code.

storrisi:
@jeffmcaffer we've updated the PR addressing your suggestions.
I don't know why sometimes the checks of the CI are pending?

providers/curation/github.js
jeffmcaffer: When you roll this back, perhaps add a code comment to the effect that it MUST be 1 or GitHub gets upset.

business/definitionService.js
```diff
@@ -88,6 +101,32 @@ class DefinitionService {
...
+  /**
+   * Get a list of all the definitions that exist in the store matching the given coordinates
+   * @param {EntityCoordinates[]} coordinatesList
+   * @returns {Object[]} A list of all components that have definitions that are available
+   */
+  async listAll(coordinatesList) {
+    //Take the array of coordinates, strip out the revision and only return uniques
+    const searchCoordinates = uniqBy(coordinatesList.map(coordinates => coordinates.asRevisionless()), isEqual)
+    const promises = searchCoordinates.map(
+      throat(1, async coordinates => {
```
jeffmcaffer: this should be `throat(10, ...)` or something bigger than 1

providers/curation/github.js
storrisi: I've added a comment on this instruction

storrisi:
I've updated the PR with the latest suggestions, and also added a comment over the throat method used for writing patches.

Issue #290, opened 2018-11-21
Create a new suggestions API


storrisi:
This one address the service implementation for https://github.com/clearlydefined/website/issues/339

* the client asks for a `GET /suggestions` API, passing in the path of the definition, like `["npm/npmjs/-/async/0.2.6"]`
* then, the client should answer with an object containing all the suggestions for any missing field of the specific version of the definition, taken from any related data
* suggestions should come back from the service with some match ordering
* Each suggested value should have be annotated with whether or not it is from a curation (and possibly who) as well as the date of the curation and the PR number of the curation. This enables users to make informed choices.
* a sample of the service response is:
```
{
  "npm/npmjs/-/async/0.2.6": {
    "described": {
      "sourceLocation": [
        {
          "version": "0.0.1",
          "value": {
            "type": "git",
            "provider": "github",
            "namespace": "caolan",
            "name": "async",
            "revision": "e265181a366552260d34d64172149f659ac797be",
            "url": "https://github.com/caolan/async/commit/e265181a366552260d34d64172149f659ac797be"
          },
         "date": '2018-11-19', 
          "curation": '365', 
          "curator": '@storrisi'
        },
        {
          "version": "0.0.3",
          "value": {
            "type": "git",
            "provider": "github",
            "namespace": "caolan",
            "name": "async",
            "revision": "e265181a366552260d34d64172149f659ac797be",
            "url": "https://github.com/caolan/async/commit/e265181a366552260d34d64172149f659ac797be"
          },
         "date": '2018-11-19', 
          "curation": '365', 
          "curator": '@storrisi'
        }
      ],
      "releaseDate": [
        { "version": "0.0.1", "value": "2013-03-03" ,"date": '2018-11-19', "curation": '365', "curator": '@storrisi'},
        { "version": "0.0.2", "value": "2015-03-03" ,"date": '2018-11-19', "curation": '365', "curator": '@storrisi' }
      ]
    },
    "files": [
      {
        "path": "package/LICENSE",
        "license": [
          { "version": "0.0.1", "value": "MIT" ,"date": '2018-11-19', "curation": '365', "curator": '@storrisi'  },
          { "version": "0.0.2", "value": "Apache-2.0" ,"date": '2018-11-19', "curation": '365', "curator": '@storrisi'  }
        ],
        "attributions": [
[...]

PR #292, opened 2018-11-23
Suggestions API

moranthomas:
This is a WIP PR for continuing the development of the suggestions API based off of the work started on #286


moranthomas:
1. Now retrieving the list of related definitions from the store 
2. Where related means for the same component with different revisions
3. Sorting them ByReleaseDate 
4. Collecting suggestions for a given related definition and field
5. We only give suggestions if there is something missing 
6. Added some basic tests 
7. Currently only created suggestions for a subset of fields: 
'licensed.declared'
'described.sourceLocation'
'described.releaseDate'

business/definitionService.js
```diff
@@ -88,6 +88,35 @@ class DefinitionService {
     return sortedUniq([...harvest, ...curated])
   }
 
+  /**
+   * Get a list of all the definitions that exist in the store matching the given coordinates
+   * @param {Object[]} coordinatesList
```
jeffmcaffer: Can this be an array of EntityCoordinates? seems like a better API and the code in here does not have to do the  `fromObject` calls

business/definitionService.js
```diff
@@ -88,6 +88,35 @@ class DefinitionService {
...
+   * @returns {Object[]} A list of all components that have definitions that are available
+   */
+  async listAll(coordinatesList) {
+    //Take the array of coordinates, strip out the revision and only return uniques
+    const coordinatesWithoutRevision = uniqBy(
+      coordinatesList.map(coordinates => EntityCoordinates.fromObject(coordinates).asRevisionless()),
+      isEqual
+    )
+
+    const promises = coordinatesWithoutRevision.map(async coordinates => {
```
jeffmcaffer: need a throat

business/definitionService.js
```diff
@@ -88,6 +88,35 @@ class DefinitionService {
...
+      try {
+        return await this.list(coordinates)
+      } catch (error) {
+        return null
+      }
+    })
+    const foundDefinitions = await Promise.all(promises)
+    // Filter only the revisions matching the found definitions
+    return coordinatesList.filter(coordinates => {
+      return (
```
jeffmcaffer: should be able to use lodash `intersection` or some such.

business/suggestionService.js
```diff
@@ -0,0 +1,89 @@
...
+    if (!related) return null
+    const baseSuggestion = this._createBaseSuggestion(coordinates)
+    // create a struct with the strings in it and iterate instead
+    const promises = [
+      await this._collectSuggestionsForField(related, baseSuggestion, 'licensed.declared'),
+      await this._collectSuggestionsForField(related, baseSuggestion, 'described.sourceLocation'),
+      await this._collectSuggestionsForField(related, baseSuggestion, 'described.releaseDate')
+    ]
+    const result = await Promise.all(promises)
+    return last(result.filter(res => !isEmpty(res)))
```
jeffmcaffer: Full word names please

providers/stores/mongo.js
```diff
@@ -32,13 +32,15 @@ class MongoStore {
    * @param {EntityCoordinates} coordinates
    * @returns A list of matching coordinates i.e. [ 'npm/npmjs/-/JSONStream/1.3.3' ]
    */
-  async list(coordinates) {
+  async list(coordinates, expand = null) {
```
jeffmcaffer: need to modify the other definition providers

business/suggestionService.js
```diff
@@ -0,0 +1,89 @@
...
+   * Get related definitions for the given coordinates.
+   * Related here means other definitions for the same component with different revisions
+   *
+   * @param {EntityCoordinates} coordinates - The entity we are looking for related defintions to
+   */
+  async _getRelatedDefinitions(coordinates) {
+    const related = await this.definitionStore.list(coordinates.asRevisionless(), 'definitions')
+    // If the related array only has one entry then return early
+    if (Object.keys(related).length <= 1) {
+      return
```
jeffmcaffer: no need for `{}`

business/suggestionService.js
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * Get related definitions for the given coordinates.
+   * Related here means other definitions for the same component with different revisions
+   *
+   * @param {EntityCoordinates} coordinates - The entity we are looking for related defintions to
+   */
+  async _getRelatedDefinitions(coordinates) {
+    const related = await this.definitionStore.list(coordinates.asRevisionless(), 'definitions')
+    // If the related array only has one entry then return early
```
jeffmcaffer: why return early? the comment does not add anything to over the code.

business/definitionService.js
```diff
@@ -88,6 +88,35 @@ class DefinitionService {
     return sortedUniq([...harvest, ...curated])
   }
 
+  /**
+   * Get a list of all the definitions that exist in the store matching the given coordinates
+   * @param {Object[]} coordinatesList
+   * @returns {Object[]} A list of all components that have definitions that are available
+   */
+  async listAll(coordinatesList) {
```
jeffmcaffer: this function does not appear to be referenced

providers/stores/mongo.js
jeffmcaffer: feels odd to have a `list` that gets the contents. Sort of like having Bash `ls` to the equivalent of `cat`. feels more like a "query" operation. Perhaps `list` just gets renamed to `query"?

business/suggestionService.js
```diff
@@ -0,0 +1,89 @@
...
+    if (Object.keys(related).length <= 1) {
+      return
+    }
+    const coordinatesList = []
```
jeffmcaffer:
```suggestion
    const coordinatesList = related.map(element => element.coordinates)
```

business/suggestionService.js
```diff
@@ -0,0 +1,89 @@
...
+    related.forEach(element => {
+      coordinatesList.push(element.coordinates)
+    })
+    const validDefinitions = await this.definitionService.getAll(coordinatesList)
```
jeffmcaffer: above on L40 we called `list` with `'definitions'` which IIRC should get the actual definition objects. do we still need to collect up the coordinates and get them all?

business/suggestionService.js
```diff
@@ -0,0 +1,89 @@
...
+    // Split the definitions into before supplied coords and those after
+    const index = sortedByReleaseDate.findIndex(entry => entry.coordinates.revision === coordinates.revision)
+    if (index === -1) return null
+    const before = sortedByReleaseDate.slice(Math.max(index - 3, 0), index).reverse()
+    const after = sortedByReleaseDate.slice(index + 1, index + 3)
+    return { sortedByReleaseDate, index, before, after }
+  }
+
+  _createBaseSuggestion(coordinates) {
+    return {
```
jeffmcaffer: nit: one line

business/suggestionService.js
jeffmcaffer: Why `last` here? Will this ultimately be `baseSuggestion`?

business/suggestionService.js
storrisi: The related array at this stage contains all the found definitions in the store, including the current one that we are passing.
So if the related array contains only one element, it is the same one that we have passed, and so we can exit early here.

But we have some questions on this approach, because your first cut uses two arrays (before and after) to know what are the 3 revisions after the passed one, and the 3 past before.
We could maybe improve it using some `lodash` methods like `take`, but we are asking if this is really what we expect in terms of returned data or if you have any other thoughts?

business/suggestionService.js
storrisi: Given another comment, i'm gonna remove this method and return `baseSuggestion`

business/definitionService.js
storrisi: This method is the same used in #288 , we will update it from there

providers/stores/mongo.js
storrisi: Seems like this change is coming from your original commit.
Besides that, i agree with your comment. Do you want us to address it in the next commits?

business/suggestionService.js
storrisi: You are right, i've double checked this and seems a duplicate method.

business/definitionService.js
jeffmcaffer: Right. that one looks more up to date but still deals in object. I'm guessing that this branch is based on that branch or something?

storrisi: Not completely, we used only this method, and actually in a previous implementation.
So we're gonna updated it directly in the code.

providers/stores/mongo.js
jeffmcaffer: right. the original did just list the coordinates. With the addition of expand and returning content, the name should be updated. Need to see if the method is being used anywhere else, for anything else. Right now I'm personally leaning towards `find`

business/suggestionService.js
storrisi: After remove it, i've added back this method.
We use getAll() to retrieve whole definition data, because the list() methods returns only coordinates.

storrisi:
I've updated the PR solving some comments and adding the suggestions for each file of the definition.
Note that i've to changed something on `scancode.js` because otherwise it will fail when searching a component without license.

business/suggestionService.js
jeffmcaffer: I think that the shape of the return suggestion data will change considerably over the next while. In general the data coming back from the service should be well structured as suggestions. In the fullness of time we'll be getting suggestions from all over so will get more "sophisticated" than a list of all versions of the component.

I'm fine with taking some simple approach now. If you want to do that, please note the reason for returning early in the comment.

Does that cover your questoins?

storrisi: I will extend the comment in the code.
But from your answer, which is pretty reasonable, i'm still not able to understand why to get the information only from the 3 more recent and the 3 past before the current version, instead of taking, as an example, the latest 6 versions, without considering what's before or after the current revision.

jeffmcaffer: Oh, I just picked numbers. 3, 5, 6, ... as for the before and after, seemed reasonable that the user would consider examples that "surround" the definition in question. Not wed to this idea. Similarly, we should not get wed to any other approach at this point. They are very likely to be wrong. I just picked one that "seemed interesting". We need to pick something and try it.

storrisi: I see, we can go with this and see what happens.

storrisi:
I've update this PR adding some checks for null values for files of a definition.
I'd like to extend the returned data with the date of the curation (if any) and the curator, but i'm unable to get those data. 
@jeffmcaffer do you have any suggestion to retrieve those data?

Beside this, i think we are ready to go, this could be considered not a WIP anymore.

storrisi:
As discussed during the weekly meeting, this PR is almost completed.
Only the curation data (date and curator) are missing (see #290 for schema) and should be retrieved from the mongo store.
Do you guys have any advice to get those data, or do you maybe prefer to integrate your implementation on this?

business/suggestionService.js
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  /**
+   * Get related definitions for the given coordinates.
+   * Related here means other definitions for the same component with different revisions
+   *
+   * @param {EntityCoordinates} coordinates - The entity we are looking for related defintions to
+   */
+  async _getRelatedDefinitions(coordinates) {
+    const related = await this.definitionService.list(coordinates.asRevisionless(), true)
```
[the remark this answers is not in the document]
storrisi: Thanks for the suggestion. What should be a more appropriate method to call instead of `list`?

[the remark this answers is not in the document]
dabutvin: We just integrated the new find method on the definitionService/store. I'm going to get this branch up to date now

business/suggestionService.js
```diff
@@ -0,0 +1,134 @@
...
+   */
+  async get(coordinates) {
+    const related = await this._getRelatedDefinitions(coordinates)
+    if (!related) return null
+    const baseSuggestion = this._createBaseSuggestion(coordinates)
+    // create a struct with the strings in it and iterate instead
+    const promises = [
+      await this._collectSuggestionsForField(related, baseSuggestion, 'licensed.declared'),
+      await this._collectSuggestionsForField(related, baseSuggestion, 'described.sourceLocation'),
+      await this._collectSuggestionsForField(related, baseSuggestion, 'described.releaseDate'),
```
dabutvin: releaseDate doesn't really make sense in this context. We are getting other versions of the same component, they would rarely have the same release dates

business/suggestionService.js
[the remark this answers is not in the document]
storrisi: Thanks @dabutvin , i guess that that method should be used instead of `list` ?

business/suggestionService.js
storrisi: Is the release date related to the component itself or to the version?
When implementing this i supposed that the release date would change on any new version of the same component.

jeffmcaffer: releaseDate likely does not make sense

business/suggestionService.js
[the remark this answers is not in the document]
dabutvin: yep. I'll push an update here shortly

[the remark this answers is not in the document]
storrisi: Thanks, do you need my help here or somewhere else in this PR?

business/suggestionService.js
dabutvin: I was also planning to drop the file suggestions for this first iteration as we decided to leave them off of the find() results.

So that narrows this down to suggesting declared licenses and sourceLocation.

business/suggestionService.js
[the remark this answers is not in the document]
dabutvin: @storrisi  i just pushed. Would you be able to take a look at the changes I made and give me a review before we integrate this?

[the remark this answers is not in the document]
storrisi: Sure, i'm gonna do it

dabutvin:
thanks @storrisi ! I'll merge it now.

Can you get a PR open to enable this from the website side?

PR #293, opened 2018-11-27
add missing binds to promisified storage calls

dabutvin:


dabutvin:
I got webhook test failures on this pr.

image

storrisi:
@dabutvin the same thing happened in #288 

It's related to something introduced in https://github.com/clearlydefined/service/commit/6a9ce68b36d14d0546929517005de8d0b4daa54a

dabutvin:
yep good catch. Looking into it

PR #297, opened 2018-11-28
add curations/sync route

dabutvin:
new route `/curations/sync` to synchronize all the curation PRs and recompute definitions for merged ones.

This route is locked down to the curation github team `curation-dev|curation-prod` 
Authenticate with bearer auth github token

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "url": "git+https://github.com/clearlydefined/service.git"
   },
   "dependencies": {
-    "@octokit/rest": "^14.0.0",
+    "@octokit/rest": "^14.0.9",
     "ajv": "^6.5.2",
     "ajv-errors": "^1.0.0",
     "applicationinsights": "^1.0.6",
--- a/providers/curation/github.js
+++ b/providers/curation/github.js
@@ -12,7 +12,6 @@ const Curation = require('../../lib/curation')
 const EntityCoordinates = require('../../lib/entityCoordinates')
 const tmp = require('tmp')
 tmp.setGracefulCleanup()
-const logger = require('../logging/logger')()
 
 // Responsible for managing curation patches in a store
 //
@@ -27,6 +26,7 @@ class GitHubCurationService {
     this.curationUpdateTime = null
     this.tempLocation = null
     this.github = Github.getClient(options)
+    this.logger = require('../logging/logger')()
   }
 
   get tmpOptions() {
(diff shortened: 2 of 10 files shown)


providers/curation/github.js
```diff
@@ -37,42 +37,63 @@ class GitHubCurationService {
...
+      repo: this.options.repo,
+      per_page: 100,
+      state: 'all'
+    })
+    let prs = response.data
+    while (this.github.hasNextPage(response)) {
+      response = await this.github.getNextPage(response)
+      prs = prs.concat(response.data)
+    }
+    for (let pr of prs) {
```
dabutvin: do this in one list - no need to get all of them in memory

routes/curations.js
```diff
@@ -69,6 +70,13 @@ async function updateCurations(request, response) {
   })
 }
 
+router.post('/sync', permissionsCheck('curate'), asyncMiddleware(syncAllContributions))
```
dabutvin: make sure this setting is in all the environments. 

`AUTH_CURATION_TEAM=curation-dev|curation-prod`

providers/curation/github.js
```diff
@@ -37,42 +38,66 @@ class GitHubCurationService {
...
-  prClosed(pr) {
-    return this.store.updateContribution(pr)
+  async updateContribution(pr) {
+    const curations = await this._getContributedCurations(pr.number, pr.head.sha)
+    const data = {
+      number: pr.number,
+      id: pr.id,
+      state: pr.state,
+      title: pr.title,
+      user: { login: get(pr, 'user.login') },
```
jeffmcaffer: is this going to be the logged in user the clearlydefinedbot user?

providers/curation/github.js
```diff
@@ -37,42 +38,66 @@ class GitHubCurationService {
...
+   * @param {*} pr - The GitHub PR object
    * @returns Promise indicating the operation is complete. The value of the resolved promise is undefined.
-   * @throws Exception with `code` === 404 if the given PR is missing. Other exceptions may be thrown related
-   * to interaction with GitHub or PR storage
    */
-  prClosed(pr) {
-    return this.store.updateContribution(pr)
+  async updateContribution(pr) {
+    const curations = await this._getContributedCurations(pr.number, pr.head.sha)
+    const data = {
```
jeffmcaffer: feels like "not enough of the data" but that is just a gut feel.

routes/webhook.js
```diff
@@ -28,16 +28,14 @@ async function handleGitHubCall(request, response) {
   const pr = body.pull_request
   try {
     switch (body.action) {
-      case 'opened': {
-        await curationService.prOpened(pr)
+      case 'opened':
+      case 'synchronize': {
+        curationService.validateContributions(pr.number, pr.head.sha)
```
jeffmcaffer: little hard to tell but it looks like this will `_getPrFiles` twice as both validate and update call `_getContributedCurations` which goes to GitHub and gets the files. truth?

providers/curation/github.js
jeffmcaffer: might be able to use lodash `pick` https://lodash.com/docs/4.17.11#pick

package.json
```diff
@@ -13,7 +13,7 @@
     "url": "git+https://github.com/clearlydefined/service.git"
   },
   "dependencies": {
-    "@octokit/rest": "^14.0.0",
```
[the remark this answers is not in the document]
dabutvin: We probably should, there are ALOT of breaking changes.

Opened #302

providers/curation/github.js
dabutvin: this is who opened the PR, so as long as they are logged in it will be the contributor, bot the bot.

For example, https://github.com/clearlydefined/curated-data/pull/785

image

providers/curation/github.js
dabutvin: see above for snapshot of data captured.  

I went through each field trying to think about how a UI would be built against this data and what info it would need

routes/webhook.js
dabutvin: nice find!

I'll push an update

providers/curation/github.js
dabutvin: cool. I'll give it a go - it might get tricky with some of the nested objects, but could be a win

dabutvin:
build passed here: https://clearlydefined.visualstudio.com/ClearlyDefined/_build/results?buildId=3426

not sure why checks are not synced

PR #309, opened 2018-12-04
run spdx-correct in the summarizers

dabutvin:


providers/summary/clearlydefined.js
```diff
@@ -91,7 +91,7 @@ class ClearlyDescribedSummarizer {
         'licensed.declared',
         license
           .split('/')
-          .map(normalizeSpdx)
+          .map(correct)
```
jeffmcaffer: are you going to pass in the option to turn off upgrades?

I noticed this being called in multiple places. If we care about the params being passed in, should we have a helper method that everyone uses to ensure consistency? Perhaps revive `normalizeSPDX` but have it just do `correct` with the designated args.

providers/summary/scancode.js
```diff
@@ -82,7 +83,7 @@ class ScanCodeSummarizer {
 
   _toExpression(licenses) {
     if (!licenses) return null
-    const list = setToArray(licenses)
+    const list = setToArray(licenses).map(correct)
```
jeffmcaffer: Similarly here, turn off upgrades? or is that the default?

providers/summary/fossology.js
```diff
@@ -28,7 +29,8 @@ class FOSSologySummarizer {
     return files
       .map(file => {
         const path = get(/^File (.*?) contains/.exec(file), '[1]')
-        const license = normalizeSpdx(get(/license\(s\) (.*?)$/.exec(file), '[1]'))
+        let license = get(/license\(s\) (.*?)$/.exec(file), '[1]')
+        if (license) license = correct(license)
```
jeffmcaffer: and here

providers/summary/scancode.js
dabutvin: Ahh. Duh. Thanks for finding

PR #313, opened 2018-12-04
normalizeSpdx in cdSummarizer for interesting files

dabutvin:


providers/summary/clearlydefined.js
```diff
@@ -64,8 +64,16 @@ class ClearlyDescribedSummarizer {
...
 
   addInterestingFiles(result, data) {
+    if (!data.interestingFiles) return
     const newDefinition = cloneDeep(result)
-    setIfValue(newDefinition, 'files', data.interestingFiles)
+    set(
+      newDefinition,
+      'files',
+      data.interestingFiles.map(file => {
+        file.license = normalizeSpdx(file.license)
```
jeffmcaffer: need to use `setIfValue` here assuming that normalize can return null.

Also, this is more of a `forEach` scenario. we are just looping over the files and tweaking them.

` ...forEach(file => setIfValue(file, 'license', normalize(file.license)`

dabutvin: I want to set it to null

The case I am fixing is when file.license is 'NOASSERTION'

jeffmcaffer: There is a policy in the definitions that values are never null. properties either have a value or are not set. I've tried to do that everywhere and would like the schema to enforce. If we put in null here then the result of this summarizer would not be schema valid.

Not sure how no assertion (and the other non-license value) should be accounted for here. Perhaps we add these to normalize if they are real? Would be a bummer when we do things like make expressions (NOASSERTION AND MIT) just feels odd. Just dropping the NOASSERTION makes me wonder why we had it in the first place? That is, why is NOASSERTION important to retain in the single license case but not in the expression case?

Arrgh, the world would be simpler if people just did the right thing...

dabutvin: okay, got it - I'll update to not have a property instead of setting the property to null.

The NOASSERTION is in the harvested data, because it is an output from a tool - in this case licensee.

dabutvin: > Perhaps we add these to normalize if they are real?

That would not work in the current state, `statisfies()` does not accept NOASSERTION at all

PR #315, opened 2018-12-04
run spdxCorrect on npm manifest.license

dabutvin:
- some of these packages have 'SEE LICENSE IN...'
 - needs to be valid for computing scores


dabutvin:
a good example is https://github.com/ScottHamper/Cookies/blob/master/package.json#L7

jeffmcaffer:
Likely should not overwrite the entire declared prop.  Rather use setifvalue(result, And the whole path.  The function will create objects along the way
________________________________

dabutvin:
good point - opened https://github.com/clearlydefined/service/pull/317

PR #316, opened 2018-12-04
remove spdx-correct from normalizeSpdx

dabutvin:
spdx-correct is too lenient to be used for summarizers in clearlydefined.

This PR draws the line for fuzzy matches at casing and whitespace.

We will trim and normalize casing for spdx identifiers, other than that they will have to exactly match


dabutvin:
fyi @iamwillbar @iamwillbar

lib/utils.js
```diff
@@ -82,9 +82,12 @@ function extractLicenseFromLicenseUrl(licenseUrl) {
 }
 
 function normalizeSpdx(input) {
-  if (input) input = input.trim()
   if (!input) return null
-  return correct(input, { upgrade: false })
+  const loweredInput = input.toLowerCase().trim()
+  for (let spdx of spdxLicenseList) {
```
dabutvin: let's turn this into a lookup

lib/utils.js
```diff
@@ -82,9 +83,8 @@ function extractLicenseFromLicenseUrl(licenseUrl) {
 }
 
 function normalizeSpdx(input) {
-  if (input) input = input.trim()
   if (!input) return null
-  return correct(input, { upgrade: false })
+  return lowerSpdxLicenseMap.get(input.toLowerCase().trim()) || null
```
dabutvin: note: I think we would need to update this to handle expressions as input

lib/utils.js
```diff
@@ -82,9 +83,8 @@ function extractLicenseFromLicenseUrl(licenseUrl) {
 }
 
 function normalizeSpdx(input) {
```
jeffmcaffer: Cant see it now but @dabutvin commented that we should handle expressions here.

I agree. Basically we traffic in expressions everywhere so it would be great to have a solid function that cleans up expressions (including just simple ids). That way we don't really have to think hard, just normalize.

I't might be that what we actually do is augment the spdx-expression-parser to take a *visitor* that processes each leaf node in the expression AST.  We would pass in what is effectively this function now, other users could pass in spdx-correct, ... I don't remember the architecture of the parser but seems like it should be possible.

dabutvin: Yeah that comment disappeared ...

Makes a lot of sense, instead of making the parser opiniated.

dabutvin:
@jeffmcaffer take a look at https://github.com/clearlydefined/service/blob/c5e9f92412791e7345070f604ca3d47ce3a9784d/lib/spdx.js and let me know what you think

providers/summary/scancode.js
```diff
@@ -85,7 +86,7 @@ class ScanCodeSummarizer {
     const list = setToArray(licenses)
     if (!list) return null
     return list
-      .map(normalizeSpdx)
+      .map(SPDX.normalize)
       .filter(x => x)
       .join(' AND ')
```
dabutvin: the work to make this right is in #322 

Once we get through this piece here, I can update that branch

dabutvin:
good point @jeffmcaffer I updated to only add parens around nested OR conjunctions.

PR #319, opened 2018-12-06
add test for filtering invalid licenses from interestingFiles

dabutvin:


test/providers/summary/clearlydefined.js
```diff
@@ -120,6 +120,34 @@ describe('ClearlyDescribedSummarizer addLicenseFromFiles', () => {
   })
 })
 
+describe('ClearlyDescribedSummarizer addInterestingFiles', () => {
+  it('should filter invalid license properties', () => {
+    const data = new Map([
+      [{ path: 'LICENSE', token: 'abcd', license: 'MIT' }, { path: 'LICENSE', token: 'abcd', license: 'MIT' }],
```
jeffmcaffer: Cool. FWIW, I think there is an acutally Mocha test construct for running the same set of tests over a set of data. The collection of data is put outside the tests and then the test just says run over these. No particular need to change here, just thought I'd point it out.

PR #321, opened 2018-12-07
use licenseExpression from nuget manifest to declare license

dabutvin:
based on https://github.com/NuGet/Home/wiki/Packaging-License-within-the-nupkg-(Technical-Spec)#license-expression-pack


dabutvin:
there's a mention of `PackageLicenseExpressionVersion` in the spec too, but I don't think anyone is using it or is supposed to

Issue #226, opened 2018-09-03
Design "big picture" data source


jeffmcaffer:
as a caller of the CD apis I should be able to get a bigger picture of the data in ClearlyDefined. In particular:

* set of merged or pending curations across all (or a set of) components
* set of components for which we have data

This is primarily in support of user workflows where we want to show the user curations and definitions related to what they are currently observing.

First step is to create design doc that captures the scenarios and possible approaches.

jeffmcaffer:
I started a design doc on this. See https://docs.google.com/document/d/1pB5ue8It7mRO1K4tesLGMTXcbZWVg7Q8N0ibTIft1fA/edit#

Teasing out the scenarios and the API/Data that would be needed to address them.

jeffmcaffer:
The basics are done, implemented and in prod now. All definitions are stored in a doc db. next we need to add the PR data. That is captured in #253

PR #323, opened 2018-12-11
add updated property to defintion

dabutvin:
comes through in mongo as 

image

### DIFF
--- a/business/definitionService.js
+++ b/business/definitionService.js
@@ -128,6 +128,7 @@ class DefinitionService {
   }
 
   async _store(definition) {
+    definition.updated = new Date()
     await this.definitionStore.store(definition)
     return this.search.store(definition)
   }


dabutvin:
we can do queries like this

```
db.getCollection('definitions').find({ updated: { $gt: new Date('2018-12-11') } })
```

or more precisesly

```
db.getCollection('definitions').find({ updated: { $gt: new Date('2018-12-11T16:31:00.000Z') } })
```

business/definitionService.js
```diff
@@ -128,6 +128,7 @@ class DefinitionService {
   }
 
   async _store(definition) {
+    definition.updated = new Date()
```
jeffmcaffer: A few questions about this
* should we be adding this to the schema? If the tests are not failing now then the schema is likely wrong. Should not allow additional properties
* In the future we might have more of these sort of `meta` properties. Wanna group them into a parent prop? (e.g., _meta or some such? Suggest we do NOT use _metadata as that will conflict/confuse in some of our store providers)

dabutvin:
re: date

I think we should change to just straight up json.

here's some background if you're interested from [source](https://www.compose.com/articles/from-json-to-bson-and-back/)

> When working with MongoDB dates, ISODate only appears in the shell and can only be consumed by the shell. Everywhere else, Date's are a Date instance. With BSON Dates being stored, date fields become an easily sortable, comparable and indexable number for the database.

It changes the way the data is exported

given:
```
{ "created": new Date("Feb 12 2014 15:28:31 GMT") }
``` 

will get exported as:
```
{
  "created": { "$date": 1392218911000 }
}
```

which is not what we would want

business/definitionService.js
dabutvin: added to schema and grouped in `_meta` object

dabutvin:
Now the data looks better. Good call

image

dabutvin:
the queries are now

```
db.getCollection('definitions').find({ '_meta.updated': { $gt: '2018-12-13' } })
```

```
db.getCollection('definitions').find({ '_meta.updated': { $gt: '2018-12-13T20:33:00' } })
```

storrisi:
Seems like this change of schema has introduced a bug on the UI.
As mentioned in [this doc](https://docs.google.com/document/d/1pB5ue8It7mRO1K4tesLGMTXcbZWVg7Q8N0ibTIft1fA/edit#) from issue #226 , the _meta property was expected to be a different type of object.

I'm gonna fix it with a quick PR on the website.

PR #326, opened 2018-12-12
update rubygem summarizer to read licenses from registryData

dabutvin:
see spec https://guides.rubygems.org/specification-reference/#license=

There is `license (string)` and `licenses (array)`

Example of a dual licensed: https://rubygems.org/api/v1/gems/mocha.json

Interpreting to join with `OR`

image

### DIFF
--- a/test/summary/clearlyDefined.js
+++ b/test/summary/clearlyDefined.js
@@ -7,6 +7,7 @@ const { expect } = chai
 const Summarizer = require('../../providers/summary/clearlydefined')
 const { setIfValue } = require('../../lib/utils')
 const EntityCoordinates = require('../../lib/entityCoordinates')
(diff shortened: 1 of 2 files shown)


test/summary/clearlyDefined.js
```diff
@@ -266,11 +282,11 @@ describe('ClearlyDefined Gem summarizer', () => {
   })
 })
 
-function setupGem(releaseDate, license, sourceInfo) {
+function setupGem(releaseDate, licenses, sourceInfo) {
```
jeffmcaffer: propose just have this take a string or an array for the second arg. If it is a string then set the `license` prop. If an array, set the `licenses` prop

dabutvin: done

PR #329, opened 2018-12-14
add missing rust identifiers to curations schema

dabutvin:


jeffmcaffer:
To avoid this in the future, can we pull the relevant lists from the schema so that things just don't work if they are not in the schema? Or design up some sort of test? Or remove it from the schema and just let the system work with whatever it works with and reject the others? 

I'm also wondering a bit about how they system will be extensible when/if we start handling a lot more different providers and types and they are effectively just plugged in.

PR #353, opened 2019-01-15
Integrate chive to generate notice files

dabutvin:
See [chive](https://github.com/amzn/tiny-attribution-generator)

Example: 

POST to https://api.clearlydefined.io/notices
body:
```json
{
	"coordinates": [
		"npm/npmjs/-/lodash/4.17.11",
		"npm/npmjs/-/junk/5.88.0",
		"npm/npmjs/@ant-design/icons/1.1.15",
		"nuget/nuget/-/newtonsoft.json/12.0.1"
	],
	"output": "text|html|template|json",
	"options": {"template": "<optional handlebars template>"}	
}
```

see https://github.com/amzn/tiny-attribution-generator/blob/master/default-htmltemplate.hbs for a template example

We will use the default textRenderer in chive if one is not provided

### DIFF
--- a/business/noticeService.js
+++ b/business/noticeService.js
@@ -20,28 +20,26 @@ class NoticeService {
     const renderer = template ? new TemplateRenderer(template) : new TextRenderer()
     const docbuilder = new DocBuilder(renderer)
     const result = await this.definitionService.getAll(coordinates)
-    const source = new JsonSource(
-      JSON.stringify({
-        packages: await Promise.all(
-          Object.keys(result).map(async id => {
-            const definition = result[id]
-            return {
-              name: [definition.coordinates.namespace, definition.coordinates.name].filter(x => x).join('/'),
-              version: get(definition, 'coordinates.revision'),
-              license: get(definition, 'licensed.declared'),
-              copyrights: get(definition, 'licensed.facets.core.attribution.parties'),
-              website: get(definition, 'described.projectWebsite') || '',
-              text: await this._getPackageText(definition)
-            }
-          })
-        )
+    const packages = await Promise.all(
+      Object.keys(result).map(async id => {
+        const definition = result[id]
+        return {
+          name: [definition.coordinates.namespace, definition.coordinates.name].filter(x => x).join('/'),
+          version: get(definition, 'coordinates.revision'),
+          license: get(definition, 'licensed.declared'),
+          copyrights: get(definition, 'licensed.facets.core.attribution.parties'),
+          website: get(definition, 'described.projectWebsite') || '',
+          text: await this._getPackageText(definition)
+        }
       })
     )
+    const source = new JsonSource(JSON.stringify({ packages: packages.filter(x => x.license) }))
     await docbuilder.read(source)
     return docbuilder.build()
   }
 
   async _getPackageText(definition) {
+    if (!definition.files) return ''
     const texts = await Promise.all(
       definition.files
         .filter(file => file.token && file.natures && file.natures.includes('license'))
--- a/business/noticeService.js
+++ b/business/noticeService.js
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: MIT
 
 const { get } = require('lodash')
-const validator = require('../schemas/validator')
 const DocBuilder = require('tiny-attribution-generator/lib/docbuilder').default
 const TextRenderer = require('tiny-attribution-generator/lib/outputs/text').default
 const TemplateRenderer = require('tiny-attribution-generator/lib/outputs/template').default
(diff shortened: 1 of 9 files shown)


business/noticeService.js
```diff
@@ -0,0 +1,52 @@
...
+
+class NoticeService {
+  constructor(definitionService, attachmentStore) {
+    this.definitionService = definitionService
+    this.attachmentStore = attachmentStore
+    this.logger = logger()
+  }
+
+  async generate(coordinates, template = '') {
+    const renderer = template ? new TemplateRenderer(template) : new TextRenderer()
```
jeffmcaffer: question: does the template say what kind of renderer to use? For example, what if we want to use the JSON renderer? Feels like we should be saying what kind of render and then providing parameters to that. This would also apply at the REST API level

business/noticeService.js
```diff
@@ -0,0 +1,52 @@
...
+    await docbuilder.read(source)
+    return docbuilder.build()
+  }
+
+  async _getPackageText(definition) {
+    if (!definition.files) return ''
+    const texts = await Promise.all(
+      definition.files
+        .filter(file => file.token && file.natures && file.natures.includes('license'))
+        .map(file => this.attachmentStore.get(file.token))
```
jeffmcaffer: might want a throat somewhere in here (or higher) as this will flood the store with requests.

routes/notices.js
```diff
@@ -0,0 +1,26 @@
...
+
+const asyncMiddleware = require('../middleware/asyncMiddleware')
+const express = require('express')
+const router = express.Router()
+const EntityCoordinates = require('../lib/entityCoordinates')
+const validator = require('../schemas/validator')
+
+router.post(
+  '/',
+  asyncMiddleware(async (request, response) => {
```
jeffmcaffer: suggest using the new style of having the handler as a separate method that is just referenced from the router call.  Looks better, easier to read and more testable.

schemas/notice-request.json
```diff
@@ -0,0 +1,26 @@
+{
```
jeffmcaffer: Love it!

schemas/notice-request.json
```diff
@@ -0,0 +1,26 @@
...
+  "$id": "https://api.clearlydefined.io/schemas/notice-request.json#",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "noticeRequest",
+  "type": "object",
+  "additionalProperties": false,
+  "required": ["coordinates"],
+  "properties": {
+    "coordinates": {
+      "type": "array",
+      "errorMessage": {
```
jeffmcaffer: this should have an `"items": { "type": "string"}` entry

business/noticeService.js
```diff
@@ -0,0 +1,52 @@
...
+    this.logger = logger()
+  }
+
+  async generate(coordinates, template = '') {
+    const renderer = template ? new TemplateRenderer(template) : new TextRenderer()
+    const docbuilder = new DocBuilder(renderer)
+    const result = await this.definitionService.getAll(coordinates)
+    const packages = await Promise.all(
+      Object.keys(result).map(async id => {
+        const definition = result[id]
```
jeffmcaffer: what do we do when definitions are missing? the `getAll` above I think will return an entry but it will be largely empty. How/where do we accumulate all the errors etc? Would be good for the response to have that as well.

business/noticeService.js
dabutvin: I was having that thought too when putting this together. I was actually wondering, should the attachmentStore itself should be the one throttling requests instead of every caller of the store?

The callers don't need to know that it is blob storage vs some db query or something which would have different tolerances.

business/noticeService.js
dabutvin: let's implement all of chive and it's options

business/noticeService.js
dabutvin: let's implement all of chive and it's options

business/noticeService.js
dabutvin: let's return the json instead of the raw content with the summary of what happened

dabutvin: problem buckets

 1. no license declared
 2. no tools run - meaning does not exist
 3. no copyright for component

business/noticeService.js
dabutvin: let's put bottleneck in the attachment store set to ~1000


```js
get = limiter.wrap(token => {
   store....
})
```

business/noticeService.js
dabutvin: image

PR #361, opened 2019-01-22
remove missing property from noticeService

dabutvin:
It looked like a bug to me when I looked at it, maybe others will too?

@jeffmcaffer what do you think?

### DIFF
--- a/business/noticeService.js
+++ b/business/noticeService.js
@@ -30,7 +30,7 @@ class NoticeService {
       content,
       summary: {
         total: coordinates.length,
-        missing: packages.noDefinition.length,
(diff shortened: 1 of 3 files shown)


business/noticeService.js
```diff
@@ -30,7 +30,7 @@ class NoticeService {
       content,
       summary: {
         total: coordinates.length,
-        missing: packages.noDefinition.length,
```
jeffmcaffer: prehaps we should just remove `missing` altogether, `warnings.noDefinition.length` is equivalent to the original and your new version could also subtract `warnings.noLicense.length`

dabutvin:
okay. I updated to remove it altogether

PR #363, opened 2019-01-25
add telemetry to defintionService

dabutvin:
three new log statements:
  - 'current definition exists',
  - 'definition is available'
  - 'definition not available'

we want to be able to query the percentage of requests that we have
information available for. This is the sum of
'current definition exists' and 'definition is available' divided by
the sum of all three

### DIFF
--- a/business/definitionService.js
+++ b/business/definitionService.js
@@ -59,7 +59,7 @@ class DefinitionService {
     const existing = force ? null : await this.definitionStore.get(coordinates)
     let result
     if (get(existing, '_meta.schemaVersion') === currentSchema) {
-      this.logger.info('current definition exists', { coordinates: coordinates.toString() })
+      this.logger.info('computed definition available', { coordinates: coordinates.toString() })
       result = existing
     } else result = await this.computeAndStore(coordinates)
     return this._cast(result)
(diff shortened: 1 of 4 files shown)


business/definitionService.js
```diff
@@ -57,8 +57,11 @@ class DefinitionService {
       return this.compute(coordinates, curation)
     }
     const existing = force ? null : await this.definitionStore.get(coordinates)
-    const result =
-      get(existing, '_meta.schemaVersion') === currentSchema ? existing : await this.computeAndStore(coordinates)
+    let result
+    if (get(existing, '_meta.schemaVersion') === currentSchema) {
+      this.logger.info('current definition exists', { coordinates: coordinates.toString() })
```
jeffmcaffer: perhaps the following three messages to capture all the states:

* "computed definition available"
* "recomputed definition available"
* "definition not available"

Question, do you think we'll ever care about the difference between the first time we compute a definition (e.g., `existing === null`) and the N times we recompute? And for recompute, how often it is becuase the schema was out of date vs force === true?

dabutvin: force=true should only ever be in local dev.

Maybe we could start to track when definitions are invalidated?

PR #364, opened 2019-01-29
paginate the defintions in the mongo store (fixes #254)

### LINKED ISSUE #254: API error for heavy definitions
If you are checking for a very large definition, like: https://dev.clearlydefined.io/share/eJwty0EOgyAQheG7zNqFlqrIstdoXADOtJMoYwBNjPHuxaTL9+X9JxDPGSOY86ogScyvA0zY5rkCLxInDjZjAvM+IR8rgoEPZ6hgjbLzdIc3fDdXLNgF02r9/VrYR0lC+e+F9uRlwrIj7pxYQrHnUCP1XeP6zilSTmn1QKLetlpb6/XQYEuoa7jG6wfYiziO

 while getting the data the /definitions API returns an error: `Request size is too large`

dabutvin:
For documents that have more than 1000 files, we will split them across different documents in the store.

This should be transparent to any callers of the store. we will get and set as one logical document.

Things to note:
 - adding a partitionKey set to the canonical coordinates so we can keep multiple pages of a single document on a single partition
 - storing currentPage and totalPages as metadata on the document in a `_mongo` property specific to this store and not surfaced as part of the definition
 - working with some larger documents surfaced some bugs in our harvested data around the file lists, we are now treating the clearlydefined tool as the source of truth for absolute file list

### DIFF
--- a/business/aggregator.js
+++ b/business/aggregator.js
@@ -30,7 +30,7 @@
 //
 
 const { getLatestVersion, mergeDefinitions } = require('../lib/utils')
-const { flattenDeep, get, set, differenceBy, intersectionBy } = require('lodash')
+const { flattenDeep, get, set, intersectionBy } = require('lodash')
 const logger = require('../providers/logging/logger')
 
 class AggregationService {
@@ -77,10 +77,10 @@ class AggregationService {
   _normalizeFiles(result, summarized, coordinates) {
     const cdFiles = get(this._findData('clearlydefined', summarized), 'summary.files')
     if (!cdFiles) return
-    const difference = differenceBy(result.files, cdFiles, 'path')
-    if (!difference.length) return
+    const difference = result.files.length - cdFiles.length
+    if (!difference) return
     this.logger.info('difference between summary file count and cd file count', {
-      count: difference.length,
+      count: difference,
       coordinates: coordinates.toString()
     })
     result.files = intersectionBy(result.files, cdFiles, 'path')
--- a/business/aggregator.js
+++ b/business/aggregator.js
@@ -30,7 +30,8 @@
 //
 
 const { getLatestVersion, mergeDefinitions } = require('../lib/utils')
-const { flattenDeep, set } = require('lodash')
+const { flattenDeep, get, set, differenceBy, intersectionBy } = require('lodash')
+const logger = require('../providers/logging/logger')
 
 class AggregationService {
   constructor(options) {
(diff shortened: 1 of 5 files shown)


business/aggregator.js
```diff
@@ -66,6 +69,22 @@ class AggregationService {
...
+  _normalizeFiles(result, summarized, coordinates) {
+    const cdFiles = get(this._findData('clearlydefined', summarized), 'summary.files')
+    if (!cdFiles) return
+    const difference = differenceBy(result.files, cdFiles, 'path')
+    if (!difference.length) return
+    this.logger.info('difference between summary file count and cd file count', {
+      count: difference.length,
+      coordinates: coordinates.toString()
+    })
+    result.files = intersectionBy(result.files, cdFiles, 'path')
```
jeffmcaffer: order sometimes matters in these operations. have you looked at that?

Also, can we get the difference length by taking `result.files.length - intersection.length`? At least in theory that should be the same. Would just save us some time on larger file lists.

providers/stores/mongo.js
```diff
@@ -49,18 +50,50 @@ class MongoStore {
    * @param {Coordinates} coordinates - The coordinates of the object to get
    * @returns The loaded object
    */
-  get(coordinates) {
-    return this.collection.findOne({ _id: this._getId(coordinates) }, { projection: { _id: 0 } })
+  async get(coordinates) {
+    const cursor = await this.collection.find(
+      { _id: new RegExp(`^${this._getId(coordinates)}`) },
+      { projection: { _id: 0, _mongo: 0 }, sort: { _id: 1 } }
```
jeffmcaffer: will this sorting work for docs that have more than 10 pages?
```
../../1
../../10
../../11
../../2
../../3
```
perhaps we should just get the all the pages, then `sortBy(.. mongo.currentPage`) and `flatten`

providers/stores/mongo.js
```diff
@@ -49,18 +50,50 @@ class MongoStore {
...
+          if (index === 0) {
+            const definitionPage = clone(definition)
+            if (definition.files) definitionPage.files = definition.files.slice(0, pageSize)
+            return { ...definitionPage, _mongo: { partitionKey: definition._id, currentPage: 1, totalPages: pages } }
+          }
+          return {
+            _id: definition._id + `/${index}`,
+            _mongo: {
+              partitionKey: definition._id,
+              currentPage: index + 1,
```
jeffmcaffer: perhaps `page` rather than `currentPage`. we do not have a "next" or "previous"

providers/stores/mongo.js
dabutvin: I found that I can sort by `_mongo.page` on the server and it respects number order
sorting on the database would be better I think. 

I also updated the queries to be by partitionKey instead of regexing the IDs

jeffmcaffer:
Cool. Ship it when you get those changes in

PR #365, opened 2019-01-30
swallow the error unless we are in dev

dabutvin:
dev 404:
image
prod 404:
image

dev error from a service:
image

prod error from a service:
image


jeffmcaffer:
That's a good change but I think there is still an issue. The code that calls a service internally I think does not always throw. `requestPromise...` may simply return a response that has a status code like 401. in the current code we just return the result of `requestPromise`. that is a leaky abstraction. 

IIRC some of the request libraries have ways of controlling the throwing of errors vs returning status codes. We should ensure that the right behavior is happening and either way, should likely not return the result of the `requestPromise` lest someone start thinking that they could/should use it.

PR #370, opened 2019-02-01
implement getAll in the definition store

dabutvin:
- instead of going 1 by 1 to mongo we do a $in for the whole list
- the blob and file implementations still go 1 by 1

The thinking is that we will get big performance gains by doing 1 roundtrip to the DB instead of many little trips. This uses the `cursor.forEach()` to stream the results.

Any coordinates that are not found we call the original get() with force=true in order to compute, store, and return the definition if available.

### DIFF
--- a/business/definitionService.js
+++ b/business/definitionService.js
@@ -80,10 +80,19 @@ class DefinitionService {
    * @returns A list of all components that have definitions and the definitions that are available
    */
   async getAll(coordinatesList, force = false) {
-    const result = {}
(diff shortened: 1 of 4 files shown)


dabutvin:
I want to make sure this gets tested more thoroughly before it ships. It was an idea I stumbled upon and wanted to get it down before I lost it

dabutvin:
closing as this is outdated and didn't turn out to be worth the trouble

PR #373, opened 2019-02-02
Add find endpoint to defintions

dabutvin:
POST to `/definitions/find`

see schema https://github.com/dabutvin/service/blob/5fee86e15f21e7ae80911567fd3b114511968f2e/schemas/definitions-find.json

### DIFF
--- a/providers/stores/abstractAzblobStore.js
+++ b/providers/stores/abstractAzblobStore.js
@@ -4,6 +4,7 @@
 const azure = require('azure-storage')
 const AbstractFileStore = require('./abstractFileStore')
 const logger = require('../logging/logger')
+const EntityCoordinates = require('../../lib/entityCoordinates')
 
 const { promisify } = require('util')
 
@@ -64,12 +65,52 @@ class AbstractAzBlobStore {
     }
   }
 
+  /**
+   * Query and return the objects based on the query
+   *
+   * @param {object} query - The filters and sorts for the request
+   * @returns The data and continuationToken if there is more results
+   */
+  async find(query, continuationToken = '') {
+    const prefix = this._buildPrefix(query)
+    const result = await promisify(this.blobService.listBlobsSegmentedWithPrefix).bind(this.blobService)(
+      this.containerName,
+      prefix,
+      continuationToken,
+      {
+        include: azure.BlobUtilities.BlobListingDetails.METADATA
+      }
+    )
+    const data = []
+    result.entries
+      .forEach(entry => {
+        const path = entry.metadata.id
+        if (!path) return
+        data.push(this.get(EntityCoordinates.fromString(path)))
+      })
+      .filter(x => x)
+    return { data, continuationToken: result.continuationToken }
+  }
+
   _toStoragePathFromCoordinates(coordinates) {
     return AbstractFileStore.toStoragePathFromCoordinates(coordinates)
   }
 
   _toResultCoordinatesFromStoragePath(path) {
     return AbstractFileStore.toResultCoordinatesFromStoragePath(path)
+  }
+
+  _buildPrefix(parameters) {
+    let result = ''
+    if (!parameters.type) return result
+    result += `${parameters.type}/`
+    if (!parameters.provider) return result
+    result += `${parameters.provider}/`
+    if (!parameters.namespace) return result
+    result += `${parameters.namespace}/`
+    if (!parameters.name) return result
+    result += `${parameters.name}/`
+    return result
   }
 }
 
--- a/providers/stores/abstractAzblobStore.js
+++ b/providers/stores/abstractAzblobStore.js
@@ -4,7 +4,6 @@
 const azure = require('azure-storage')
 const AbstractFileStore = require('./abstractFileStore')
 const logger = require('../logging/logger')
-const EntityCoordinates = require('../../lib/entityCoordinates')
 
 const { promisify } = require('util')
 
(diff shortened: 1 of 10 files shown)


providers/stores/abstractFileStore.js
```diff
@@ -55,6 +55,42 @@ class AbstractFileStore {
...
+  /**
+   * Query and return the objects based on the query
+   *
+   * @param {object} query - The filters and sorts for the request
+   * @returns The data and continuationToken if there is more results
+   */
+  async find(query) {
+    const paths = await recursive(this.options.location, ['.DS_Store'])
+    return (await Promise.all(
+      paths.map(async path => {
```
jeffmcaffer: nit: could just be the `filter` here

providers/stores/abstractAzblobStore.js
```diff
@@ -64,13 +65,53 @@ class AbstractAzBlobStore {
...
   }
 
   _toResultCoordinatesFromStoragePath(path) {
     return AbstractFileStore.toResultCoordinatesFromStoragePath(path)
   }
+
+  _buildPrefix(parameters) {
+    let result = ''
+    if (!parameters.type) return result
+    result += `${parameters.type}/`
```
jeffmcaffer: interesting problem. Since we can only do prefix in blob listing all the parts need to be provided.

Meta: I wonder if all providers need to implement `find`. Perhaps here we just throw saying "find is not supported"

providers/stores/mongo.js
```diff
@@ -63,6 +60,29 @@ class MongoStore {
...
+    const sort = this._buildSort(query)
+    const cursor = await this.collection.find(filter, {
+      projection: { _id: 0, files: 0 },
+      sort: sort,
+      limit: pageSize
+    })
+    const data = await cursor.toArray()
+    continuationToken = this._getContinuationToken(pageSize, data)
+    data.forEach(def => {
+      delete def._mongo
```
jeffmcaffer: can we use the `projection` here to remove `_mongo` or does that mess up the query?

providers/stores/mongo.js
```diff
@@ -103,6 +123,61 @@ class MongoStore {
...
+    if (parameters.maxLicensedScore) filter['licensed.score.total'] = { $lt: parseInt(parameters.maxLicensedScore) }
+    if (parameters.minDescribedScore) filter['described.score.total'] = { $gt: parseInt(parameters.minDescribedScore) }
+    if (parameters.maxDescribedScore) filter['described.score.total'] = { $lt: parseInt(parameters.maxDescribedScore) }
+    if (continuationToken) filter['_mongo.partitionKey'] = { $gt: base64.decode(continuationToken) }
+    return filter
+  }
+
+  _buildSort(parameters) {
+    let sort
+    switch (parameters.sort) {
```
jeffmcaffer: feels like this could be a table lookup

routes/definitions.js
```diff
@@ -79,6 +79,13 @@ router.post(
   })
 )
 
+router.post('/find', asyncMiddleware(findDefinitions))
```
jeffmcaffer: 'find' feels not RESTful. Couple of thoughts

* could we do GET with query parameters?
* if POST is the way to go, can we hang this off `/` and basically have two different body formats, the array of coords, and this new one.

dabutvin: makes sense. Let's do the GET with query parameters

dabutvin: makes sense. Let's do the POST and check for array vs object for the query vs list.
The GET with query parameters is used to list coordinates from search

providers/stores/mongo.js
```diff
@@ -103,6 +124,46 @@ class MongoStore {
...
+    if (continuationToken) filter['_mongo.partitionKey'] = { $gt: base64.decode(continuationToken) }
+    return filter
+  }
+
+  _buildSort(parameters) {
+    const sort = this.sortOptions[parameters.sort] || '_mongo.partitionKey'
+    return { [sort]: parameters.sortDesc ? -1 : 1 }
+  }
+
+  _getSortOptions() {
```
jeffmcaffer: Perhaps just make this a constant in the file? Put this up around L9

```javascript
const sortOptions = {
  type: 'coordinates.type',
  provider: 'coordinates.provider',
  name: 'coordinates.name',
  namespace: 'coordinates.namespace',
  license: 'licensed.declared',
  releaseDate: 'described.releaseDate',
  licensedScore: 'licensed.score.total',
  describedScore: 'described.score.total'
}
```

routes/definitions.js
```diff
@@ -69,15 +69,25 @@ router.post(
...
   '/',
   asyncMiddleware(async (request, response) => {
-    const coordinatesList = request.body.map(entry => EntityCoordinates.fromString(entry))
-    if (coordinatesList.length > 1000)
-      return response.status(400).send(`Body contains too many coordinates: ${coordinatesList.length}`)
-    // if running on localhost, allow a force arg for testing without webhooks to invalidate the caches
-    const force = request.hostname.includes('localhost') ? request.query.force || false : false
-    const result = await definitionService.getAll(coordinatesList, force)
-    response.status(200).send(result)
+    if (Array.isArray(request.body)) await listDefinitions(request, response)
```
jeffmcaffer: why `await` here. can you just return the promise that comes from the *Definitions function?

routes/definitions.js
```diff
@@ -69,15 +69,25 @@ router.post(
...
+  // if running on localhost, allow a force arg for testing without webhooks to invalidate the caches
+  const force = request.hostname.includes('localhost') ? request.query.force || false : false
+  const result = await definitionService.getAll(coordinatesList, force)
+  response.status(200).send(result)
+}
+
+async function findDefinitions(request, response) {
+  if (!validator.validate('definitions-find', request.body)) return response.status(400).send(validator.errorsText())
+  const result = await definitionService.find(request.body)
+  response.send(result)
```
jeffmcaffer: no status code? L83 has one.

schemas/definitions-find.json
```diff
@@ -0,0 +1,99 @@
+{
+  "$id": "https://api.clearlydefined.io/schemas/definitions-find.json#",
```
jeffmcaffer: probably should look at the swagger as well. don't remember where we are on using the JSON schema in the swagger.

dabutvin:
thanks for the review.
I am going to look at getting the suggestions service up to date now

PR #376, opened 2019-02-05
update mongo list() to find multiple revisions (fixes #375)

### LINKED ISSUE #375: Curation endpoint fails with a 400 error
This happens only in the dev environment for now.
Any kind of contribution fails with a 400 error, without any other additional message.

Payload example: 
`{"contributionInfo":{"summary":"AUTOMATION TEST","details":"AUTOMATION TEST","resolution":"AUTOMATION TEST","type":"missing","removeDefinitions":false},"patches":[{"coordinates":{"type":"npm","provider":"npmjs","name":"async"},"revisions":{"2.6.1":{"licensed":{"declared":"MIT-0"}}}}]}`

dabutvin:
- also remove the try/catch/400 and let the app errorhandler take this

### DIFF
--- a/providers/stores/mongo.js
+++ b/providers/stores/mongo.js
@@ -46,7 +46,7 @@ class MongoStore {
    */
   async list(coordinates) {
     const list = await this.collection.find(
-      { '_mongo.partitionKey': this._getId(coordinates), '_mongo.page': 1 },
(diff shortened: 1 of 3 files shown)


dabutvin:
@storrisi fyi

PR #389, opened 2019-02-09
setup redis caching provider

dabutvin:
- set up the service `config.caching.service()` as `memory` or `redis` and attach this one with the middleware
- plug in the cache in the definitionService
- use memorycache explicitly for github permissions
- compress the data going into redis with [pako](https://www.npmjs.com/package/pako)
- put 1hr default expiration on the memorycache


business/definitionService.js
```diff
@@ -190,6 +200,7 @@ class DefinitionService {
 
   async _store(definition) {
     await this.definitionStore.store(definition)
+    await this.cache.set(this._getCacheKey(definition.coordinates), definition)
```
jeffmcaffer: suggest putting the cache last. failing to put it in the cache would stop it going into the search which would be a functional problem (users could not find the definition). Failure to put it in the cache will just be a perf issue that should self-correct (if we add it when missing. see other comment above).

business/definitionService.js
```diff
@@ -63,7 +64,9 @@ class DefinitionService {
       const curation = this.curationService.get(coordinates, pr)
       return this.compute(coordinates, curation)
     }
-    const existing = force ? null : await this.definitionStore.get(coordinates)
+    const existing = force
+      ? null
+      : (await this.cache.get(this._getCacheKey(coordinates))) || (await this.definitionStore.get(coordinates))
```
jeffmcaffer: if it exists and is all ok and is not in the cache, do we want to add it? Otherwise, once something is ejected, it will never get back in and MRU will be broken

business/definitionService.js
```diff
@@ -498,7 +509,16 @@ class DefinitionService {
   _getDefinitionCoordinates(coordinates) {
     return Object.assign({}, coordinates, { tool: 'definition', toolVersion: 1 })
   }
+
+  _getCacheKey(coordinates) {
+    return (
+      'co_' +
```
jeffmcaffer: nit but as a clarity thing, the cache key should relate to the type of the value not the type of the key. So here perhaps `'def_'` rather than `'co_'` as this is the key used to store definitions in the flat cache namespace.

middleware/github.js
```diff
@@ -56,11 +56,11 @@ async function setupUserClient(req, token) {
 
 // Get GitHub user info and attach it to the request
 async function setupInfo(req, cacheKey, client) {
-  let info = await req.app.locals.cache.get(cacheKey)
+  let info = await options.cache.get(cacheKey)
```
jeffmcaffer: Looking at this now I wonder if the cache could/should be passed in explicitly rather than as part of `options`. I don't recall why it is not. Just feels odd. Everything else in options is really just data, not functional.

Concretely I'm suggesting that the factory function at the end take an additional arg which is the cache service to use (or a factory or options to use to create the cache). Not sure this is a huge deal.  Just sticks out a bit.

Also, if we are using the injected option, then we don't need the middleware. The middleware just adds `req.app.local.cache` but all uses of that are now using the injected options object.

middleware/githubConfig.js
```diff
@@ -4,19 +4,18 @@ const config = require('painless-config')
...
   org: config.get('AUTH_GITHUB_ORG') || 'clearlydefined',
-  timeouts: {
-    info: 10 * 60 // 10 mins
-  },
   permissions: {
     harvest: [config.get('AUTH_HARVEST_TEAM') || 'harvest-dev'],
     curate: [config.get('AUTH_CURATION_TEAM'), 'curation-dev']
-  }
+  },
+  cache: memoryCache({ defaultExpirationSeconds: 10 * 60 /* 10 mins */ })
```
jeffmcaffer: yeah, so down here (actually L23) perhaps we pass along the cache in addition to the declarative options rather than adding the cache into options.

providers/caching/redis.js
```diff
@@ -3,31 +3,48 @@
...
 
   async set(item, value, expirationSeconds) {
     if (typeof value !== 'string') value = objectPrefix + JSON.stringify(value)
-    await this._redisSet(item, value, 'EX', expirationSeconds)
+    const data = pako.deflate(value, { to: 'string' })
+    if (expirationSeconds) await this._redisSet(item, data, 'EX', expirationSeconds)
+    else await this._redisSet(item, data)
+  }
+
+  async delete(item) {
```
dabutvin: Need to implement this in memory cache

middleware/githubConfig.js
jeffmcaffer: `ttlSeconds` rather than `defaultExpirationSeconds`

providers/caching/memory.js
```diff
@@ -4,17 +4,21 @@
...
 
+  initialize() {}
+
   async get(item) {
     return this.cache.get(item)
   }
 
-  async set(item, value, expirationSeconds) {
-    this.cache.put(item, value, 1000 * expirationSeconds)
+  async set(item, value, expirationSeconds = null) {
```
jeffmcaffer: `ttlSeconds`?

providers/caching/memoryConfig.js
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+
+const memory = require('./memory')
+
+function serviceFactory() {
+  return memory({ defaultExpirationSeconds: 60 * 60 })
+}
+
+module.exports = serviceFactory
```
jeffmcaffer: memory.js can be its own factory. It already is one. Just need to add the default options in memory.js

providers/caching/redis.js
```diff
@@ -3,31 +3,48 @@
...
+    this._redisSet = util.promisify(this.redis.set).bind(this.redis)
+    this._redisDel = util.promisify(this.redis.del).bind(this.redis)
   }
 
   async get(item) {
     const cacheItem = await this._redisGet(item)
-    if (cacheItem && cacheItem.startsWith(objectPrefix)) {
+    if (!cacheItem) return null
+    const result = pako.inflate(cacheItem, { to: 'string' })
+    if (result.startsWith(objectPrefix)) {
```
jeffmcaffer: would invert the logic here to exit fast. 

```js
if (!result.startsWith(objectPrefix)) return result
```

providers/caching/redis.js
```diff
@@ -3,31 +3,48 @@
...
     }
-    return cacheItem
+    return result
   }
 
   async set(item, value, expirationSeconds) {
     if (typeof value !== 'string') value = objectPrefix + JSON.stringify(value)
-    await this._redisSet(item, value, 'EX', expirationSeconds)
+    const data = pako.deflate(value, { to: 'string' })
+    if (expirationSeconds) await this._redisSet(item, data, 'EX', expirationSeconds)
```
jeffmcaffer: with no expiration will they just monotonically fill up until the server stop or is there a more global MRU policy or some such?

business/definitionService.js
dabutvin: Ah right. It gets added when it is computed, but this needs to account for it being in the store, but not in cache. Good call

providers/caching/redis.js
dabutvin: The redis itself has a maxmemory policy to evict keys when there is pressure.

dabutvin:
thanks for going through this @jeffmcaffer 

want to take another look before we integrate?

PR #391, opened 2019-02-11
move definition query into GET route

dabutvin:
we want to cache this with cloudflare so moving from POST to GET

### DIFF
--- a/routes/definitions.js
+++ b/routes/definitions.js
@@ -24,24 +24,21 @@ async function getDefinition(request, response) {
 
 // Get a list of autocomplete suggestions of components for which we have any kind of definition.
 // and match the given query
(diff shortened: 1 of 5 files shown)


routes/definitions.js
```diff
@@ -24,24 +24,19 @@ async function getDefinition(request, response) {
 
 // Get a list of autocomplete suggestions of components for which we have any kind of definition.
 // and match the given query
-router.get('', asyncMiddleware(getDefinitionSuggestions))
-async function getDefinitionSuggestions(request, response) {
+router.get('', asyncMiddleware(getDefinitions))
```
jeffmcaffer: Probably need to update the Swagger.

dabutvin: ahh yes. totally forgot

PR #397, opened 2019-02-14
bump up the size limit on webhook body

dabutvin:
we've been dropping a lot of webhooks because of `PayloadTooLargeError` 

Want to see if we can bring the number of dropped hooks down without causing too much trouble

### DIFF
--- a/app.js
+++ b/app.js
@@ -98,7 +98,7 @@ function createApp(config) {
 
   const swaggerOptions = { swaggerUrl: `${config.endpoints.service}/schemas/swagger.yaml` }
   app.use('/api-docs', swaggerUi.serve, swaggerUi.setup(null, swaggerOptions))
-  app.use('/webhook', bodyParser.raw({ limit: '5mb', type: '*/*' }), webhookRoute)
+  app.use('/webhook', bodyParser.raw({ limit: '10mb', type: '*/*' }), webhookRoute)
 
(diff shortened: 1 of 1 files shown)


jeffmcaffer:
Just took a look and we don't do anything useful with the whole body. Just need the coordinates. Suggest that we change the crawler to send a smaller body.

```js
  const body = JSON.parse(request.body)
  const urn = get(body, '_metadata.links.self.href')
  if (!urn) return info(request, response, 400, 'Missing or invalid "self" link')
  const coordinates = EntityCoordinates.fromUrn(urn)
  // TODO validate the coordinates are complete
  await definitionService.computeAndStore(coordinates)
  logger.info(`Handled Crawler update event for ${urn}`)
  response.status(200).end()
```

dabutvin:
good point - i assumed these were github webhooks, but they could very well be our webhooks from the crawler

jeffmcaffer:
Ah, right, could be from GitHub but we ought not have a) that many webhook calls from GitHub and b) they should not be that large. At most they'd include the info for the contribution which should be very small.

jeffmcaffer:
Go ahead and bump the limit but we should also look at the crawler to see what it's doing.

dabutvin:
https://github.com/clearlydefined/crawler/pull/259

PR #406, opened 2019-02-21
update commons-clause exception handling

dabutvin:
we should not be accepting of all exceptions
if a non-spdx exception is used then the license is NOASSERTION

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6168,7 +6168,7 @@
       "integrity": "sha512-2XQACfElKi9SlVb1CYadKDXvoajPgBVPn/gOQLrTvHdElaVhr7ZEbqJaRnJLVNeaI4cMEAgVCeBMKF6MWRDCRA=="
     },
     "spdx-expression-parse": {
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -6168,7 +6168,7 @@
       "integrity": "sha512-2XQACfElKi9SlVb1CYadKDXvoajPgBVPn/gOQLrTvHdElaVhr7ZEbqJaRnJLVNeaI4cMEAgVCeBMKF6MWRDCRA=="
     },
     "spdx-expression-parse": {
-      "version": "github:clearlydefined/spdx-expression-parse.js#c4827de980e293302e225fe924c9edd89f7d1064",
+      "version": "github:clearlydefined/spdx-expression-parse.js#9c9211181d82c9df70dae6e2599ee84fd19e4858",
```
dabutvin: see https://github.com/clearlydefined/spdx-expression-parse.js/commit/9c9211181d82c9df70dae6e2599ee84fd19e4858

PR #421, opened 2019-03-02
More intelligent license merging

### LINKED ISSUE #383: Merging licenses should union
There are cases where one tool finds a license that another does not. Since aggregation is done in a layered "precedence" ordering, it is possible for that extra license to be overwritten by a higher precedence tool. In general we don't really want precedence but have to make a choice. Here we can merge the license lists.

In particular, in utils.js

```js
function _mergeLicense(base, proposed) {
  return !proposed || proposed === 'NOASSERTION' ? base : proposed
}
```

should be enhanced to start with the base and add in the proposed licenses only if they are not already satisfied by the base. There are probably some edge cases and interesting questions but it should be doable/deterministic.

An example of this scenario https://clearlydefined.io/definitions/npm/npmjs/@angular/compiler/6.1.10
The following files have Apache in them but show as only MIT. ScanCode discovers Apache but only with a score of 74 (below the threshold of 80) and so is ignored. Nomos sees both but because ScanCode is higher precedence, that value is ignored.

* bundles/compiler-testing.umd.js
* bundles/compiler-testing.umd.js.map
* bundles/compiler-testing.umd.min.js.map
* bundles/compiler.umd.js
* bundles/compiler.umd.js.map
* bundles/compiler.umd.min.js.map

jeffmcaffer:
Fixes #383 

* Add a set of utils for SPDX expression manipulation (expand, flatten, merge)
* update the license merging to use the new SPDX merge capability
* Add a mess of tests


test/lib/util.js
```diff
@@ -129,7 +149,6 @@ describe('Utils mergeDefinitions', () => {
     }
     utils.mergeDefinitions(base, newDefinition)
     const file = base.files[0]
-    expect(file.license).to.eq('GPL-3.0')
```
dabutvin: let's add this back with `GPL-3.0 AND MIT`

PR #425, opened 2019-03-03
Add support for embedded components

jeffmcaffer:
Some components have embedded components, we want to capture that to enable people better understanding their dependencies (i..e, what they are getting) and identify potential security vulnerabilities etc. 

This PR adds a `described.embedded` property which is an array of entries each of which identifies base information about a discovered embedded component. In particular:
* coordinates: as a string
* location: location in the parent component that the embedded component was found
* other?

Open questions:
- [ ]  should the `location` be the manifest file or the directory? Currently set as the file but in terms of managing/organizing the files in the parent component, using the directory would be better (easier filtering). @pombredanne, how can one reliably find the directory in which the embedded package is located? is that always `summary.packages[N].files[0].path`? 
- [ ] We have to guess at the provider for the embedded component. Right now the code uses the canonical mapping to default provider given a type. That may or may not be true but without some sort of clarity in the origin content we have to guess. Is this a point of curation?
- [ ] Are there other properties we need to track of the embedded components?
- [ ] @pombredanne in the summary info, is there a canonical way to tell which `package` entry represents the package at hand rather than the embedded packages? Is is always the first entry? Do we look at the `path` and decide based on that?
- [ ] @pombredanne are there some good examples of packages of different types that have embedded packages? would be great to have a suite of sample output off which we can code and test.


jeffmcaffer:
Ping @pombredanne

pombredanne:
For your first question and this is not specific to embedded packages

> Should the location be the manifest file or the directory? Currently set as the file but in terms of managing/organizing the files in the parent component, using the directory would be better (easier filtering).

In general it should be the directory, unless the file itself is both a package and a manifest (rare, this is the cases for RPMs for instance) and it depends on each type and even then, there could be variations: an npm package.json manifest means the parent directory is the package. But for a Maven POM, it could be sometimes true or sometimes when a POM is under a META-INF it could be something else. As a result there is specific code in ScanCode for each package type that handles figuring out where is the package root given a a manifest file in a certain context   


>  @pombredanne, how can one reliably find the directory in which the embedded package is located? is that always summary.packages[N].files[0].path?

In a scan `summary.packages` would list all the packages embedded or not and `summary.packages[N].files` list the locations where these are found. Beyond the file system containment there is no notion of embedding for now. So to if you consider embedding as being when there is a file/directory containment, then you should walk the `files` system tree and collect all the packages, taking note of the containment as you go.

> We have to guess at the provider for the embedded component. Right now the code uses the canonical mapping to default provider given a type. That may or may not be true but without some sort of clarity in the origin content we have to guess. Is this a point of curation?

Each one returned by ScanCode has a Package URL ( as a string and as discrete attributes) As we have discussed there is no notion of a provider in a Package URL, but you could map a Package URL type and come with a default. 

> Are there other properties we need to track of the embedded components?

Would faceting is handled separately? Do you care about an embed to also be a referenced direct dependency in a manifest or be listed in a lock file? Would the embed then be its own record otherwise?

> in the summary info, is there a canonical way to tell which package entry represents the package at hand rather than the embedded packages? Is is always the first entry? Do we look at the path and decide based on that?

It would be one of the first entries in the top most directory but there are packages that have multiple personalities (e.g. they have a package.json, a bower.json, a gemspec, etc) in which case which one is the top from a set of peer top level is really usage-context dependent. 


> are there some good examples of packages of different types that have embedded packages? would be great to have a suite of sample output off which we can code and test.

This is somewhat common for Maven in source form with nested POMs and more common in Go (with vendoring). This is seen in some npms that also contain an installed "node_modules".
I will come with a few URLs

pombredanne:
Some examples:
Repos that contains several POMs at once: 
- https://github.com/prestodb/presto
- https://github.com/jeremylong/DependencyCheck

Maven Jars that combine other 5+ other ones:
- http://central.maven.org/maven2/io/fabric8/kubernetes-openshift-uberjar/4.1.3/


More examples later...

jeffmcaffer:
Thanks for the examples. More would be great. Example output would be cool too.

Interestingly, Prestodb appears to be more of a monorepo. Not sure, just looking at it. That feels like a different scenario (also needs additional support) as the top level should not be considered a component, rather the folders in the repo would be the components.

pombredanne:
> Interestingly, Prestodb appears to be more of a monorepo.

That's a common layout for Maven projects and there is usually a hierarchical relationship between each of the packages. You could see this as some sort of a monorepo of course since this is a fairly different layout from the actual built bits.


https://github.com/cloudfoundry/cli/tree/master/vendor is a typical Go project with its vendor directory where there are eventually many embeds.

A similar practice in Ruby https://github.com/cloudfoundry/bosh-linux-stemcell-builder/tree/master/vendor/cache


An example of node embeddeding https://github.com/bazelbuild/rules_nodejs/tree/master/third_party/github.com


And one on Python https://github.com/nexB/scancode-toolkit/tree/develop/thirdparty ;)

pombredanne:
@jeffmcaffer glad you resurfaced this one. On the topic of package manifest vs. package root directory, the latest version of ScanCode (not yet fully released and not yet used in CD) takes an updated and slightly different approach to handle the manifest vs. the root directory of a package. 
A manifest record now holds a pointer to the package root directory if any.  And there is also a new way to know all the files that are part of a given package (including handling the case of nested packages) where a file can optionally have a pointer to its package location (which is the always the package manifest and never some a root directory)

geneh:
@pombredanne Should we wait for the new scancode version to be released?

Notes to self: 
- [ ]  check if the schemaVersion needs to be incremented.
- [ ]  test npm, maven, github and other examples - see above.
- [ ]  re-queue the affected definitions once merged.

tsteenbe:
Wondering back into ClearlyDefined land after being away for a long time ... 

Question for me is do you do the static or dynamic analysis to determine embedded software?  We in OSS Review Toolkit have been dealing with package managers for a while - dynamic will get you the actual resolve packages and versions 100% of the time but can be error prone. Static only work well for certain languages but does not work well for say Gradle and SBT and usually uses regular expressions which can get things like the exact version number wrong. Each method has its pros and cons.

P.S.  @jeffmcaffer Welcome to package manager I knww at some CD would start taking package manifest into account ;-)

nellshamrell:
Hello all!

Thank you so much for this contribution and the discussion on it! Due to the age of this pull request, I'm going to go ahead and close it for now (though the discussion will be preserved and we can refer to it in the future when needed). If this is something you wish to discuss further, please feel free to open a new pull request or a new issue and we can discuss how to move forward.

Thank you again! Community discussion is the backbone of OSS!

PR #426, opened 2019-03-04
Issue #419 - Move URL computation to ClearlyDefined summarizer

### LINKED ISSUE #419: Move URL computation to ClearlyDefined summarizer
in #410 we added various URLs for each definition. Given that that code is necessarily provider-specific, it would be good to isolate it with the other provider-specific code that is in the ClearlyDefined summarizer (https://github.com/clearlydefined/service/blob/master/providers/summary/clearlydefined.js). 

In the eventual future all of the provider-specific code will be factored out into separate modules for each provider do lumping it all together is a good deal. 

Related, it is likely that the PyPi data needed to get or construct the download URL is already available in the harvested data that is made available to the summarizer. So rather than calling pypi to get the URL, we should be able to access it from the existing data.

storrisi:
Fixes #419

### DIFF
--- a/business/definitionService.js
+++ b/business/definitionService.js
@@ -230,9 +230,9 @@ class DefinitionService {
     const summaries = await this.summaryService.summarizeAll(coordinates, raw)
     const aggregatedDefinition = (await this.aggregationService.process(summaries, coordinates)) || {}
     aggregatedDefinition.coordinates = coordinates
-    await this._ensureToolScores(coordinates, aggregatedDefinition)
+    await this._ensureToolScores(coordinates, aggregatedDefinition, raw)
     const definition = await this.curationService.apply(coordinates, curationSpec, aggregatedDefinition)
-    await this._finalizeDefinition(coordinates, definition)
+    await this._finalizeDefinition(coordinates, definition, raw)
     this._ensureCuratedScores(definition)
     // protect against any element of the compute producing an invalid definition
     this._ensureNoNulls(definition)
@@ -261,9 +261,9 @@ class DefinitionService {
 
   // Compute and store the scored for the given definition but do it in a way that does not affect the
   // definition so that further curations can be done.
-  async _ensureToolScores(coordinates, definition) {
+  async _ensureToolScores(coordinates, definition, raw) {
     const rawDefinition = extend(true, {}, definition)
-    await this._finalizeDefinition(coordinates, rawDefinition)
+    await this._finalizeDefinition(coordinates, rawDefinition, raw)
     const { describedScore, licensedScore } = this._computeScores(rawDefinition)
     set(definition, 'described.toolScore', describedScore)
     set(definition, 'licensed.toolScore', licensedScore)
@@ -275,10 +275,10 @@ class DefinitionService {
     set(definition, 'licensed.score', licensedScore)
   }
 
-  async _finalizeDefinition(coordinates, definition) {
+  async _finalizeDefinition(coordinates, definition, raw) {
     this._ensureFacets(definition)
     this._ensureSourceLocation(coordinates, definition)
-    await this._ensureUrls(coordinates, definition)
+    await this._ensureUrls(coordinates, definition, raw)
     set(definition, '_meta.schemaVersion', currentSchema)
     set(definition, '_meta.updated', new Date().toISOString())
   }
@@ -504,10 +504,10 @@ class DefinitionService {
     }
   }
 
-  async _ensureUrls(coordinates, definition) {
+  async _ensureUrls(coordinates, definition, raw) {
     const registryUrl = buildRegistryUrl(coordinates)
     const versionUrl = buildVersionUrl(coordinates)
-    const downloadUrl = await buildDownloadUrl(coordinates)
+    const downloadUrl = await buildDownloadUrl(coordinates, raw)
     this._ensureDescribed(definition)
     definition.described.urls = { registry: registryUrl, version: versionUrl, download: downloadUrl }
   }
--- a/business/definitionService.js
+++ b/business/definitionService.js
@@ -227,9 +227,9 @@ class DefinitionService {
     const summaries = await this.summaryService.summarizeAll(coordinates, raw)
     const aggregatedDefinition = (await this.aggregationService.process(summaries, coordinates)) || {}
     aggregatedDefinition.coordinates = coordinates
-    await this._ensureToolScores(coordinates, aggregatedDefinition, raw)
+    this._ensureToolScores(coordinates, aggregatedDefinition)
     const definition = await this.curationService.apply(coordinates, curationSpec, aggregatedDefinition)
-    await this._finalizeDefinition(coordinates, definition)
+    this._finalizeDefinition(coordinates, definition)
     this._ensureCuratedScores(definition)
     // protect against any element of the compute producing an invalid definition
     this._ensureNoNulls(definition)
@@ -258,9 +258,9 @@ class DefinitionService {
 
   // Compute and store the scored for the given definition but do it in a way that does not affect the
   // definition so that further curations can be done.
-  async _ensureToolScores(coordinates, definition, raw) {
+  _ensureToolScores(coordinates, definition) {
     const rawDefinition = extend(true, {}, definition)
-    await this._finalizeDefinition(coordinates, rawDefinition, raw)
+    this._finalizeDefinition(coordinates, rawDefinition)
     const { describedScore, licensedScore } = this._computeScores(rawDefinition)
     set(definition, 'described.toolScore', describedScore)
     set(definition, 'licensed.toolScore', licensedScore)
@@ -272,7 +272,7 @@ class DefinitionService {
     set(definition, 'licensed.score', licensedScore)
   }
 
-  async _finalizeDefinition(coordinates, definition) {
+  _finalizeDefinition(coordinates, definition) {
     this._ensureFacets(definition)
     this._ensureSourceLocation(coordinates, definition)
     set(definition, '_meta.schemaVersion', currentSchema)
--- a/business/definitionService.js
+++ b/business/definitionService.js
@@ -21,9 +21,6 @@ const {
   setToArray,
   addArrayToSet,
   buildSourceUrl,
-  buildRegistryUrl,
-  buildVersionUrl,
-  buildDownloadUrl,
   isDeclaredLicense,
   simplifyAttributions,
   updateSourceLocation
@@ -232,7 +229,7 @@ class DefinitionService {
     aggregatedDefinition.coordinates = coordinates
     await this._ensureToolScores(coordinates, aggregatedDefinition, raw)
     const definition = await this.curationService.apply(coordinates, curationSpec, aggregatedDefinition)
-    await this._finalizeDefinition(coordinates, definition, raw)
+    await this._finalizeDefinition(coordinates, definition)
     this._ensureCuratedScores(definition)
     // protect against any element of the compute producing an invalid definition
     this._ensureNoNulls(definition)
@@ -275,10 +272,9 @@ class DefinitionService {
     set(definition, 'licensed.score', licensedScore)
   }
 
-  async _finalizeDefinition(coordinates, definition, raw) {
+  async _finalizeDefinition(coordinates, definition) {
     this._ensureFacets(definition)
     this._ensureSourceLocation(coordinates, definition)
-    await this._ensureUrls(coordinates, definition, raw)
     set(definition, '_meta.schemaVersion', currentSchema)
     set(definition, '_meta.updated', new Date().toISOString())
   }
@@ -502,14 +498,6 @@ class DefinitionService {
       default:
         return
     }
-  }
-
-  async _ensureUrls(coordinates, definition, raw) {
-    const registryUrl = buildRegistryUrl(coordinates)
-    const versionUrl = buildVersionUrl(coordinates)
-    const downloadUrl = await buildDownloadUrl(coordinates, raw)
-    this._ensureDescribed(definition)
-    definition.described.urls = { registry: registryUrl, version: versionUrl, download: downloadUrl }
   }
 
   _getDefinitionCoordinates(coordinates) {
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -6,7 +6,7 @@ const he = require('he')
 const EntityCoordinates = require('./entityCoordinates')
 const ResultCoordinates = require('./resultCoordinates')
 const moment = require('moment')
-const { set, unset, union, sortBy, trim, uniqBy, get, forEach } = require('lodash')
+const { set, unset, union, sortBy, trim, uniqBy } = require('lodash')
 const extend = require('extend')
 const SPDX = require('./spdx')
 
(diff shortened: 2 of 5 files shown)


business/definitionService.js
```diff
@@ -504,10 +504,10 @@ class DefinitionService {
     }
   }
 
-  async _ensureUrls(coordinates, definition) {
```
jeffmcaffer: push this level of function down to clearlydefined.js or the appropriate summarizer. Should not need to access the raw results after the `summarizeAll` call.

storrisi:
@jeffmcaffer i've updated the PR moving the URL building methods into the summarizer.
I've also added URL supporting for both `pod` and `sourcearchive`, which fixes #426 

For pod i didn't found a method to retrieve a specific download url in the harvested data, looks like everything is based on github. Is that right?

providers/summary/clearlydefined.js
```diff
@@ -132,20 +135,35 @@ class ClearlyDescribedSummarizer {
     let licenses = licenseUrls.map(extractLicenseFromLicenseUrl).filter(x => x)
     if (!licenses.length) licenses = licenseNames.map(x => SPDX.lookupByName(x) || x).filter(x => x)
     if (licenses.length) setIfValue(result, 'licensed.declared', SPDX.normalize(licenses.join(' OR ')))
+    const registryUrl = buildRegistryUrl(coordinates)
```
jeffmcaffer: This code is replicated many times. Break it out into a function that is shared. Something like `addStandardUrls` perhaps

providers/summary/clearlydefined.js
```diff
@@ -132,20 +135,35 @@ class ClearlyDescribedSummarizer {
...
+    const versionUrl = buildVersionUrl(coordinates)
+    const downloadUrl = buildDownloadUrl(coordinates)
+    if (get(result, 'described'))
+      setIfValue(result, 'described.urls', { registry: registryUrl, version: versionUrl, download: downloadUrl })
```
jeffmcaffer: If the consts happened to have matching names (e.g., `registry` rather than `registryUrl`) then this could just be `{ registry, version, download }`

providers/summary/clearlydefined.js
```diff
@@ -132,20 +135,35 @@ class ClearlyDescribedSummarizer {
     let licenses = licenseUrls.map(extractLicenseFromLicenseUrl).filter(x => x)
     if (!licenses.length) licenses = licenseNames.map(x => SPDX.lookupByName(x) || x).filter(x => x)
     if (licenses.length) setIfValue(result, 'licensed.declared', SPDX.normalize(licenses.join(' OR ')))
+    const registryUrl = buildRegistryUrl(coordinates)
+    const versionUrl = buildVersionUrl(coordinates)
+    const downloadUrl = buildDownloadUrl(coordinates)
+    if (get(result, 'described'))
```
jeffmcaffer: The if is not needed. We want to set these values regardless of whether there is a `described` prop. Setting will create the `described` prop

providers/summary/clearlydefined.js
```diff
@@ -179,30 +202,58 @@ class ClearlyDescribedSummarizer {
...
   }
 
-  addPyPiData(result, data) {
+  addPyPiData(result, data, coordinates) {
     setIfValue(result, 'described.releaseDate', extractDate(data.releaseDate))
     setIfValue(result, 'licensed.declared', data.declaredLicense)
+    const releases = get(data, 'registryData.releases')
+    let downloadUrl
+    forEach(releases[coordinates.revision], revision => {
+      if (revision.filename.includes('tar.gz')) downloadUrl = revision.url
```
jeffmcaffer: Do you want the first or last one?

business/definitionService.js
```diff
@@ -261,9 +258,9 @@ class DefinitionService {
 
   // Compute and store the scored for the given definition but do it in a way that does not affect the
   // definition so that further curations can be done.
-  async _ensureToolScores(coordinates, definition) {
+  async _ensureToolScores(coordinates, definition, raw) {
```
jeffmcaffer: this whole call chain no longer needs async

storrisi: Actually this method could be removed at all. I will fix it. Thanks!

providers/summary/clearlydefined.js
storrisi: I've changed this after updating tests. 
Looks like that in some cases the described prop should be undefined at all, like `expect(summary.described).to.be.undefined`

providers/summary/clearlydefined.js
storrisi: Right, we can save some code doing this

business/definitionService.js
jeffmcaffer: not sure what you mean about removing methods. All the methods here were here before and need to stay. Just the async/awaits no longer need to be added.

providers/summary/clearlydefined.js
jeffmcaffer: We were likely expecting undefined in the tests because previously there was nothing being added to `described` so there should not have been one. Now we have changed the system to add in URLs so the expectation should change.

providers/summary/clearlydefined.js
jeffmcaffer: mostly just cleaner. The real code saving comes from removing all the duplicate code.

providers/summary/clearlydefined.js
storrisi: That's a good question.
Looks like that Pypi could have multiple objects for the same release.

For example:

```
"0.7": [
            {
              "comment_text": "",
              "digests": {
                "md5": "235729aa2bf99e1b940eba53fcbfb37c",
                "sha256": "bd053a6d0e5231bec41da6ef524369d64dc9a7f27c5075914ae1b1abc4dd33e5"
              },
              "downloads": -1,
              "filename": "idna-0.7.tar.gz",
              "has_sig": false,
              "md5_digest": "235729aa2bf99e1b940eba53fcbfb37c",
              "packagetype": "sdist",
              "python_version": "source",
              "size": 20294,
              "upload_time": "2014-07-10T03:37:29",
              "url": "https://files.pythonhosted.org/packages/cf/57/53ac056e1a9d028dfce896929d0fc769e53a1e4d0917c2ec163acae4528f/idna-0.7.tar.gz"
            }
          ],
```

has only one object that contains the .tar.gz file URL, in other cases i found multiple objects for the same release, with different file URL and extension.

What do you think could be the best way in this case?

jeffmcaffer: No idea unfortunately of how to distinguish between multiple entries for the same component version. Propose the following
* add a TODO comment here noting that we are randomly picking the last (or first) entry that has a URL.
* Open an issue to figure out what the right answer is and in that issue, put some examples of components that have multiple entries for the same version with different URLs

From there we can get some pypi experts to weigh in on the situation and give us advice.

storrisi: Thanks, i will follow your suggestions

business/definitionService.js
storrisi: Ah, sorry, i thought you were referring to the `ensureUrls` method that actually is not used anymore. My bad.

storrisi:
@jeffmcaffer i've updated the PR based on your suggestions, as well as tests.

lib/utils.js
```diff
@@ -239,12 +247,14 @@ function buildVersionUrl(spec) {
...
       return `${buildRegistryUrl(spec)}/versions/${spec.revision}`
+    case 'cocoapods':
+      return `https://cocoapods.org/pods/${spec.name}`
     default:
       return null
   }
 }
 
-async function buildDownloadUrl(spec) {
+function buildDownloadUrl(spec) {
```
jeffmcaffer: might be good to add a comment here that this will build URLs for any protocol with URLs that can be predicted from the identity of the component

test/summary/clearlydefinedTests.js
```diff
@@ -140,17 +146,55 @@ describe('ClearlyDescribedSummarizer addNpmData', () => {
...
     // should work in the type field as well
     for (let license of Object.keys(data)) {
       let result = {}
-      summarizer.addNpmData(result, { registryData: { manifest: { license: { type: license } } } })
-      if (data[license]) assert.deepEqual(result, { licensed: { declared: data[license] } })
-      else assert.deepEqual(result, {})
+      summarizer.addNpmData(result, { registryData: { manifest: { license: { type: license } } } }, npmTestCoordinates)
+      if (data[license])
+        assert.deepEqual(result, {
+          described: {
```
jeffmcaffer: this duplicate code can be collapsed by conditionally creating the target opbject and then asserting. So 

```js
const expected = {
  described: {
    urls: {
      download: 'https://registry.npmjs.com/test/-/test-1.0.tgz',
      registry: 'https://npmjs.com/package/test',
      version: 'https://npmjs.com/package/test/v/1.0'
  }
}
if (data[license]) expected.license = { declared: data[license] }
assert.deepEqual(result, expected)
```

Since the `urls` part is reused so much thoughout the tests, perhaps that should be a constant somewhere as well.

test/summary/clearlydefinedTests.js
```diff
@@ -166,51 +210,135 @@ describe('ClearlyDescribedSummarizer addNpmData', () => {
     }
     for (let date of Object.keys(data)) {
       let result = {}
-      summarizer.addNpmData(result, { registryData: { releaseDate: date } })
-      assert.deepEqual(result, { described: { releaseDate: data[date] } })
+      summarizer.addNpmData(result, { registryData: { releaseDate: date } }, npmTestCoordinates)
+      assert.deepEqual(result, {
+        described: {
+          releaseDate: data[date],
+          urls: {
```
jeffmcaffer: make a reusable `urls` prop so we don't have to duplicate this everywhere

test/summary/clearlydefinedTests.js
```diff
@@ -226,9 +354,28 @@ describe('ClearlyDescribedSummarizer addNuGetData', () => {
 
     for (let licenseExpression of Object.keys(data)) {
       let result = {}
-      summarizer.addNuGetData(result, { manifest: { licenseExpression } })
-      if (data[licenseExpression]) assert.deepEqual(result, { licensed: { declared: data[licenseExpression] } })
-      else assert.deepEqual(result, {})
+      summarizer.addNuGetData(result, { manifest: { licenseExpression } }, nugetTestCoordinates)
+      if (data[licenseExpression])
+        assert.deepEqual(result, {
```
jeffmcaffer: see above comments on duplicate code elimination

storrisi:
Thanks @jeffmcaffer , i've updated the PR and you can review it again.

providers/summary/clearlydefined.js
```diff
@@ -181,28 +189,44 @@ class ClearlyDescribedSummarizer {
...
     if (license) {
       setIfValue(result, 'licensed.declared', SPDX.normalize(typeof license === 'string' ? license : license.type))
     }
+
+    const homepage = get(data, 'registryData.homepage')
+    const revision = get(data, 'registryData.sourceInfo.revision')
+
+    const registry = buildRegistryUrl(coordinates)
+    const version = homepage && revision ? `${homepage}/tree/${revision}` : null
+    const download = homepage && revision ? `${homepage}/archive/${revision}.zip` : null
```
dabutvin: This only works when the homepage is on github

This one for example would not work, https://api.clearlydefined.io/definitions/pod/cocoapods/-/crashlytics/3.12.0

dabutvin: take a look at https://github.com/clearlydefined/crawler/blob/master/providers/fetch/podFetch.js#L75 for how we resolve it.

providers/summary/clearlydefined.js
```diff
@@ -181,28 +189,44 @@ class ClearlyDescribedSummarizer {
...
+  addPyPiData(result, data, coordinates) {
     setIfValue(result, 'described.releaseDate', extractDate(data.releaseDate))
     setIfValue(result, 'licensed.declared', data.declaredLicense)
+    const releases = get(data, 'registryData.releases')
+    // TODO: we are currently picking the first url that contains a tar.gz extension
+    // we should understand what's the correct process on a pypi definition that contains multiple object for the same release
+    const revision = find(releases[coordinates.revision], revision => revision.filename.includes('tar.gz'))
+    const registry = buildRegistryUrl(coordinates)
+    const version = buildVersionUrl(coordinates)
+    setIfValue(result, 'described.urls', { registry, version, download: revision.url })
```
dabutvin: setIfValue doesn't do much here since we are providing an object for the value inline. It will never be falsy and not set.

You may want to setIfValue each individual property

providers/summary/clearlydefined.js
```diff
@@ -11,7 +11,10 @@ const {
   isDeclaredLicense,
   isLicenseFile,
   updateSourceLocation,
-  mergeDefinitions
+  mergeDefinitions,
+  buildUrls,
+  buildRegistryUrl,
+  buildVersionUrl
 } = require('../../lib/utils')
```
dabutvin: I think we should probably move `buildUrls` out of utils now that it won't handle all coordinates thrown at it.
Probably better just to put each in it's respective spot here.

so for example `addNpmData` will add the urls for npms

storrisi: I'm not 100% sure about this.
Most of the providers use the same path to retrieve the URLs.
Only in some specific cases (Pypi, Cocoapods) we must use different ways.

Removing that function would potentially cause some code duplication.

Thoughts?

providers/summary/clearlydefined.js
storrisi: Right, thanks

providers/summary/clearlydefined.js
storrisi: Ah, thanks, i hadn't any example for that case.
Looking at the harvested data of the example, looks like that we can't link to a specific version URL, since it doesn't exists.

providers/summary/clearlydefined.js
dabutvin: I don't think code duplication will be a problem here

It would be slotting these into each of the methods instead of living in a big partial switch

```js
function buildRegistryUrl(spec) {
  switch (spec.provider) {
    case 'github':
      return `https://github.com/${spec.namespace}/${spec.name}`
    case 'npmjs':
      return `https://npmjs.com/package/${spec.namespace ? spec.namespace + '/' + spec.name : spec.name}`
    case 'nuget':
      return `https://nuget.org/packages/${spec.name}`
    case 'cratesio':
      return `https://crates.io/crates/${spec.name}`
    case 'mavencentral':
      return `https://mvnrepository.com/artifact/${spec.namespace}/${spec.name}`
    case 'pypi':
      return `https://pypi.org/project/${spec.name}`
    case 'rubygems':
      return `https://rubygems.org/gems/${spec.name}`
    case 'cocoapods':
      return `https://cocoapods.org/pods/${spec.name}`
    default:
      return null
  }
```

providers/summary/clearlydefined.js
dabutvin: that's what I found too :( that's okay

providers/summary/clearlydefined.js
storrisi: Right, will address this soon.

storrisi: Maybe i'm wrong but i can't found the specific method for `github` in the summarizer. Is that right?

dabutvin: you're right. We'll have to add one - good catch

storrisi:
@dabutvin i've refactored the methods following your suggestions.
I've also added the support for `git` type.

dabutvin:
👍

PR #434, opened 2019-03-06
add lock around writing to mongo

dabutvin:
- concurrent writes are throwing primary key violations
- we are deleting all the document pages and then inserting them back
- doing this in parallel requests (like the webhook setup) cause failures

try adding ?w=1 to see this better


dabutvin:
this could be improved greatly by moving to a more atomic write setup.
Instead of deleting and then inserting with the lock, we could always insert somewhere new and then point to the new collection

jeffmcaffer:
decided on a different approach

PR #435, opened 2019-03-07
filelist pointer for upserting documents

dabutvin:
this will grow the database indefinitely if there is no cleanup in place

### DIFF
--- a/providers/stores/mongo.js
+++ b/providers/stores/mongo.js
@@ -7,6 +7,7 @@ const EntityCoordinates = require('../../lib/entityCoordinates')
 const logger = require('../logging/logger')
 const { clone, get, range } = require('lodash')
 const base64 = require('base-64')
+const uuid = require('uuid').v4
 
 const sortOptions = {
   type: 'coordinates.type',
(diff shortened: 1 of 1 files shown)


providers/stores/mongo.js
```diff
@@ -126,24 +128,43 @@ class MongoStore {
...
+    const result = await this.collection.bulkWrite(
       range(pages).map(
         index => {
           if (index === 0) {
             const definitionPage = clone(definition)
             if (definition.files) definitionPage.files = definition.files.slice(0, pageSize)
-            return { ...definitionPage, _mongo: { partitionKey: definition._id, page: 1, totalPages: pages } }
+            return {
+              updateOne: {
+                filter: { _id: definition._id },
```
dabutvin: We should update this filter to include

```
'described.tools.length' : { $lt: definition.described.tools.length }
```

This will preserve the order, so we won't write unless the number of tools are increasing.

jeffmcaffer: not entirely convinced as the set can be different but the same length. I suppose that would be a lost up date situation but there is no clear winner.

dabutvin: good point. this just came up as a mechanism that I wanted to capture while we figure out exactly what to do here

jeffmcaffer:
decided on a different approach

PR #439, opened 2019-03-08
add top level scores definition element

### LINKED ISSUE #436: Improve find definitions with effective and tools score
As mentioned in https://github.com/clearlydefined/website/issues/495 , we want to be able to search definitions that have a specific effective and tools score value.

To do that, the mongoDb query could be possibly be improved in order to search on the average value between licensed.score and described.score (effective score) and licensed.toolScore and described.toolScore (tools score).

@dabutvin @jeffmcaffer @pombredanne thoughts?

jeffmcaffer:
Fixes #436 

Here we add a new top level definition element for `scores`. It captures the average score across all of the dimensions (currently only licensed and described) for both curated and tools scores. The curated score is generally the score that the UI renders for the user.

### DIFF
--- a/providers/stores/mongo.js
+++ b/providers/stores/mongo.js
@@ -16,7 +16,9 @@ const sortOptions = {
   license: 'licensed.declared',
   releaseDate: 'described.releaseDate',
   licensedScore: 'licensed.score.total',
-  describedScore: 'described.score.total'
+  describedScore: 'described.score.total',
+  effectiveScore: 'scores.effective',
+  toolScore: 'scores.tool'
 }
 
 class MongoStore {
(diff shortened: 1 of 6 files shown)


jeffmcaffer:
Need also to add scores.curated and scores.tool to the definitions query capability. Separate PR?

providers/stores/mongo.js
```diff
@@ -177,6 +179,10 @@ class MongoStore {
     if (parameters.license) filter['licensed.declared'] = parameters.license
     if (parameters.releasedAfter) filter['described.releaseDate'] = { $gt: parameters.releasedAfter }
     if (parameters.releasedBefore) filter['described.releaseDate'] = { $lt: parameters.releasedBefore }
+    if (parameters.minEffectiveScore) filter['scores.effective'] = { $gt: parseInt(parameters.minEffectiveScore) }
+    if (parameters.maxEffectiveScore) filter['scores.effective'] = { $lt: parseInt(parameters.maxEffectiveScore) }
+    if (parameters.minToolScore) filter['scores.tool'] = { $gt: parseInt(parameters.minToolScore) }
+    if (parameters.maxToolScore) filter['scores.tool'] = { $lt: parseInt(parameters.maxToolScore) }
```
storrisi: With these new params in place we can now query the definitions for effective and tool score, which would cover https://github.com/clearlydefined/website/issues/501 requirement.
Should we need to add another combined field to query? Like `scores.total` perhaps , which is, actually, the real score of the entire definition, as required in https://github.com/clearlydefined/website/issues/495

dabutvin: the real score of the entire definition will be `definition.scores.effective`

dabutvin:
👍

PR #446, opened 2019-03-11
Added support for Pypi download package extension

### LINKED ISSUE #445: add support for python zip files for download urls
see https://pypi.org/pypi/azure-mgmt-rdbms/json as an example

it is not a `.gz` it is a `.zip` so it doesn't get picked up

see https://github.com/clearlydefined/service/blob/master/providers/summary/clearlydefined.js#L286-L297

storrisi:
Fixes #445


dabutvin:
👍

PR #451, opened 2019-03-21
check the store for existing curations

dabutvin:
we were checking github directly with HEAD requests
this was happening on the order of ~50K per day
our new curation requests started getting throttled
This increases load on the DB, but should reduce our rate in GitHub


dabutvin:
replaced with #452

PR #452, opened 2019-03-22
use smartgit to get curations out of github

dabutvin:
we were checking github directly with HEAD requests
this was happening on the order of ~50K per day
our new curation requests started getting throttled

### DIFF
--- a/providers/curation/github.js
+++ b/providers/curation/github.js
@@ -288,8 +288,8 @@ ${this._formatDefinitions(patch.patches)}`
   async _getCurations(coordinates, pr = null) {
     const path = this._getCurationPath(coordinates)
     const { owner, repo } = this.options
-    const smartGit = geit(`https://github.com/${owner}/${repo}.git`)
-    const tree = await smartGit.tree(pr ? `refs/pull/${pr}/head` : this.options.branch)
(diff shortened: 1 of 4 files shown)


providers/curation/github.js
```diff
@@ -286,42 +286,20 @@ ${this._formatDefinitions(patch.patches)}`
...
-      return null
-    }
-  }
-
-  async _getBranchAndSha(number) {
-    if (!number) return { ref: this.options.branch }
-    const { owner, repo } = this.options
-    const result = await this.github.pullRequests.get({ owner, repo, number })
-    return { ref: result.data.head.ref, sha: result.data.head.sha }
+    const smartGit = geit(`https://github.com/${owner}/${repo}.git`)
```
iamwillbar: I would put an `encodeURIComponent` around `owner` and `repo` to prevent URL injection.

providers/curation/github.js
```diff
@@ -286,42 +286,20 @@ ${this._formatDefinitions(patch.patches)}`
...
+    const tree = await smartGit.tree(pr ? `refs/pull/${pr}/head` : this.options.branch)
```
iamwillbar: Similarly, encode `pr` and `branch` (unless `tree` does this internally.

providers/curation/github.js
dabutvin: ahh yes

PR #466, opened 2019-03-29
add definition store dispatcher

dabutvin:
- removed average that was DOSing the DB
 - add dispatch to read/write to mongo + blob storage

Plan is to use mongo as the primary and store to both blob and mongo moving forward.
Once we catch up and sync the stores, blob can become primary for listing and point lookups while mongo can be primary for queries.

Blob data will be helpful for mirroring story as well as potential search indexing.


dabutvin:
bitmoji

PR #488, opened 2019-04-10
[suggestions] sort by release date on the client

dabutvin:
if this property is not defined then the mongoDb omits from results

We were seeing missing data because of this

### DIFF
--- a/business/suggestionService.js
+++ b/business/suggestionService.js
@@ -3,6 +3,7 @@
 
 const { get } = require('lodash')
 const { isDeclaredLicense, setIfValue } = require('../lib/utils')
+const moment = require('moment')
 
 class SuggestionService {
   constructor(definitionService) {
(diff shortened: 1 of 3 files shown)


jeffmcaffer:
How can this work? you can sort the values on the page but if you want to do a global sort, the client would need to get all the definitions and then sort. For most components there will only be one page of versions but there are a non-trivial number that have lots.

This actually raises an interesting question. In computing suggestions, the idea was to get the versions "near" the target version. Not sure the current implementation is doing a good job of that. Hmm...

dabutvin:
This is just for suggestions endpoint, not sure what you mean by global sort.
This is what is happening here to find related definitions to pull suggestions from

1. given coordinates
2. chop off the revision
3. find all definitions with the same name/namespace/type/provider
4. sort these definitions by releaseDate
5. find where the given coordinates are in this set
6. use the -3/+3 definitions as a window for related definitions

if we sort by `releaseDate` on the mongo side, it filters any documents without that property and we get no results in a lot of cases.

Does this make sense?

Your comment makes me wonder if we would be better off overall by doing a sort of version number in some cases as opposed to releaseDate, but that could get complicated for non-semver style revisions

PR #492, opened 2019-04-11
ensure duplicate slashes are collapsed

dabutvin:
getting inputs like maven/mavencentral/com.sun.faces/jsf-impl//2.1.0-b03

we can correct these here without causing too much fuss


dabutvin:
bitmoji

PR #503, opened 2019-04-30
Purge cloudflare edge cache from the service

### LINKED ISSUE #418: purge cloudflare edge cache from the service
when a definition or curation are invalidated we clear the redis cache and recompute as necessary.
We should also call out to clear the edge cache so the updates are visible right away.

see https://api.cloudflare.com/#zone-purge-files-by-url

We should be able to purge by url and get the benefits of the edge caching and have instant updates

storrisi:
Fixes #418


providers/caching/cdn.js
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+
+const requestPromise = require('request-promise-native')
+const waterMark = 30
```
AlexWebYourmind: Quick one: is the watermark and defaultIntervalMs a conf that we shall move into the fullenv.json file? or is there a particular reason to keep it hardcoded in cdn.js?

test/routes/definitions.js
```diff
@@ -0,0 +1,47 @@
...
+function createRoutes(definitionService) {
+  return definitionRoutes(definitionService, true)
+}
+
+function createRequest() {
+  return httpMocks.createRequest({
+    method: 'GET',
+    url: '/definitions',
+    params: {
+      name: 'FastMicroservices.Marconi.Consul',
```
AlexWebYourmind: While I do appreciate the Irony here, can we use a random name in the params?

providers/caching/cdn.js
ignacionr: Will do, treating it just as it happens with RATE_LIMIT_MAX (parsed as integer, no provisions for not-a-number).

test/routes/definitions.js
ignacionr: Actually, changing the test example to the result of Microsoft-AdaptiveCards (which returns a few item coordinates) helped me find the need to support multiple coords. Thanks! -please see both feedback items incorporated now.

bin/config.js
```diff
@@ -64,6 +64,9 @@ module.exports = {
   caching: {
     service: loadFactory(config.get('CACHING_PROVIDER') || 'memory', 'caching')
   },
+  cdn: {
+    service: loadFactory('cf', 'cdn')
```
dabutvin: want to spell out `cloudflare` here instead of using `cf`?

business/definitionService.js
```diff
@@ -38,15 +38,27 @@ const currentSchema = '1.6.1'
...
 
+function stringHash(src) {
+  var hash = 0, i, chr
+  if (src.length === 0) return hash
+  for (i = 0; i < src.length; i++) {
+    chr = src.charCodeAt(i)
+    hash = ((hash << 5) - hash) + chr
+    hash |= 0
+  }
+  return hash
```
dabutvin: Is it possible to use node crypto library instead of custom hash function?
https://nodejs.org/docs/latest-v10.x/api/crypto.html#crypto_crypto_createhash_algorithm_options

business/definitionService.js
```diff
@@ -237,6 +250,25 @@ class DefinitionService {
...
+    if (Array.isArray(coordinates)) {
+      return coordinates.map(item => this.tagFromCoordinates(item.coordinates))
+        .filter((v, i, a) => v && a.indexOf(v) === i)
+        .join(',')
+    }
+    else if (coordinates) {
+      let hashBase = [coordinates.type, coordinates.name, coordinates.revision].join('|')
+      result = stringHash(hashBase).toString()
+    }
+    return result
```
dabutvin: I think we should move the tag concept into the `Cdn` class and just pass coordinates around.

the tag is very cloudflare specific in this case

above would be:
`this.cdn.invalidate(coordinates)`

What do you think?

business/definitionService.js
ignacionr: Hi! It is possible, but unnecessarily expensive (we do not care about the same kind of hash algo properties that you look for when it’s security related; this is only being use to reduce the payload).

bin/config.js
ignacionr: I just didn’t want anyone to find out product searching for the for-profit company, when it doesn’t participate in it afaik.

business/definitionService.js
ignacionr: You’re right that it is specific to content delivery, but not particularly to Cloudflare as you can see here https://developer.akamai.com/blog/2019/03/25/introducing-ability-purge-cache-tag

routes/definitions.js
```diff
@@ -37,6 +47,9 @@ async function getDefinitions(request, response) {
   if (pattern) return response.send(await definitionService.suggestCoordinates(pattern))
   if (!validator.validate('definitions-find', request.query)) return response.status(400).send(validator.errorsText())
   const result = await definitionService.find(request.query)
+  if (Array.isArray(result.data)) {
+    response.set('cache-tag', result.data.map(project_tag).join())
```
dabutvin: Is this set up so that when any of the definitions that come back as a result of the `find()` get invalidated the cache of the query is purged?

business/definitionService.js
dabutvin: interesting - thanks for the link 👍

routes/definitions.js
ignacionr: Correct. You get a number of items with coordinates from find; the result is tagged with cache tags, one per item returned. Then when any of those definitions is changed, the CDN is asked to invalidate any and all results containing said tag. I believe it’s quite good

dabutvin: 👍 awesome

business/definitionService.js
ignacionr: For further illustration the article at https://searchsqlserver.techtarget.com/definition/hashing ends with the following assertion.
> A hash function that works well for database storage and retrieval, however, might not work as for cryptographic or error-checking purposes.

ignacionr:
This one seems ready now. Kindly take a look to see if we're good to merge. Otherwise, new opportunities are welcome too.

Thanks.

routes/definitions.js
```diff
@@ -17,12 +17,22 @@ async function getDefinition(request, response) {
   const coordinates = utils.toEntityCoordinatesFromRequest(request)
   const pr = request.params.pr
   // if running on localhost, allow a force arg for testing without webhooks to invalidate the caches
-  const force = request.hostname.includes('localhost') ? request.query.force || false : false
+  const force = (request.hostname && request.hostname.includes('localhost')) ? request.query.force || false : false
   const expand = request.query.expand === '-files' ? '-files' : null // only support '-files' for now
   const result = await definitionService.get(coordinates, pr, force, expand)
+  if (Array.isArray(result.data) && result.data.length > 0) {
+    response.set('cache-tag', definitionService.tagFromCoordinates(result.data))
+  }
```
dabutvin: in this case result is a definition object - there is no Array or data property here, it is always a single one

```
result == {
 described: { ... },
 licensed: { ... },
 ...
}
```

routes/definitions.js
dabutvin: do we need to filter out the null's that come back from project_tag?

ignacionr: feedback incorporated at https://github.com/clearlydefined/service/pull/503/commits/ae49039da28a6fcaaa8e2c39f99df4c1a35ba41e

routes/definitions.js
ignacionr: Thanks for catching this. Incorporated at https://github.com/clearlydefined/service/pull/503/commits/796203435c418209592871e49ea0f592b65aeaed

routes/definitions.js
```diff
@@ -17,12 +17,20 @@ async function getDefinition(request, response) {
   const coordinates = utils.toEntityCoordinatesFromRequest(request)
   const pr = request.params.pr
   // if running on localhost, allow a force arg for testing without webhooks to invalidate the caches
-  const force = request.hostname.includes('localhost') ? request.query.force || false : false
+  const force = (request.hostname && request.hostname.includes('localhost')) ? request.query.force || false : false
   const expand = request.query.expand === '-files' ? '-files' : null // only support '-files' for now
   const result = await definitionService.get(coordinates, pr, force, expand)
+  response.set('cache-tag', definitionService.tagFromCoordinates(result))
```
dabutvin: in order to build the hash correctly this should be 

```
response.set('cache-tag', definitionService.tagFromCoordinates(result.coordinates))
```

ignacionr: I understand, you’re right; otherwise any tool additions will make it a different hash. Thanks for catching this!

iamwillbar:
@geneh can you take a look and see if this is ready to merge?

geneh:
I am closing this pull request. Please check the corresponding https://github.com/clearlydefined/service/issues/418.

PR #507, opened 2019-05-03
Consider 'discovered' as suggestions for licenses

### LINKED ISSUE #453: [Suggestions] Include 'discovered' licenses for declared license suggestions
This would only show up if the declared license was missing or 'NOASSERTION' and there was a discovered license available within this definition.

A simple case for example:
package.json contains `"license": "BSD"` which is not SPDX parseable, but the package contains a LICENSE file that is detected as the `BSD-3-Clause` license, we can suggest `BSD-3-Clause` as the declared license

storrisi:
Fixes #453


ignacionr:
@dabutvin if you'd like to take a look before moving it to 'ready for review', you're very welcome to do so. Thanks in advance.

business/suggestionService.js
```diff
@@ -19,12 +19,30 @@ class SuggestionService {
...
+          .filter(marked => marked.cutAt > 0)
+          .map(marked => { return { value: marked.exp.substr(0, marked.cutAt), version: marked.exp.substr(marked.cutAt + 1) } })
+        let suggestedSoFar = get(suggestion, 'licensed.declared') || []
+        hasSuggested |= setIfValue(suggestion, 'licensed.declared', appendDeclared
+          .concat(suggestedSoFar)
+          .filter((v, i, a) => a.indexOf(v) === i))
+      }
+    }
+
+    return hasSuggested ? suggestion : null
```
dabutvin: this has the affect of returning Not Found instead of the shell of a suggestion

```
{
  "coordinates": {
    "type": "nuget",
    "provider": "nuget",
    "name": "name",
    "revision": "12.0.1"
  }
}
```

I think this is a good change, do you know if the website code is setup to handle this?

business/suggestionService.js
```diff
@@ -19,12 +19,30 @@ class SuggestionService {
...
+      hasSuggested |= this._collectSuggestionsForField(definitions, suggestion, 'described.sourceLocation')
+    }
+
+    if (!isDeclaredLicense(get(definitions.original, 'licensed.declared'))) {
+      let discoveredExpressions = get(definitions.original, 'licensed.facets.core.discovered.expressions')
+      if (Array.isArray(discoveredExpressions)) {
+        let appendDeclared = discoveredExpressions
+          .map(exp => { return { exp, cutAt: exp.indexOf('-') } })
+          .filter(marked => marked.cutAt > 0)
+          .map(marked => { return { value: marked.exp.substr(0, marked.cutAt), version: marked.exp.substr(marked.cutAt + 1) } })
```
dabutvin: the 'version' here is not meant to be the license version.

it is meant to signify where the data was sourced from.
We should probably put  the original`coordinates.revision` in this property and leave the license expression in-tact

does that make sense?

test/business/suggestionServiceTest.js
```diff
@@ -80,6 +78,18 @@ describe('Suggestion Service', () => {
...
+
+  it('will include \'discovered\' licenses for declared license suggestions', async () => {
+    const t2 = EntityCoordinates.fromString('gem/rubygems/-/autobuild/1.6.2.b8')
+    const sample_definition = require('./evidence/issue-453-sample-1.json')
+    const service = setup(sample_definition, [])
+    const suggestions = await service.get(t2)
+    expect(suggestions).to.not.be.null
+    const declared = get(suggestions, 'licensed.declared')
+    expect(declared).to.equalInAnyOrder([
+      { value: 'GPL', version: '2.0' }
```
dabutvin: see comment above about license version.

this should be `{ value: 'GPL-2.0', version: '1.6.2.b8' }`

business/suggestionService.js
ignacionr: Oh, that was my bad. Yes I understand. Let me take care of it.

test/business/suggestionServiceTest.js
ignacionr: Yes, got it.

business/suggestionService.js
ignacionr: Took it from a test, because in other cases there was a null being returned already (I think it's better if we keep consistent). So I assume the website will handle this; it won't hurt to check with Simone, which I will do, though.

dabutvin: 👍 makes sense. thanks

test/business/suggestionServiceTest.js
ignacionr: Incorporated at https://github.com/clearlydefined/service/pull/507/commits/84a2dfb19418482f183eeff276d4aa3478004600

business/suggestionService.js
ignacionr: Incorporated at https://github.com/clearlydefined/service/pull/507/commits/debdca3170bae574c80682f769097da9c47dfe4b

dabutvin:
👍

PR #508, opened 2019-05-06
commit curation as the logged in user (fixes #252)

### LINKED ISSUE #252: validate the id of the committer in curation PRs
ideally the committer tracked in PRs would be the user who did the contribution rather than the clearlydefinedbot. We attempt to use the user's info but fall back if we can't find it. However, I was logged in and the service should have been able to identify me and my email etc yet my contribution commits appeared to come from clearlydefinedbot

dabutvin:
example: https://github.com/clearlydefined/curated-data-dev/pull/881


jeffmcaffer:
Lovin' it!

Issue #312, opened 2018-12-04
the definition query is wrongly falling back to beta release


dabutvin:
When I query for the 12.0.1 release it gives me 12.0.1-beta

https://api.clearlydefined.io/definitions/nuget/nuget/-/Newtonsoft.Json/12.0.1

It should resolve to the exact revision

dabutvin:
this is an example that still works to repro the issue

https://api.clearlydefined.io/definitions/npm/npmjs/-/typescript/2.1.0

dabutvin:
Here's another example that shows this is still happening

https://api.clearlydefined.io/definitions/npm/npmjs/-/electron/2.0.1 is returning the definition for https://api.clearlydefined.io/definitions/npm/npmjs/-/electron/2.0.11

ignacionr:
### Implementation Notes

We need to figure out if this is happening at the definition service or at the curation service (the latter gets called by the former). After a quick read at the code, it would look like it's on the Curation service (especially given the examples which are highly curated), but it's worth creating unit tests for both separately (so we don't only fix but also get regression insurance).

dabutvin:
👍

ignacionr:
Summary: for definitions that aren't cached, they will be 'computed' out of the existing harvested data. The definition service will ask the harvest store to 'getAll' items produced for this set of coordinates (up to the revision, as a comment states). On Azure, items are stored with a name created from all parts of the coordinates, including tool name and tool version. The store component would query Azure for all items 'prefixed' with the composition up to revision, and expect to find a set of different tools used.

Turns out, the prefix will also match other revisions which completely contain the name of the revision being sought. This will happen for '12.0.1-beta' which completely contains '12.0.1' or even more odd '2.0.11' which completely contains '2.0.1'.

The proposed solution will only consider matches where right after the name we are looking for, there is anuy of:

- nothing (perfect match)
- the '.json' file extension (saved for manual inspection convenience)
- the character '/' which implies the text to the right is an indication of the tool in use

PR #519, opened 2019-05-13
investigations and tests around #312

ignacionr:
These relate to #312


dabutvin:
I just tried this out and it works!
Awesome you were able to track it down - the fix looks 👍 
I'm going to integrate this and get it rolled out right away - so happy to close this one out

fixes #312

Issue #449, opened 2019-03-18
maven url generation issue


dabutvin:
the download url is off for some packages.

for example
https://dev.clearlydefined.io/definitions/maven/mavencentral/com.fasterxml.uuid/java-uuid-generator/3.1.5

should have 
http://central.maven.org/maven2/com/fasterxml/uuid/java-uuid-generator/3.1.5/java-uuid-generator-3.1.5.jar

but we are generating
http://central.maven.org/maven2/org/com.fasterxml.uuid/java-uuid-generator/3.1.5/java-uuid-generator-3.1.5.jar

PR #524, opened 2019-05-14
Webyourmind/issue 449

ignacionr:
This should take care of #449 ; pretty confident that the new url generation is sound, because I see it used at the crawler's mavenCentralFetch -buildUrl.


dabutvin:
👍

PR #528, opened 2019-05-14
add queue processing to the service

dabutvin:
create a listener on a queue for github webhooks to process new curation events

This adds 
 - `/providers/curation/process.js` to process new curations (curations)
 - `/providers/harvest/process.js` to process new hooks from the crawlers (harvests)
 - `/providers/harvest/crawlerQueueConfig.js` to queue messages directly to the crawler

Instead of having all the crawlers call back to the service over a webhook for harvest complete notifications, moving towards the crawlers queueing messages instead.

Instead of github calling over a webhook directly to the service for curation notifications, moving towards github hooks queueing messages directly instead

Instead of calling the crawlers from the service over HTTP to pass on new harvest requests, queue the request directly into storage so any of the crawlers can pick it up.

This is the first step towards an improved resiliency/disaster recovery setup

### DIFF
--- a/providers/curation/process.js
+++ b/providers/curation/process.js
@@ -1,0 +1,45 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+
+const { get } = require('lodash')
+
+async function work() {
+  try {
+    let message = await queue.dequeue()
+    if (!get(message, 'data.pull_request') || !get(message, 'data.action')) return
+    const pr = message.data.pull_request
+    const action = message.data.action
+    switch (action) {
+      case 'opened':
+      case 'synchronize': {
+        const curations = await curationService.getContributedCurations(pr.number, pr.head.sha)
+        await curationService.validateContributions(pr.number, pr.head.sha, curations)
+        await curationService.updateContribution(pr, curations)
+        break
+      }
+      case 'closed': {
+        await curationService.updateContribution(pr)
+        break
+      }
+    }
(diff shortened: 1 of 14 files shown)


providers/curation/process.js
```diff
@@ -0,0 +1,45 @@
...
+  } finally {
+    setTimeout(work, 30000)
+  }
+}
+
+let queue
+let curationService
+let logger
+
+function setup(_queue, _curationService, _logger) {
```
iamwillbar: Should we have any protection against someone calling this method twice?

providers/curation/process.js
```diff
@@ -0,0 +1,45 @@
...
+      }
+      case 'closed': {
+        await curationService.updateContribution(pr)
+        break
+      }
+    }
+    logger.info(`Handled GitHub event "${action}" for PR#${pr.number}`)
+    await queue.delete(message)
+  } catch (error) {
+    logger.error(error)
```
iamwillbar: What will happen to the queue message in this case?

dabutvin: when you call `.dequeue()` it makes the message invisible so the next call to `.dequeue()` will get the next one
After a time interval the message becomes visible again and will be in the front of the line

the DQ count increases each time this happens, and when the DQ count exceeds 5 we delete the message

providers/curation/process.js
dabutvin: As it stands, call this method as many times as you want distinct workers to be pulling from the queue.

I feel like that behavior is appropriate, what do you think?

PR #530, opened 2019-05-18
adding curation override on merge, fixes #526

### LINKED ISSUE #526: Preview API merge licenses on files of a definition
https://github.com/clearlydefined/website/issues/709 and https://github.com/clearlydefined/website/issues/678 are reporting an error when posting a PATCH that contains changes to a license on a specific file of a definition.

The preview API, called suddenly after the change, returns a sum of the licenses instead of updating it.

ignacionr:
This should resolve #526

### DIFF
--- a/lib/curation.js
+++ b/lib/curation.js
@@ -17,7 +17,7 @@ class Curation {
   }
 
   static apply(definition, curation) {
-    utils.mergeDefinitions(definition, curation)
+    utils.mergeDefinitions(definition, curation, true)
     return definition
   }
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -108,15 +108,15 @@ function extractLicenseFromLicenseUrl(licenseUrl) {
 
 // merge the given definition onto the base definition. Be careful to handle various arrays
 // correctly by finding matching entried and merging them as appropriate.
-function mergeDefinitions(base, proposed) {
+function mergeDefinitions(base, proposed, override) {
   if (!proposed) return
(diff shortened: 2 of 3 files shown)


storrisi:
Also fixes https://github.com/clearlydefined/website/issues/709 ,  https://github.com/clearlydefined/website/issues/678

lib/curation.js
```diff
@@ -17,7 +17,7 @@ class Curation {
   }
 
   static apply(definition, curation) {
-    utils.mergeDefinitions(definition, curation)
+    utils.mergeDefinitions(definition, curation, true)
```
dabutvin: instead of overloading utils.mergeDefinitions, can we replace this with lodash merge?

`_.merge(definition, curation)`

ignacionr: The problem here is that _.merge is not smart enough to match files by path/hash as per the current structure. This could change if we also applied the change proposed to make file properties key by hash...

dabutvin: Ahh makes sense. Thanks for looking into it. We can keep the current implementation then with the overload.

dabutvin:
this didn't seem to work when doing a final round of testing :(

see https://github.com/clearlydefined/service/issues/537

ignacionr:
> this didn't seem to work when doing a final round of testing :(

Actually, it works -for the problem as was originally specified (merging licenses on files only). Your test curation applies to the package (the new description on #537 does help, so I'm on it!).

PR #532, opened 2019-05-21
added "expand" to documentation

### LINKED ISSUE #474: add ?expand=-files param to the swagger doc
the definitions routes take an expand parameter now with a valid value restricted to `-files`

we should add to swagger in a way that can be extended for future accepted values

ignacionr:
Resolves #474


dabutvin:
👍

PR #554, opened 2019-06-12
updated doc

### LINKED ISSUE #541: REST Api documentation on https://api.clearlydefined.io/api-docs seems to be outdated
On https://api.clearlydefined.io/api-docs/#/harvest/post_harvest is the request documented to issue harvesting of data. But If I try the example request, which posts something like
```
[
  {
    "tool": "source",
    "path": "git/github/microsoft/redie/194269b5b7010ad6f8dc4ef608c88128615031ca"
  }
]
```
via
```
curl -X POST "https://api.clearlydefined.io/harvest" -H "accept: */*" -H "Content-Type: application/json" -d "[{\"tool\":\"source\",\"path\":\"git/github/microsoft/redie/194269b5b7010ad6f8dc4ef608c88128615031ca\"}]"
```
I get a **400** in return with the message
```
data/0 should NOT have additional properties, data/0 should have required property 'coordinates'
```

This looks like that the documentation is outdated and does not represent the current requirement of sending coordinates instead of a path.

ignacionr:
This resolves #541 ; it's interesting since it's related to local harvesting (important for debugging purposes).

### DIFF
--- a/schemas/swagger.yaml
+++ b/schemas/swagger.yaml
@@ -411,13 +411,13 @@ components:
             properties:
               tool:
                 type: string
-              path:
+              coordinates:
                 type: string
             example:
               - tool: package
-                path: npm/npmjs/-/redie/0.3.0
+                coordinates: npm/npmjs/-/redie/0.3.0
               - tool: source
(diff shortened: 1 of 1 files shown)


geneh:
Thanks!

PR #578, opened 2019-07-10
Add ability to get user public emails from Github

storrisi:
Needed by https://github.com/clearlydefined/website/issues/778

Since the user needs to be informed if his contribution and all the commits would be attributed to him, we need to get from his profile his public emails, if at least one is available.

This change requires a new scope to the GitHub request so we might upgrade the service token.

### DIFF
--- a/routes/auth.js
+++ b/routes/auth.js
@@ -54,7 +54,7 @@ router.get('/github/start', passportOrPat(), (req, res) => {
 
 router.get('/github/finalize', passportOrPat(), async (req, res) => {
   const token = req.user.githubAccessToken
-  const { publicEmails, permissions } = await getPermissions(token, options.org)
+  const { publicEmails, permissions } = await getUserDetails(token, options.org)
   const username = req.user.username
 
   const result = JSON.stringify({ type: 'github-token', token, permissions, username, publicEmails })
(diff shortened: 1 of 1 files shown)


routes/auth.js
```diff
@@ -122,13 +127,14 @@ function usePassport() {
 }
 
 function getStrategy() {
+  console.log('getStrategy')
```
geneh: Should console.log be removed?

routes/auth.js
```diff
@@ -54,10 +54,10 @@ router.get('/github/start', passportOrPat(), (req, res) => {
 
 router.get('/github/finalize', passportOrPat(), async (req, res) => {
   const token = req.user.githubAccessToken
-  const permissions = await getPermissions(token, options.org)
+  const { publicEmails, permissions } = await getPermissions(token, options.org)
```
geneh: getPermissions should either be renamed or split into two functions since it now gets both permissions and public emails.

routes/auth.js
```diff
@@ -122,13 +127,14 @@ function usePassport() {
...
 function getStrategy() {
+  console.log('getStrategy')
   return new GitHubStrategy(
     {
       clientID: options.clientId,
       clientSecret: options.clientSecret,
       // this needs to match the callback url on the oauth app on github
       callbackURL: `${endpoints.service}/auth/github/finalize`,
-      scope: ['public_repo', 'read:user', 'read:org']
+      scope: ['public_repo', 'read:user', 'read:org', 'user', 'user:email']
```
geneh: Do we need both user and user:email scopes? Will the token's scopes need to be updated before deployment?

routes/auth.js
storrisi: I've done it order to avoid 2 separate calls to the APIs , we can think on renaming it in getUserDetails perhaps?

routes/auth.js
storrisi: The octokit documentation tells that both of them are required to get the emails.
Of course the token scope needs to be updated.

geneh: It looks like there is no additional action for this OAuth app in GitHub settings.

PR #586, opened 2019-07-25
adds patchesInError array with source of issues

### LINKED ISSUE #584: Produce error context information to help track problems
Given some context that is very difficult to reproduce, curators receive an error when they try to make a contribution.
In issues reported such as https://github.com/clearlydefined/service/issues/583 , the error is a schema validation failing (sourceLocation not an object).
Since validation errors imply that one or more curation yaml files were created, being able to produce the yaml that fails validation, would be of great help in finding the root cause.

ignacionr:
Resolves #584 by adding an array of patch sources on the returned error information object.

### DIFF
--- a/routes/curations.js
+++ b/routes/curations.js
@@ -71,11 +71,13 @@ async function updateCurations(request, response) {
   const userGithub = request.app.locals.user.github.client
   const info = request.app.locals.user.github.info
   let curationErrors = []
+  let patchesInError = []
   request.body.patches.forEach(entry => {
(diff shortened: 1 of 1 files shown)


routes/curations.js
```diff
@@ -71,11 +71,13 @@ async function updateCurations(request, response) {
   const userGithub = request.app.locals.user.github.client
   const info = request.app.locals.user.github.info
   let curationErrors = []
+  let patchesInError = []
```
geneh: It may be const here

PR #600, opened 2019-08-14
Increase origin search size

### LINKED ISSUE #599: Harvest page drop-down menus do not work well with "common" user/org names
**Issue Type:** Bug
Issue Description:

The following component is unable to be harvested via the [ClearlyDefined.io/harvest](https://clearlydefined.io/harvest) page.

**http://github.com/erica/iOS-7-Cookbook**

Selections made on "[ClearlyDefined.io/harvest](https://clearlydefined.io/harvest)"...

_Technology:_ **GitHub**

_User/Organization:_ **erica** (entered in manually, does not show up in "suggestions" list)

_Repo:_ **Can't even get to this point with the selection above.**

_Pick a tag or commit hash:_ **Can't even get to this point with the selection above.**

As I type into the "User/Organization" field it does not appear to provide a recommendation just for "erica". Many other non-relevant results are "suggested" in the drop-down such as "_ericacx_" and "_ericauuu_", but not just simple "_erica_". Furthermore, it does not appear I am able to manually enter my selection (i.e. type: "_erica_") into any of the drop-down boxes without it being in the recommended list.

Right now, I am unable to harvest this particular component using this page/facility.

I've attached the [JSON response body](https://github.com/clearlydefined/website/files/3499142/response.json.txt) from the moment I completed typing in my text of "erica" into the User/Organization field. No value for "erica" appears present in the list.

storrisi:
As mentioned in #599  , the API was returning just the 30 items, which hides some useful data.

Fixes #599

### DIFF
--- a/routes/originGitHub.js
+++ b/routes/originGitHub.js
@@ -62,14 +62,14 @@ router.get(
       const answer = await github.search.repos({
         q: `${repo || ''}+user:${login}+in:name+fork:true`,
         sort: 'stars',
-        per_page: 1000
+        per_page: 100
       })
       const result = answer.data.items.map(item => {
         return { id: item.full_name }
       })
       return response.status(200).send(result)
     }
-    const answer = await github.search.users({ q: `${login}+repos:>0`, sort: 'repositories', per_page: 1000 })
+    const answer = await github.search.users({ q: `${login}+repos:>0`, sort: 'repositories', per_page: 100 })
     const result = answer.data.items.map(item => {
       return { id: item.login }
     })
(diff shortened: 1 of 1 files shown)


routes/originGitHub.js
```diff
@@ -62,14 +62,14 @@ router.get(
       const answer = await github.search.repos({
         q: `${repo || ''}+user:${login}+in:name+fork:true`,
         sort: 'stars',
-        per_page: 100
+        per_page: 1000
```
geneh: Does it actually work? I thought 100 was the maximum?

storrisi: You are right. I've checked the doc and the maximum amount allowed is 100.

PR #601, opened 2019-08-20
[PHP] Add packagist.org support (WIP)

Fredrick-Tam:
This PR is to address issue https://github.com/clearlydefined/crawler/issues/230 by starting to enable php license crawling.

Using https://github.com/clearlydefined/service/pull/273 as a guide

### DIFF
--- a/routes/originComposer.js
+++ b/routes/originComposer.js
@@ -1,2 +1,35 @@
 // Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
 // SPDX-License-Identifier: MIT
+
+const asyncMiddleware = require('../middleware/asyncMiddleware')
+const router = require('express').Router()
+const requestPromise = require('request-promise-native')
+const { uniq } = require('lodash')
+
+router.get(
+  '/:namespace?/:name/revisions',
+  asyncMiddleware(async (request, response) => {
+    const { name } = request.params
+    const url = `https://packagist.org/packages/${name}.json`
+    const answer = await requestPromise({ url, method: 'GET', json: true })
+    const result = answer.package.versions
+    return response.status(200).send(result)
+  })
+)
+
+router.get(
+  '/:name',
+  asyncMiddleware(async (request, response) => {
+    const { name } = request.params
+    const url = `https://packagist.org/packages/${name}.json`
+    const answer = await requestPromise({ url, method: 'GET', json: true })
+    const result = answer.package
+    return response.status(200).send(result)
+  })
+)
+
+function setup() {
+  return router
+}
+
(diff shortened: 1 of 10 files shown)


routes/originComposer.js
```diff
@@ -0,0 +1,35 @@
...
+
+const asyncMiddleware = require('../middleware/asyncMiddleware')
+const router = require('express').Router()
+const requestPromise = require('request-promise-native')
+const { uniq } = require('lodash')
+
+router.get(
+  '/:namespace?/:name/revisions',
+  asyncMiddleware(async (request, response) => {
+    const { name } = request.params
```
geneh: It should probably be `const { namespace, name } = request.params`

routes/originComposer.js
```diff
@@ -0,0 +1,35 @@
...
+    const url = `https://packagist.org/packages/${name}.json`
```
geneh: It should probably be 
```
const url = https://repo.packagist.org/p/`${namespace}/${name}`.json
```

routes/originComposer.js
```diff
@@ -0,0 +1,35 @@
...
+    const answer = await requestPromise({ url, method: 'GET', json: true })
+    const result = answer.package.versions
```
geneh: Should it be packages instead of package?

geneh: The returned versions seem to be unsorted. Could you please sort them similarly to originNpm.js?

routes/originComposer.js
```diff
@@ -0,0 +1,35 @@
...
+    return response.status(200).send(uniq(result))
+  })
+)
+
+router.get(
+  '/:name',
```
geneh: It should be `/:namespace/:name`

routes/originComposer.js
```diff
@@ -0,0 +1,35 @@
...
+  asyncMiddleware(async (request, response) => {
+    const { name } = request.params
+    const url = `https://packagist.org/packages/${name}.json`
+    const answer = await requestPromise({ url, method: 'GET', json: true })
+    const result = answer.package
+    return response.status(200).send(result)
```
geneh: This whole section should be updated to be similar to the previous route but with data specific to the version.

geneh:
In addition, statsService.js should be updated

routes/originComposer.js
Fredrick-Tam: Its package because this is a more specific api endpoint https://packagist.org/apidoc#get-package-data

GET https://packagist.org/packages/[vendor]/[package].json has more package information and takes us straight to package details instead of having to search for package from list of packages returned.

PR #611, opened 2019-09-04
logging invalid curations on hook and from UI

### LINKED ISSUE #609: Curation status check fails
Status check fails for the following curation: https://github.com/clearlydefined/curated-data/pull/2316

Besides resolving this problem, it would be super helpful to add logging for whenever a status check fails.

@ariel11 FYI

ignacionr:
Resolves #609 at @geneh suggestion


geneh:
@ignacionr Thanks!

Issue #568, opened 2019-06-21
Add service support for Debian package & source


geneh:
Add service changes required for supporting Debian package & source.
More details can be found at https://github.com/clearlydefined/crawler/issues/321.

PR #612, opened 2019-09-04
Add Debian support

geneh:
Resolves https://github.com/clearlydefined/service/issues/568
Related to https://github.com/clearlydefined/crawler/pull/337 and https://github.com/clearlydefined/website/pull/772


providers/summary/clearlydefined.js
```diff
@@ -340,6 +344,43 @@ class ClearlyDescribedSummarizer {
...
+      'http://ftp.debian.org/debian/' + data.registryData.find(entry => entry.Path.includes('.orig.tar.')).Path
+    ).href
+    // There is also patches URL which is related to sources but it's not part of the schema
+    setIfValue(result, 'described.urls.download', downloadUrl)
+  }
+
+  getDebianRegistryUrl(registryData) {
+    const registryPath = registryData[0].Path
+    if (registryPath) {
+      // Example: ./pool/main/0/0ad/0ad_0.0.17-1.debian.tar.xz -> http://ftp.debian.org/debian/pool/main/0/0ad
```
jeffmendoza: Is this example correct? Should the result have the file on the end?

geneh: Yes, it's supposed to be a URL leading to a human readable website page. Unfortunately, with Debian the registry/package manager doesn't really exist, so that's the closest page we can get.

PR #665, opened 2019-11-12
patches to array on swagger

### LINKED ISSUE #663: Schema for patches seems to be wrong
In the schema, patches are defined to simply refer to a curation:

https://github.com/clearlydefined/service/blob/70b5362d448ec16e4b21a0ffabf57ab93f5d0701/schemas/swagger.yaml#L98-L99

However, in test code like

https://github.com/clearlydefined/service/blob/7292bd3057695f54ddb248f06b54157bede62973/test/providers/curation/github.js#L149-L154

patches are an array of `coordinates` / `revisions`. I guess the schema is outdated and needs to be updated accordingly.

ignacionr:
fixes #663 ; good catch from @sschuberth !

### DIFF
--- a/schemas/swagger.yaml
+++ b/schemas/swagger.yaml
@@ -96,7 +96,9 @@ paths:
                     removedDefinitions:
                       type: boolean
                 patches:
-                  $ref: /schemas/curation-1.0.json
+                  type: array
+                  items:
+                    $ref: /schemas/curation-1.0.json
       responses:
         200:
           description: A summary of the curation (PR) created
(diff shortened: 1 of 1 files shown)


schemas/swagger.yaml
```diff
@@ -96,7 +96,9 @@ paths:
                     removedDefinitions:
                       type: boolean
                 patches:
-                  $ref: /schemas/curation-1.0.json
+                  type: array
+                  items:
+                    $ref: /schemas/curation-1.0.json
```
sschuberth: Actually, I believe this is still not correct. If I understand the test code correctly, `patches` is not an array of `curation`, but an array of a `patch` object which in turn contains `coordinates` and `revisions`, and `revisions` is a map of versions to `curation`.

ignacionr: Nice catch again. An 's' was missing.

PR #678, opened 2019-11-27
Quiets noisy npm tests

### LINKED ISSUE #676: npm test is noisy
In running `npm test` locally and, apparently, in CI as demonstrated in a recent 
[build](https://clearlydefined.visualstudio.com/ClearlyDefined/_build/results?buildId=6665) there are two Deprecation warnings and one exception thrown _but_ the tests apparently pass:

```
   ✓ should handle falsy input
Utils extractDate
Deprecation warning: value provided is not in a recognized RFC2822 or ISO format.

    ✓ validates valid PR change
Exception: 
Error: Invalid curations: 
    at AzureApplicationInsightsLogger.log (/home/tmarble/src/github/clearlydefined

    ✓ getCuration should access tree with children
(node:855475) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security

```
@geneh could you please help me understand if the Exception _should_ cause **npm test** to fail -or- can be safely ignored. If the former then a separate issue should be filed to figure out how to get the exception to cause the test to fail.

In either case this bug is to find and fix the Deprecation warnings (and possibly the source of the exception).

tmarble:
Updates extractDate momentjs date parsing to use the recommended
'String+Format' method to explicity test for ISO 8601, "MM-DD-YYYY", or
"YYYY-MM-DD" date formats. FFI https://momentjs.com/docs/#/parsing/string/

Intercepts and verifies the logging output from an invalid contribution
as to quiet the 'validates invalid PR change' test.

Updates nodejs constructors for Buffer to safer alternatives in githubPRs.js per
https://nodejs.org/api/buffer.html#buffer_buffer_from_buffer_alloc_and_buffer_allocunsafe

Resolves #676

Signed-off-by: Tom Marble <tmarble@info9.net>

### DIFF
--- a/test/providers/curation/github.js
+++ b/test/providers/curation/github.js
@@ -52,10 +52,10 @@ describe('Github Curation Service', () => {
     })
     const curations = await service.getContributedCurations(42, 'testBranch')
     service.logger = { // intercept and verify invalid contribution
-      error: function(description, invalidCurations) {
+      error: function(description /* , invalidCurations */) {
         expect(description).to.be.eq('Invalid curations: ')
       }
-    };
+    }
     await service.validateContributions('42', 'testBranch', curations)
     expect(service._postCommitStatus.calledTwice).to.be.true
     expect(service._postCommitStatus.getCall(0).args[2]).to.be.eq('pending')
--- a/test/providers/curation/github.js
+++ b/test/providers/curation/github.js
@@ -51,6 +51,11 @@ describe('Github Curation Service', () => {
       return [createInvalidCuration()]
     })
     const curations = await service.getContributedCurations(42, 'testBranch')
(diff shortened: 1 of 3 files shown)


test/providers/curation/github.js
```diff
@@ -51,6 +51,11 @@ describe('Github Curation Service', () => {
       return [createInvalidCuration()]
     })
     const curations = await service.getContributedCurations(42, 'testBranch')
+    service.logger = { // intercept and verify invalid contribution
+      error: function(description /* , invalidCurations */) {
```
geneh: Let's change it to an arrow function `error: (description) =>`

tmarble: done

PR #681, opened 2019-12-04
Restoring Configuration section to README.

### LINKED ISSUE #680: README references Configuration section that does not exist
In multiple places the README references a Configuration section that
does not exist. The README should not reference a section that does
not exist. This section did exist but was removed in commit c1d49f6
"update readme and swagger".

Maybe the original section should be restored? Or maybe it is out-of-date or should be linked to something else? The commit that removed it doesn't really explain why it was removed (at least to me).

ThomasHintz:
In multiple places the README references a Configuration section that
does not exist. The README should not reference a section that does
not exist. This section did exist but was removed in commit c1d49f6
"update readme and swagger". The Configuration section has been
restored from that commit.

Fixes #680

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,13 @@ You may want to get the sample data. Clone the [Havested-data](https://github.co
 
 TBD
 
+# Configuration
+
+Configuration properties can be found at:
+
+* [service/full.env.json](https://github.com/clearlydefined/service/blob/master/full.env.json)
+* [service/bin/config.js](https://github.com/clearlydefined/service/blob/master/bin/config.js)
+
 # Contributing
 
 This project welcomes contributions and suggestions, and we've documented the details in the [contribution policy](CONTRIBUTING.md).


geneh:
The list is incomplete. Perhaps it was removed to make sure there are no discrepancies between README and the actual configs.

ThomasHintz:
What is the preference for it then? Is it just incomplete or also inaccurate? I see a few options:

1. Remove the Configuration section and all references to it. I don't prefer this since the I found the removed part to actually be useful. Unless that info is somewhere else?
2. Add the missing options
3. Keep the Configuration section as is with a note that there are more options

geneh:
Please check
https://github.com/clearlydefined/service/blob/master/full.env.json
https://github.com/clearlydefined/service/blob/master/bin/config.js

ThomasHintz:
It appears to be a correct list, just incomplete.

geneh:
Maybe the documentation should point to the file/s instead.

ThomasHintz:
Works for me. Is the thinking that it is just too much work to keep them in sync/not a priority?

geneh:
> Works for me. Is the thinking that it is just too much work to keep them in sync/not a priority?

Yes

ThomasHintz:
Ah of course will do.

PR #684, opened 2019-12-05
Changing github harvest recommendations to sort by best match.

### LINKED ISSUE #661: Problems persist with harvest page GitHub fields; inconsistent recommendations returned
**Issue Type:** Bug
**Issue Description:**

An effort was made to resolve in #599, but there still appears to be some problems with the GitHub fields on the [harvesting page](https://clearlydefined.io/harvest). When you enter in distinct org names such as "_gnu_" or "_erica_", the results returned in the recommendations do not appear to be sorted and there's lots of noise. In the worst case, what you type does not even show up in the list at all.

See examples below:

Example A: org "_gnu_" **does not** show up in list.

Example B: org "_erica_" shows up in list, but is not sorted to be at top.

Example C: org "_kubernetes_" only shows if the letters "_e_" and "_s_" are removed at end.

As a user, I'd naturally expect that if I enter "_gnu_" into the GH org field, that result would be at the very top. All fields that have the recommendations feature would probably work better this way. 

The more critical issue is when you cannot harvest something because it's org does not show up. 

The responses recorded from my browser are attached [here. (responses.txt)](https://github.com/clearlydefined/service/files/3823148/responses.txt)

**Browser:** Chrome 64-bit ver 78.0 on Windows

ThomasHintz:
The user/organization recommendations sorting for GitHub Harvest have
been changed from "repositories" to best match (the GitHub
default). This fixes an issue where an exact match is not the first
item in the results and a case where an exact match doesn't show up at
all in the recommendations (the user/org has few repos but their ID is
a substring of many other IDs; since we limit to first 100 results).

Fixes #661

### DIFF
--- a/routes/originGitHub.js
+++ b/routes/originGitHub.js
@@ -69,7 +69,7 @@ router.get(
       })
       return response.status(200).send(result)
     }
-    const answer = await github.search.users({ q: `${login}+repos:>0`, sort: 'repositories', per_page: 100 })
+    const answer = await github.search.users({ q: `${login}+repos:>0`, per_page: 100 })
     const result = answer.data.items.map(item => {
       return { id: item.login }
     })


geneh:
Thanks, @ThomasHintz! Next time please add me as a reviewer, not an assignee.

ThomasHintz:
Thanks @geneh! Ah yes that makes more sense.

PR #714, opened 2020-03-09
Only use github data as needed

### LINKED ISSUE #716: ClearlyDefined return 503 or slow response when we call notices api.
We have an issue with CD that needs to be addressed before notice generation goes out. It's taking 20+ seconds sometimes even service available 503 to get notice details.

The api we use is: noticeFile = await _clearlyDefinedClient.PostNoticesAsync(noticeRequest); We use batch api to send at most 200 registrations per call.

The small payload we use hit the similar issue:
component: @entur/component-library
version: 0.2.3
componentType: npm 

component: @entur/sdk
version: 0.10.1
componentType: npm 

component: @nestjs/swagger
version: 4.1.9
componentType: npm 

component: @storybook/client-api
version: 5.3.0-rc.13
componentType: npm 

component: is-sorted
version: 1.0.5
componentType: npm 

component: quirky
version: 0.1.2
componentType: npm

component: stripe
version: 8.0.1
componentType: npm 

component: watchel
version: 0.0.1
componentType: npm

brphelps:
Closes #716

### DIFF
--- a/middleware/permissions.js
+++ b/middleware/permissions.js
@@ -13,10 +13,12 @@ function permissionsCheck(permission) {
     const requiredTeams = permissions[permission]
     const intersection = requiredTeams.filter(t => (userTeams || []).includes(t))
     if (requiredTeams.length !== 0 && intersection.length == 0) {
-      return response.status(401).send(`No permission to '${permission}' (needs team membership)`)
+      const error = new Error(`No permission to '${permission}' (needs team membership)`)
+      error.status = 401
+      next(error)
+    } else {
+      next()
     }
-
-    next()
   })
 }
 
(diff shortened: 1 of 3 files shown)


middleware/permissions.js
```diff
@@ -1,21 +1,23 @@
...
-    const error = new Error(`No permission to '${permission}' (needs team membership)`)
-    error.status = 401
-    next(error)
-  }
+    const intersection = requiredTeams.filter(t => (userTeams || []).includes(t))
+    if (requiredTeams.length !== 0 && intersection.length == 0) {
+      return response.status(401).send(`No permission to '${permission}' (needs team membership)`)
+    }
+
+    next()
```
geneh: next(error) would bubble up and return the error as expected. I would refrain from changing these lines.

brphelps: @geneh  We do this in some of our other routes-- is it wrong there?

geneh: It's not completely wrong, just cleaner this way. Feel free to either change to the previous code or merge.

PR #715, opened 2020-03-09
Rampup doc for new maintainers

jeffmendoza:
First draft, will link from other places soon.


brphelps:
There is some overlap in this doc with what's in the readme in some places, do we care?

jeffmendoza:
> There is some overlap in this doc with what's in the readme in some places, do we care?

Nah, I think this is a prescriptive, follow these steps to ramp up doc, vs the readme being general info. I guess the generic stuff at the bottom should be de-duped.

PR #720, opened 2020-03-12
Dont cache chonks and add a TTL for definitions

### LINKED ISSUE #716: ClearlyDefined return 503 or slow response when we call notices api.

brphelps:
Closes #716

### DIFF
--- a/business/definitionService.js
+++ b/business/definitionService.js
@@ -82,8 +82,19 @@ class DefinitionService {
     const cached = await this.cache.get(cacheKey)
     if (cached) return cached
     const stored = await this.definitionStore.get(coordinates)
-    if (stored) await this.cache.set(cacheKey, stored)
+    if (stored) this._setDefinitionInCache(cacheKey, stored)
     return stored
+  }
+
+  async _setDefinitionInCache(cacheKey, itemToStore) {
+    // 1000 is a magic number here -- we don't want to cache very large definitions, as it can impact redis ops
+    if (itemToStore.files && itemToStore.files.length > 1000) {
+      this.logger.info('Skipping caching for key', { coordinates: itemToStore.coordinates.toString() })
+      return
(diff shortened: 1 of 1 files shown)


business/definitionService.js
```diff
@@ -82,10 +82,21 @@ class DefinitionService {
...
     if (cached) return cached
     const stored = await this.definitionStore.get(coordinates)
-    if (stored) await this.cache.set(cacheKey, stored)
+    if (stored) this._setDefinitionInCache(cacheKey, stored)
     return stored
   }
 
+  async _setDefinitionInCache(cacheKey, itemToStore) {
+    // 1000 is a magic number here -- we don't want to cache very large definitions, as it can impact redis ops
+    if (itemToStore.files && itemToStore.files.length > 1000) {
```
geneh: Can itemToStore be null/undefined?

brphelps: Nope, both callers protect against that (although one caller would just blow up before hand).

PR #721, opened 2020-03-13
Truncate search text when exceeding 100 chars

grvillic:


providers/search/azureSearch.js
```diff
@@ -3,6 +3,9 @@
 
 const requestPromise = require('request-promise-native')
 const AbstractSearch = require('./abstractSearch')
+const {
+  truncateText
+} = require('../../lib/utils')
```
jeffmendoza: I think this should be 1 line

lib/utils.js
```diff
@@ -260,6 +260,28 @@ function _normalizeVersion(version) {
...
+ * @param {number} ending - optional. Default is '...'
+ * @returns {string}
+ */
+function truncateText(str, length, ending) {
+  if (length == null) {
+    length = 100;
+  }
+  if (ending == null) {
+    ending = '...';
+  }
```
jeffmendoza: I'd take out the ending param if we are not using it. yagni

PR #770, opened 2020-10-15
Consider files named `copying` the same as `license` files

adrian-sufaru:
We only consider files named `license` (+ variants) as being valid licenses right now. This change will add `copying` (+ a few extension variants) to the list of valid license file names. The list of variants is not the same between the two base names on purpose.

Updated tests for `isLicenseFile` to check that we have same behavior for both `copying` and `license` file names.

### DIFF
--- a/test/providers/summary/fossology.js
+++ b/test/providers/summary/fossology.js
@@ -394,7 +394,8 @@ describe('FOSSologySummarizer fixtures', () => {
         }
         //{ path: 'CONTRIBUTING.md' },
         //{ path: 'gemfiles/Gemfile.test-unit.2.0.1' }
-      ]
+      ],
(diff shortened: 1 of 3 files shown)


test/providers/summary/fossology.js
```diff
@@ -394,7 +394,8 @@ describe('FOSSologySummarizer fixtures', () => {
         }
         //{ path: 'CONTRIBUTING.md' },
         //{ path: 'gemfiles/Gemfile.test-unit.2.0.1' }
-      ]
+      ],
+      licensed: { declared: 'MIT' }
```
adrian-sufaru, line 398: existing test data has a COPYING.md file that is now considered a proper license

jeffmendoza (approved):

PR #771, opened 2020-10-26
Fix npm license file are not found

MichaelTsengLZ:
Many npm packages file paths start with "package/" if we only consider the root directory, the npm package license file will be skipped.

### DIFF
--- a/business/noticeService.js
+++ b/business/noticeService.js
@@ -87,7 +87,15 @@ class NoticeService {
     if (!definition.files) return ''
     const texts = await Promise.all(
       definition.files
-        .filter(file => file.token && file.natures && file.natures.includes('license') && file.path && (file.path.indexOf('/') === -1))
+        .filter(file =>
+          file.token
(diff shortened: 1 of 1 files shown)


jeffmendoza (approved):

business/noticeService.js
```diff
@@ -87,7 +87,15 @@ class NoticeService {
...
       definition.files
-        .filter(file => file.token && file.natures && file.natures.includes('license') && file.path && (file.path.indexOf('/') === -1))
+        .filter(file =>
+          file.token
+          && file.natures
+          && file.natures.includes('license')
+          && file.path
+          && (
+            file.path.indexOf('/') === -1
+            || (definition.coordinates.type === 'npm' && file.path.startsWith('package/'))
```
adrian-sufaru, line 97: do we want to add some tests for npm or at least update the existing ones for the `package/` case?

adrian-sufaru (approved):

PR #782, opened 2020-11-24
upgrades mkdirp and nyc to more secure versions

nellshamrell:
Replaces #719

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,53 +13,187 @@
         "@babel/highlight": "^7.0.0"
       }
     },
-    "@babel/generator": {
-      "version": "7.6.3",
-      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.6.3.tgz",
-      "integrity": "sha512-hLhYbAb3pHwxjlijC4AQ7mqZdcoujiNaW7izCT04CIowHK8psN0IN8QjDv0iyFtycF5FowUOTwDloIheI25aMw==",
+    "@babel/core": {
+      "version": "7.12.8",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.12.8.tgz",
+      "integrity": "sha512-ra28JXL+5z73r1IC/t+FT1ApXU5LsulFDnTDntNfLQaScJUJmcHL5Qxm/IWanCToQk3bPWQo5bflbplU5r15pg==",
       "dev": true,
       "requires": {
-        "@babel/types": "^7.6.3",
-        "jsesc": "^2.5.1",
-        "lodash": "^4.17.13",
-        "source-map": "^0.6.1"
+        "@babel/code-frame": "^7.10.4",
+        "@babel/generator": "^7.12.5",
+        "@babel/helper-module-transforms": "^7.12.1",
+        "@babel/helpers": "^7.12.5",
+        "@babel/parser": "^7.12.7",
+        "@babel/template": "^7.12.7",
+        "@babel/traverse": "^7.12.8",
+        "@babel/types": "^7.12.7",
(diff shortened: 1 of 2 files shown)


nellshamrell:
## Gathering information about the dependency

- [ ] Find the dependency on npm
- [ ] Find the source code repo for the dependency
- [ ] Look at the releases for the dependency
- [ ] See if there are any red flags in the release notes for the upgrade (and the releases between the current version and the version to update to)

## Gathering information about how the dependency is used in our application

- [ ] Trace how the dependency is used in the code (especially what API calls interact with it)
- [ ] Check if there is a test that covers the behavior

nellshamrell:
# Basic Info

## mkdirp
* [npm page](https://www.npmjs.com/package/mkdirp)
* [github repo](https://github.com/isaacs/node-mkdirp)

## nyc
* [npm page](https://www.npmjs.com/package/nyc)
* [github repo](https://github.com/istanbuljs/nyc)

nellshamrell:
nyc is a testing dependency and the tests are running fine locally, so I think the nyc change is good to merge.

Digging in a little more on mkdirp to make sure it's ok to merge.

nellshamrell:
# Release notes for mkdirp

* Upgrading from 0.5.1 to 1.0.0
* release notes look pretty sparse, going to need to do some more manual checking https://github.com/isaacs/node-mkdirp/releases

nellshamrell:
Looks like we use mkdirp here:

**service/providers/stores/fileDefinitionStore.js**
```javascript
(...)
const mkdirp = require('mkdirp')
(...)
  async store(coordinates, definition) {
    const filePath = this._toStoragePathFromCoordinates(coordinates) + '.json'
    const dirName = path.dirname(filePath)
    await promisify(mkdirp)(dirName)
    return promisify(fs.writeFile)(filePath, JSON.stringify(definition, null, 2), 'utf8')
  }
```

And looks like this is called here:
**service/providers/stores/fileConfig.js**
```javascript
function definition(options) {
  return require('./fileDefinitionStore')(options || { location: location + '-definition' })
}
```

Which is called from here:
**service/providers/index.js**
```javascript
module.exports = {
(...)
  definition: {
    azure: require('../providers/stores/azblobConfig').definition,
    file: require('../providers/stores/fileConfig').definition,
    mongo: require('../providers/stores/mongoConfig'),
    dispatch: require('../providers/stores/dispatchConfig')
  },
  attachment: {
    azure: require('../providers/stores/azblobConfig').attachment,
    file: require('../providers/stores/fileConfig').attachment
  },
(...)
    store: {
      azure: require('../providers/stores/azblobConfig').harvest,
      file: require('../providers/stores/fileConfig').harvest
    }
  }
}
```

Now to figure out how to test this...

jeffmcaffer:
just stumbled on this...  For testing, are you happy assuming `mkdirp` is working as expected and just looking to ensure it's being called with the right args? If so, you can setup a test that overrides the file store's `require` using `proxyquire`. There are other tests in the repo that use this approach.

nellshamrell:
Found a changelog for mkdirp https://github.com/isaacs/node-mkdirp/blob/master/CHANGELOG.md

Looks like the ^1.0.0 release is a full rewrite of the module.

I'm still working on testing it in our current code.

nellshamrell:
I've broken the nyc update into it's own pull request in #783 

Because the mkdirp update is a breaking change and because this is a medium severity issue, rather than a high severity one, I'm going to delay updating mkdirp at least for the time being. To me it's not currently worth the risk of breaking our current code (if it is later upgraded to high severity, I am definitely willing to revisit it).

If anyone else has thoughts on this, feel free to re-open the PR for discussion.

PR #793, opened 2021-01-28
Returns error when coordinates are malformed

### LINKED ISSUE #742: Get components definitions fails when one of components is unformatted
When requesting definitions for a list of components that includes an unformatted entry(for example:  "npm/npmjs/redie/0.3.0") the request failes for all components and returns uninformative error: 

`{
  "error": {
    "code": "500",
    "message": "An error has occurred",
    "innererror": {}
  }
}`

nellshamrell:
Fixes #742

One this is merged, if someone makes POST to `definitions` with a malformed component, like this:

```bash
curl -X POST "https://clearlydefined.io/definitions" -H "accept: application/json" -H "Content-Type: application/json" -d "[\"npm/npmjs/redie/0.3.0\",\"npm/npmjs/-/redie/0.3.0\"]"
```

They will receive a more useful error:

```bash
curl -X POST "https://clearlydefined.io/definitions" -H "accept: application/json" -H "Content-Type: application/json" -d "[\"npm/npmjs/redie/0.3.0\",\"npm/npmjs/-/redie/0.3.0\"]"

An error occured when trying to fetch coordinates for one of the components: Coordinates npm/npmjs/redie/0.3.0 appear to be malformed. Are they missing a namespace or revision?
```

### DIFF
--- a/business/definitionService.js
+++ b/business/definitionService.js
@@ -72,7 +72,14 @@ class DefinitionService {
     if (get(existing, '_meta.schemaVersion') === currentSchema) {
       this.logger.info('computed definition available', { coordinates: coordinates.toString() })
       result = existing
-    } else result = await this.computeAndStore(coordinates)
+    } else {
+      try {
+        result = await this.computeAndStore(coordinates)
+        return this._trimDefinition(this._cast(result), expand)
+      } catch (err) {
+        throw new Error(err.message)
+      }
+    }
     return this._trimDefinition(this._cast(result), expand)
   }
 
@@ -120,10 +127,14 @@ class DefinitionService {
     const result = {}
     const promises = coordinatesList.map(
       throat(10, async coordinates => {
-        const definition = await this.get(coordinates, null, force, expand)
-        if (!definition) return
-        const key = definition.coordinates.toString()
-        result[key] = definition
+        try {
+          const definition = await this.get(coordinates, null, force, expand)
+          if (!definition) return
+          const key = definition.coordinates.toString()
+          result[key] = definition
+        } catch (err) {
+          throw new Error(err.message)
+        }
       })
     )
     await Promise.all(promises)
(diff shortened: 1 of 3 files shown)


business/definitionService.js
```diff
@@ -72,7 +72,14 @@ class DefinitionService {
     if (get(existing, '_meta.schemaVersion') === currentSchema) {
       this.logger.info('computed definition available', { coordinates: coordinates.toString() })
       result = existing
-    } else result = await this.computeAndStore(coordinates)
+    } else {
+      try {
+        result = await this.computeAndStore(coordinates)
+        return this._trimDefinition(this._cast(result), expand)
+      } catch (err) {
+        throw new Error(err.message)
```
adrian-sufaru, line 80: Does this do anything other than recreate the original Error? Not sure if the purpose is to hide the underlying call stack by overwriting from this point.

nellshamrell: That's a good point, let me experiment a bit

nellshamrell: Thank you! I was making this considerably more complicated than it had to be. I've fixed it up!

routes/definitions.js
```diff
@@ -70,6 +70,28 @@ async function listDefinitions(request, response) {
   // if running on localhost, allow a force arg for testing without webhooks to invalidate the caches
   const force = request.hostname.includes('localhost') ? request.query.force || false : false
   const expand = request.query.expand === '-files' ? '-files' : null // only support '-files' for now
+  try {
```
jeffmcaffer, line 73: Are these changes really related to this problem? I'm not a fan of this code being in the route (perhaps better in the service itself) but either way, it seems unrelated to the issue at hand

nellshamrell: It is. The problem was we weren't capturing the the errors and returning them in a way that was useful to the user. I added this with the purpose of capturing the errors when they happen during a call and then returning them to the user. Otherwise it would just error out with a 500 whenever an error happened somewhere in the callstack.

nellshamrell: I am happy to jump on a call and talk through this more if it would be useful.

adrian-sufaru: This looks similar to the code in the next method, maybe we want to use a common function to do this consistently instead of duplicating?

nellshamrell: What happened was I forgot to delete the original code when I moved it into the try catch block. It's fixed!

adrian-sufaru (approved):

PR #794, opened 2021-02-04
corrects maven url

### LINKED ISSUE #755: Maven urls are broken
For example, in [this Maven package](https://dev.clearlydefined.io/definitions/maven/mavencentral/org.assertj/assertj-core/3.16.1), the urls in the current definition point to central.maven.org which is outdated. It should point to [search.maven.org](https://search.maven.org/) like in the sourceLocation.

The current definition urls:
```
  urls:
    download: >-
      http://central.maven.org/maven2/org/assertj/assertj-core/3.16.1/assertj-core-3.16.1.jar
    registry: 'http://central.maven.org/maven2/org/assertj/assertj-core'
    version: 'http://central.maven.org/maven2/org/assertj/assertj-core/3.16.1'
```

nellshamrell:
This fixes #755 for new Maven components, but does not yet address how to correct all of the incorrect urls already in the databases. Opening as a draft until I figure out how to do that.

### DIFF
--- a/providers/summary/clearlydefined.js
+++ b/providers/summary/clearlydefined.js
@@ -140,13 +140,13 @@ class ClearlyDescribedSummarizer {
     setIfValue(
       result,
       'described.urls.registry',
-      `http://central.maven.org/maven2/${namespaceAsFolders}/${coordinates.name}`
+      `https://search.maven.org/remotecontent?filepath=com/${namespaceAsFolders}/${coordinates.name}`
     )
     setIfValue(result, 'described.urls.version', `${get(result, 'described.urls.registry')}/${coordinates.revision}`)
     setIfValue(
       result,
       'described.urls.download',
-      `http://central.maven.org/maven2/${namespaceAsFolders}/${coordinates.name}/${coordinates.revision}/${coordinates.name}-${coordinates.revision}.jar`
+      `https://search.maven.org/remotecontent?filepath=com/${namespaceAsFolders}/${coordinates.name}/${coordinates.revision}/${coordinates.name}-${coordinates.revision}.jar`
     )
     const projectSummaryLicenses =
       get(data, 'manifest.summary.licenses') || get(data, 'manifest.summary.project.licenses') // the project layer was removed in 1.2.0
@@ -245,16 +245,14 @@ class ClearlyDescribedSummarizer {
     setIfValue(
       result,
       'described.urls.registry',
-      `https://npmjs.com/package/${
-        coordinates.namespace ? coordinates.namespace + '/' + coordinates.name : coordinates.name
+      `https://npmjs.com/package/${coordinates.namespace ? coordinates.namespace + '/' + coordinates.name : coordinates.name
       }`
     )
     setIfValue(result, 'described.urls.version', `${get(result, 'described.urls.registry')}/v/${coordinates.revision}`)
     setIfValue(
       result,
       'described.urls.download',
-      `https://registry.npmjs.com/${
-        coordinates.namespace ? coordinates.namespace + '/' + coordinates.name : coordinates.name
+      `https://registry.npmjs.com/${coordinates.namespace ? coordinates.namespace + '/' + coordinates.name : coordinates.name
       }/-/${coordinates.name}-${coordinates.revision}.tgz`
     )
     const manifest = get(data, 'registryData.manifest')
--- a/providers/summary/clearlydefined.js
+++ b/providers/summary/clearlydefined.js
@@ -140,13 +140,13 @@ class ClearlyDescribedSummarizer {
     setIfValue(
       result,
       'described.urls.registry',
-      `https://search.maven.org/remotecontent?filepath=com/${namespaceAsFolders}/${coordinates.name}`
+      `https://repo1.maven.org/maven2/${namespaceAsFolders}/${coordinates.name}`
     )
     setIfValue(result, 'described.urls.version', `${get(result, 'described.urls.registry')}/${coordinates.revision}`)
     setIfValue(
       result,
       'described.urls.download',
-      `https://search.maven.org/remotecontent?filepath=com/${namespaceAsFolders}/${coordinates.name}/${coordinates.revision}/${coordinates.name}-${coordinates.revision}.jar`
+      `https://repo1.maven.org/maven2/${namespaceAsFolders}/${coordinates.name}/${coordinates.revision}/${coordinates.name}-${coordinates.revision}.jar`
     )
     const projectSummaryLicenses =
       get(data, 'manifest.summary.licenses') || get(data, 'manifest.summary.project.licenses') // the project layer was removed in 1.2.0
--- a/test/summary/clearlydefinedTests.js
+++ b/test/summary/clearlydefinedTests.js
@@ -135,7 +135,7 @@ describe('ClearlyDescribedSummarizer addComposerData', () => {
 
   it('does not declare license when manifest is silent on the license', () => {
     let result = {}
-    summarizer.addComposerData(result, { registryData: { manifest: { } } }, composerTestCoordinates)
+    summarizer.addComposerData(result, { registryData: { manifest: {} } }, composerTestCoordinates)
 
     assert.strictEqual(get(result, 'licensed.declared'), undefined)
   })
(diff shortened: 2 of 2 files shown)


nellshamrell:
We need to update the stored definition for each maven definition after this is deployed.

I've been thinking this over and considering a few different approaches.

### Use POST /definitions 

One option I considered was to 
* create a list of all of the coordinates for maven definitions (which, [according to the ClearlyDefined stats page](https://clearlydefined.io/stats) is 898,744 definitions) 
* pass that list with a [POST to /definitions](https://github.com/clearlydefined/service/blob/324017ef0ccfe73eb9d1593faf777d19b621a216/routes/definitions.js#L65) 
* include the **force** parameter, which would force a recomputation of each item in the list

The problem is that API call is limited to lists of 1000 or less and, as mentioned above, there are currently 898,744 Maven definitions. So we would need to do 899 API calls.

We could remove the limit...but I'm betting that limit is in place to avoid too much pressure on our infrastructure at one time. Even if we did 899 separate calls with 1000 coordinates each...that's still a lot to put on our infrastructure, and I don't think we need to do that to get the results we want.

### Do an update directly in Mongo DB

Another option was to do an update query directly in the Mongo DB.

```bash
db.collection_name.update(
  SELECTION_QUERY,
  UPDATED_DATA
)
```

If this were a simple find and replace like this:

```bash
db.collection_name.update(
  {'animal':'dog'},
   {$set:{'animal':'catl'}},{multi:true})
)
```

I could see this working. However, the transform that needs to happen is more complex.

A definition with these values for its urls:

```bash
      "urls": {
        "registry": "http://central.maven.org/maven2/io/clearlydefined/test",
        "version": "http://central.maven.org/maven2/io/clearlydefined/test/1.0",
        "download": "'http://central.maven.org/maven2/io/clearlydefined/test/1.0/test-1.0.jar"
      },
```

Needs to see those urls transformed to:

```bash
      "urls": {
        "registry": "https://search.maven.org/remotecontent?filepath=com/io/clearlydefined/test'",
        "version": "https://search.maven.org/remotecontent?filepath=com/io/clearlydefined/test/1.0",
        "download": "https://search.maven.org/remotecontent?filepath=com/io/clearlydefined/test/1.0/test-1.0.jar"
      },
```

We MIGHT be able to do that with a Mongo query, but it feels like we're making it harder than it needs to be, and the possibility of error (even if we test it) feels high.

### Write a custom script

I think the best approach is to write a custom script which we can use to connect to our Mongo DB and collection, find the relevant docs in the collection, and update them. This feels more thoroughly testable and easier to review.

nellshamrell:
I've created a script that will update the historic urls for existing maven definitions.

The script is in [this GitHub repo](https://github.com/clearlydefined/correct-cd-maven-urls).

The most relevant files are:

### README.md
* [README.md](https://github.com/clearlydefined/correct-cd-maven-urls)
* this contains information about how to use the repo and, at the end, information about how I tested it with data from [our ClearlyDefined dev environment](https://dev.clearlydefined.io/)

### index.js
* [bin/index.js](https://github.com/clearlydefined/correct-cd-maven-urls/blob/main/bin/index.js)
* this file handles the connection to the mongo db, the queries, and the updates

### updateUrl.js
* [src/updateUrl.js](https://github.com/clearlydefined/correct-cd-maven-urls/blob/main/src/updateUrl.js)
* [tests/updateUrlTest.js](https://github.com/clearlydefined/correct-cd-maven-urls/blob/main/tests/updateUrlTest.js)
* this parses the incorrect url and turns it into a correct url

nellshamrell:
How this should be deployed (I'm erring on the side of caution here)

## Verifying on the dev environment (and rehearsing for production)
- [ ] merge to master
- [ ] create dump of the `definitions-paged` collection in `clearlydefined` database in the `clearlydefined-dev` Azure CosmosDB account
- [ ] use `mongorestore` to create a new collection in the `clearlydefined' database from the dumped data 
* This is meant to be a "hot" copy that we can switch to if something goes wrong with the data update. While we do take periodic backups of our CosmosDB databases and collections, you have to contact Azure support to restore from a backup. By creating a new collection from the dump, we have a collection we can switch to immediately, rather than having to wait for Azure support.
- [ ] Practice switching to the "hot copy" collection by updating the env vars for the Clearly Defined dev env
- [ ] Switch back the env vars to the original collection
- [ ] Run the [update script](https://github.com/clearlydefined/correct-cd-maven-urls) on the Clearly Defined dev Cosmos DB (it will update the "definitions-paged" collection)
- [ ] Clear the cache for the ClearlyDefined dev service
- [ ] Verify the updated data through the ClearlyDefined dev environment UI

## Updating the production data
- [ ] create dump of the `definitions-paged` collection in `clearlydefined` database in the `clearlydefined-prod` Azure CosmosDB account 
- [ ] use `mongorestore` to create a new collection in the `clearlydefined' database from the dumped data (this is the "hot copy" we can switch too if something goes wrong with the update)
- [ ] Run the [update script](https://github.com/clearlydefined/correct-cd-maven-urls) on the Clearly Defined prod Cosmos DB (it will update the "definitions-paged" collection)
- [ ] Clear the cache for the prod instance of ClearlyDefined (this is in Cloudflare)
- [ ] Verify the updated data through the ClearlyDefined prod environment UI

jeffmcaffer:
You could also consider a somewhat lazy approach. There is already code in there IIRC that checks a schema version on GET /definition and if it's somehow old, it `compute` is called. Some variation on that would put this in production immediately for pre-existing defs. You'd still want to backfill so everything else is in order (search, ...) That could be done as a somewhat more slow burn

nellshamrell:
@jeffmcaffer we could take that approach...but I think we will still be getting error reports every time someone encountered one of the bad urls (even if they were fixed shortly after).

We could take a different kind of incremental approach - run the script to update 100,000 records (of the 800,000) initially, then do another 100,000 the next day - if we don't want to update them all at once.

jeffmcaffer:
My suggestion would immediately start returning the new definitions for all Maven packages. As I understand, simply recomputing the definition is enough. So put something in the code that detects if the pre-built, about to be returned definition is stale, force the recompute, and then return the new value.  Basically right here. https://github.com/clearlydefined/service/blob/master/business/definitionService.js#L72

Not sure you want to bump the `currentSchema` version as that would force all defs to be recomputed.

In the end though, you're right that the backfilling needs to be done to ensure that the document db and search stores are all updated appropriately. Reckon one could write a little script to fetch a few hundred at a time, slap it in an Action scheduled to run every 15 min and let it run. Then follow up with a db query to check the created dates on the definitions to ensure they are all after the deployment. 

Summary, it may not be worth engineering something here as it's relatively one off. If this happens more then...

disulliv (approved):
Code changes look good to me. I don't have a strong opinion about how you do the backfilling of outdated urls.

nellshamrell:
@jeffmcaffer Agreed - I see this as a one off, and don't think it requires further changes to ClearlyDefined. Should something like this happen again, we should definitely consider adding more functionality to the service to determine if a definition should be recomputed if it was computed prior to a certain date.

nellshamrell:
Thanks, @disulliv, I will merge (and prep for deployment) tomorrow.

nellshamrell:
Update on testing:

I've deployed the code and run the update script on the "definitions-paged" collection in the "clearlydefined" database of the "clearlydefined-dev" CosmosDB account.

Looking at maven/mavencentral/org.apache.maven/maven-plugin-registry/2.2.1

The urls are now correct when querying the database using mongo shell

```bash
(mongo-shell) $ db.getCollection("definitions-paged").find({_id: "maven/mavencentral/org.apache.maven/maven-plugin-registry/2.2.1"})

{ "_id" : "maven/mavencentral/org.apache.maven/maven-plugin-registry/2.2.1", 
"described" : { 
  "releaseDate" : "2009-08-06", 
  "sourceLocation" : { 
    "type" : "sourcearchive", 
    "provider" : "mavencentral", 
    "namespace" : "org.apache.maven", 
    "name" : "maven-plugin-registry", 
    "revision" : "2.2.1", 
    "url" : "https://repo1.maven.org/maven2/org/apache/maven/maven-plugin-registry/2.2.1/maven-plugin-registry-2.2.1-sources.jar" }, 
    "urls" : { 
      "registry" : "https://repo1.maven.org/maven2/org/apache/maven/maven-plugin-registry", 
      "version" : "https://repo1.maven.org/maven2/org/apache/maven/maven-plugin-registry/2.2.1", 
      "download" : "https://repo1.maven.org/maven2/org/apache/maven/maven-plugin-registry/2.2.1/maven-plugin-registry-2.2.1.jar"
```

Yet, when I request the record through the API, it still has the incorrect URLs.

```bash
$ curl https://dev-api.clearlydefined.io/definitions/maven/mavencentral/org.apache.maven/maven-plugin-registry/2.2.1

{"described":{
  "releaseDate":"2009-08-06",
  "sourceLocation":{
    "type":"sourcearchive",
    "provider":"mavencentral",
    "namespace":"org.apache.maven",
    "name":"maven-plugin-registry",
    "revision":"2.2.1",
    "url":"http://central.maven.org/maven2/org/apache/maven/maven-plugin-registry/2.2.1/maven-plugin-registry-2.2.1-sources.jar"},
    "urls":{
      "registry":"http://central.maven.org/maven2/org/apache/maven/maven-plugin-registry",
      "version":"http://central.maven.org/maven2/org/apache/maven/maven-plugin-registry/2.2.1",
      "download":"http://central.maven.org/maven2/org/apache/maven/maven-plugin-registry/2.2.1/maven-plugin-registry-2.2.1.jar"}
```

It is the same when I view the record in the browser at https://dev.clearlydefined.io/definitions/maven/mavencentral/org.apache.maven/maven-plugin-registry/2.2.1

There is clearly some sort of caching issue going on.

So far I have tried:
* rebooting the "clearlydefined-dev" Azure Cache for Redis (which temporarily broke the development environment, but is working now, see #798 for more details)
* Using the "Purge all" functionality for Cloudflare (clearlydefined.io's CDN cache is managed in Cloudflare)

Haven't had success yet, will try some more things.

nellshamrell:
Ran into some issues with updating the data in the development environment in Azure, moving conversation about this to #799

PR #801, opened 2021-02-23
limits scope of auth token 

nellshamrell:
We should only need to read user info, not write to it

Fix for #800 

This is a bit of an experiment - I cannot test it locally due to a GitHub OAuth app requiring a callback URL that is available over the open internet. Once this is merged and deployed to staging, I will test it there. If it breaks anything, I will role it back.


disulliv (approved):

nellshamrell:
Seems to work! Deploying to production!

PR #802, opened 2021-02-24
Adds ability to force a recompute of a definition

### LINKED ISSUE #799: Update historic maven package definitions
Definitions for Maven components in ClearlyDefined used to have incorrect source and other urls (see #755).

With this pull request #794, all new Maven definitions will have the correct urls. This has been merged and deployed.

However, there is still the matter of updating all of the existing Maven definitions (800,000+) to have correct URLs.

I initially took the approach of [creating a script to update the "definitions-paged" Mongo DB](https://github.com/clearlydefined/correct-cd-maven-urls). This worked in my local development environment.

However, it did not work in our [development environment](https://dev.clearlydefined.io/?sort=releaseDate&sortDesc=true)

This is because, in our development and production environments, we use Azure blob storage (along with Mongo) to store definition information.

As @jeffmcaffer put it in our internal Slack:

"the definitions come from blob storage. You could completely delete the cosmos db and ClearlyDefined would continue to work (though search would be affected).
2:12
The key takeaway here is that blob is the golden store. Everything else should be derived from that."

A better approach to update the historic data may be to:

* Modify [this API call](https://github.com/clearlydefined/service/blob/324017ef0ccfe73eb9d1593faf777d19b621a216/routes/definitions.js#L65) to take a parameter to force recomputation (and storage) of the definition. This is currently done automatically for calls to this API from a localhost, but there is not currently a way to do this outside of localhost. This should be a straightforward change.
* Download info for all 800,000 maven components into a big json file
* Write a script to parse through the 800,000 components and call POST /definitions for each of them (sleeping for at least a second or two between each one)
* Likely call the script on chunks of 100,000, rather than trying to do all 800,000 at once.

nellshamrell:
through an API call

This partially sets us up to fix #799  

### DIFF
--- a/routes/definitions.js
+++ b/routes/definitions.js
@@ -100,9 +100,7 @@ let definitionService
 function setup(definition, testFlag = false) {
   definitionService = definition
 
-  test = testFlag
-
-  if (test) {
+  if (testFlag) {
     router._getDefinition = getDefinition
   }
   return router
--- a/routes/definitions.js
+++ b/routes/definitions.js
@@ -16,9 +16,7 @@ router.get('/:type/:provider/:namespace/:name/:revision', asyncMiddleware(getDef
 async function getDefinition(request, response) {
   const coordinates = utils.toEntityCoordinatesFromRequest(request)
   const pr = request.params.pr
-  // if running on localhost, allow a force arg for testing without webhooks to invalidate the caches
-
-  const force = request.query.force ? true : false
+  const force = request.query.force
   const expand = request.query.expand === '-files' ? '-files' : null // only support '-files' for now
   const result = await definitionService.get(coordinates, pr, force, expand)
   response.status(200).send(result)
--- a/routes/definitions.js
+++ b/routes/definitions.js
@@ -17,7 +17,8 @@ async function getDefinition(request, response) {
   const coordinates = utils.toEntityCoordinatesFromRequest(request)
   const pr = request.params.pr
   // if running on localhost, allow a force arg for testing without webhooks to invalidate the caches
-  const force = request.hostname.includes('localhost') ? request.query.force || false : false
+
+  const force = request.query.force ? true : false
   const expand = request.query.expand === '-files' ? '-files' : null // only support '-files' for now
   const result = await definitionService.get(coordinates, pr, force, expand)
   response.status(200).send(result)
@@ -96,8 +97,15 @@ async function listDefinitions(request, response) {
 
 let definitionService
 
-function setup(definition) {
+function setup(definition, testFlag = false) {
   definitionService = definition
+
+  test = testFlag
+
+  if (test) {
+    router._getDefinition = getDefinition
+  }
   return router
 }
+
 module.exports = setup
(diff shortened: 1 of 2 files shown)


routes/definitions.js
```diff
@@ -96,8 +97,15 @@ async function listDefinitions(request, response) {
 
 let definitionService
 
-function setup(definition) {
+function setup(definition, testFlag = false) {
   definitionService = definition
+
+  test = testFlag
```
disulliv: Is there any reason why we need to reassign testFlag to test instead of just using the parameter?

test/routes/definitions.js
```diff
@@ -0,0 +1,77 @@
...
+
+describe('Definition route', () => {
+  it('accepts a good GET request', async () => {
+    const request = createGetRequest()
+    const response = httpMocks.createResponse()
+    const definitionService = createDefinitionService()
+    const router = createRoutes(definitionService)
+    await router._getDefinition(request, response)
+    expect(response.statusCode).to.be.eq(200)
+    //    const getDefinitionSpy = definitionService.get
```
disulliv, line 17: Could we remove these commented out lines?

nellshamrell:
Fixing what was flagged in the build

nellshamrell:
Done!

routes/definitions.js
```diff
@@ -17,7 +17,8 @@ async function getDefinition(request, response) {
   const coordinates = utils.toEntityCoordinatesFromRequest(request)
   const pr = request.params.pr
   // if running on localhost, allow a force arg for testing without webhooks to invalidate the caches
-  const force = request.hostname.includes('localhost') ? request.query.force || false : false
+
+  const force = request.query.force ? true : false
```
disulliv, line 21: Not sure what the ternary is doing here... This reads to me like `if request.query.force == true: return true, else false`. 

If `request.query.force` returns a boolean then you can drop the ternary, but if it's a string then you'll have to evaluate it like `request.query.force === 'true' ? true : false`

disulliv (changes requested):

test/routes/definitions.js
nellshamrell: Gah! Good catch!

routes/definitions.js
nellshamrell: Just fixed that :)

routes/definitions.js
nellshamrell: Good point, revising.

test/routes/definitions.js
disulliv: That's what PR reviews are for, haha. Happens to everyone.

disulliv (approved):

MichaelTsengLZ:
This looks good to me. 

The only thing that I'm not feel good about is the route test. I recently come across `supertest` to do an express route test. But I'm not sure whether that works well for clearlydefined server. If it doesn't fit, I'm really don't have a good idea to test express routes.

nellshamrell:
Hi @MichaelTsengLZ It's definitely something we can look into, but since that would be a larger change, I'd rather not introduce it in this pull request. Thank you for the insight, though, it took me awhile to figure out how to do the route tests and I would welcome an easier way!

PR #824, opened 2021-03-29
Automatically create curation PR for matching revision when PR get merged

MichaelTsengLZ:
Because 1/3 of the PRs created by engineers are incorrect. Adding versions with a wrong license in the PRs makes curators hard to change. Therefore, automatically create a new curation PR after an engineer-created PR has been reviewed, corrected, and merged.

### DIFF
--- a/providers/curation/github.js
+++ b/providers/curation/github.js
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
 // SPDX-License-Identifier: MIT
 
-const { concat, get, forIn, merge, isEqual, uniq, pick, flatten, flatMap, first, union } = require('lodash')
+const { concat, get, forIn, merge, isEqual, uniq, pick, flatten, flatMap, first, union, unset, uniqWith } = require('lodash')
 const moment = require('moment')
 const geit = require('geit')
 const yaml = require('js-yaml')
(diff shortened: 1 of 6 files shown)


providers/curation/github.js
```diff
@@ -440,27 +470,11 @@ class GitHubCurationService {
       body: `You can review the change introduced to the full definition at [ClearlyDefined](https://clearlydefined.io/curations/${number}).`
     }
     await serviceGithub.issues.createComment(comment)
-    await this._cleanCache(patch.patches)
     return result
   }
 
-  async _cleanCache(patches) {
```
MichaelTsengLZ, line 447: FYI, the reason why delete this is because the previous code has tried to delete the cache when the GitHub events come.

providers/curation/github.js
```diff
@@ -264,9 +262,8 @@ class GitHubCurationService {
     return { name, email, login }
   }
 
-  // return true if patch.skipMultiversionSearch is false and patch has 1 component and 1 revision
-  _isEligibleForMultiversionCuration(patch) {
-    return !patch.skipMultiversionSearch && patch.patches.length == 1 && Object.keys(patch.patches[0].revisions).length == 1
```
disulliv, line 269: What's the reason we're removing the skipMultiversionSearch parameter?

providers/curation/github.js
```diff
@@ -386,30 +383,63 @@ class GitHubCurationService {
...
+      }
+      this.logger.info('found additional versions to curate', {
+        coordinates: curatedCoordinates.toString(),
+        additionalRevisionCount: matchingRevisionAndReason.length
+      })
+      const info = {
+        type: 'auto',
+        summary: curatedCoordinates.toString(),
+        details: `Add ${get(curationRevisions, [revision, 'licensed', 'declared'])} license`,
+        resolution: `Automatically added versions base on ${pr.html_url}\n ${this._formatMultiversionCuratedRevisions(matchingRevisionAndReason)}`,
```
disulliv, line 421: nit: "based" instead of "base"

disulliv (changes requested):
If we're removing the skipMultiversionSearch parameter, make sure to remove it in the curation template as well (schemas/curations-1.0.json)

providers/curation/github.js
MichaelTsengLZ: GitHub event doesn't have `skipMultiversionSearch` property.

providers/curation/github.js
disulliv: Makes sense.

providers/curation/github.js
disulliv: Gotcha, if we won't ever need it then make sure you remove the parameter from the curation schema. https://github.com/clearlydefined/service/blob/master/schemas/curations-1.0.json

disulliv (approved):

PR #829, opened 2021-04-06
Maven google support

pnibakuze-ms:
Work in progress. Initial changes

### DIFF
--- a/providers/summary/clearlydefined.js
+++ b/providers/summary/clearlydefined.js
@@ -135,17 +135,19 @@ class ClearlyDescribedSummarizer {
 
   getMavenUrls(coordinates) {
     var urls = { download: "", registry: "" }
-    const namespaceAsFolders = coordinates.namespace ? coordinates.namespace.replace(/\./g, '/') : coordinates.namespace
 
     switch (coordinates.provider) {
       case 'mavencentral':
+        const namespaceAsFolders = coordinates.namespace ? coordinates.namespace.replace(/\./g, '/') : coordinates.namespace
         urls.registry = `https://repo1.maven.org/maven2/${namespaceAsFolders}/${coordinates.name}`
         urls.download = `https://repo1.maven.org/maven2/${namespaceAsFolders}/${coordinates.name}/${coordinates.revision}/${coordinates.name}-${coordinates.revision}.jar`
         break
 
+      //For Google's Maven Repo, the artifacts do not always have the same format and sometimes are missing. We are simply providing link for the package info and let the user
+      //decide on what to download
       case 'mavengoogle':
-        urls.registry = `https://maven.google.com/web/index.html#${coordinates.namespace}:${coordinates.name}`
-        urls.download = `https://dl.google.com/android/maven2/${namespaceAsFolders}/${coordinates.name}/${coordinates.revision}/${coordinates.name}-${coordinates.revision}.aar`
+        urls.registry = `https://maven.google.com/web/index.html#${coordinates.namespace}:${coordinates.name}:${coordinates.revision}`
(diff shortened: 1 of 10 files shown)


providers/summary/clearlydefined.js
```diff
@@ -133,20 +133,40 @@ class ClearlyDescribedSummarizer {
...
 
-  addMavenData(result, data, coordinates) {
+  getMavenUrls(coordinates) {
+    var urls = { download: '', registry: '' }
     const namespaceAsFolders = coordinates.namespace ? coordinates.namespace.replace(/\./g, '/') : coordinates.namespace
 
+    switch (coordinates.provider) {
+      //For Google's Maven Repo, the artifacts do not always have the same format and sometimes are missing. We are simply providing link for the package info and let the user
+      //decide on what to download
+      case 'mavengoogle':
```
nellshamrell, line 143: I would add in some tests for these urls similiar to the ones here for MavenCentral https://github.com/clearlydefined/service/blob/9e0a677a74c36c6ea276c4548b520b7c91db05ce/test/summary/clearlydefinedTests.js#L473

routes/originMavenGoogle.js
```diff
@@ -0,0 +1,35 @@
...
+      const { group, artifact } = request.params
+      const groupFullPath = `${group.replace(/\./g, '/')}`
+      const url = `https://dl.google.com/android/maven2/"${groupFullPath}"/"${artifact}"/group-index.xml`
+      const answerXml = await requestPromise({ url, method: 'GET', json: false })
+      const answer = await parseString(answerXml)
+      const result = answer.response.docs.map(item => item.v)
+      return response.status(200).send(uniq(result))
+    } catch (error) {
+      if (error.code === 404) return response.status(200).send([])
+      // TODO what to do on non-404 errors? Log for sure but what do we give back to the caller?
```
nellshamrell, line 25: I would go ahead and return the error code to the user (and possibly the error text as well). This will give them something to send us if they request support with the error.

nellshamrell (changes requested):

routes/originMavenGoogle.js
```diff
@@ -0,0 +1,35 @@
...
+
+const asyncMiddleware = require('../middleware/asyncMiddleware')
+const router = require('express').Router()
+const requestPromise = require('request-promise-native')
+const { promisify } = require('util')
+const parseString = promisify(require('xml2js').parseString)
+const { uniq } = require('lodash')
+
+// Get versions
+router.get(
```
nellshamrell, line 12: One more suggestion on this file - most of the routes for origins have a route similar to this one that allows for search similar to [this one](https://github.com/clearlydefined/service/blob/87181ac6d06a2029ed8d402bdf8731eef686e1d2/routes/originMaven.js#L29). I would add that route in, but have it return an error to the user indicating that search is not available for the source and to please request a specific revision. This will make it consistent with the other origins.

providers/summary/clearlydefined.js
pnibakuze-ms: Added test to check expected URLS

routes/originMavenGoogle.js
pnibakuze-ms: Added returning error message indicating the package name or revision could not be found

routes/originMavenGoogle.js
pnibakuze-ms: Added search route and returning error message indicating search is not supported

pnibakuze-ms:
To test new origin use example command below:
curl http://localhost:4000/origins/mavenGoogle/android.arch.navigation/navigation-common/revisions

nellshamrell (approved):
This is GREAT! Thank you again! Approving the changes, but holding off on merging until we are able to talk to someone at Google (which is in process behind the scenes).

PR #836, opened 2021-04-28
Nell/curation errors

nellshamrell:
Currently, when a pull request to https://github.com/clearlydefined/curated-data/ fails validation, the cause of the failure is not displayed to the user. 

This adds functionality so that, when a pull request fails validation, a GitHub comment is opened on the pull request with details about the failure.

### DIFF
--- a/providers/curation/github.js
+++ b/providers/curation/github.js
@@ -143,6 +143,7 @@ class GitHubCurationService {
     if (invalidCurations.length) {
       state = 'error'
       description = `Invalid curations: ${invalidCurations.map(x => x.path).join(', ')}`
+      this.logger.error(description, invalidCurations)
 
       let error_string = 'We discovered some errors in this curation when validating it:\n\n'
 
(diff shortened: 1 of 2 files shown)


providers/curation/github.js
```diff
@@ -144,6 +144,15 @@ class GitHubCurationService {
       state = 'error'
       description = `Invalid curations: ${invalidCurations.map(x => x.path).join(', ')}`
       this.logger.error(description, invalidCurations)
+
+      let error_string = 'We discovered some errors in this curation when validating it:\n\n'
+
+      for (const invalid_curation of invalidCurations) {
```
disulliv, line 150: TIL that `for ... of` exists! I've been using `forEach` and `for ... in`, but never heard of using `of` before.

Another pattern I've seen in CD is to use `map` to build the string, but I don't think it makes a difference here.

disulliv (approved):

PR #848, opened 2021-06-03
Fix bugs for multi-version curation feature

MichaelTsengLZ:
1. Skip auto curation for components with existing scanned declared license
2. Add reprocess API protection. (100 is not a very accurate number. Use it for now.)

### DIFF
--- a/routes/curations.js
+++ b/routes/curations.js
@@ -105,6 +105,11 @@ router.post('/reprocess', permissionsCheck('curate'), asyncMiddleware(reprocessM
 
 async function reprocessMergedCurations(request, respond) {
   const coordinatesArray = request.body.map(EntityCoordinates.fromString)
+  // Reprocess consume a lot of resource. Limit the size of the reprocess request.
+  const reprocessThreshold = 100;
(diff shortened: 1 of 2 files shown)


routes/curations.js
```diff
@@ -105,6 +105,11 @@ router.post('/reprocess', permissionsCheck('curate'), asyncMiddleware(reprocessM
 
 async function reprocessMergedCurations(request, respond) {
   const coordinatesArray = request.body.map(EntityCoordinates.fromString)
+  // Reprocess consume a lot of resource. Limit the size of the reprocess request.
+  const reprocessThreshold = 100;
+  if (coordinatesArray.length >= reprocessThreshold) {
```
disulliv, line 110: If it's over the threshold, can't we just cut it off at the threshold and reprocess what we can?

disulliv (approved):

routes/curations.js
MichaelTsengLZ: We could do that too. But I think because we are using a job. It might be better to make it successful or not successful. Otherwise, your job needs to change for partial success case.

disulliv: Makes sense

Issue #856, opened 2021-06-15
Many (most) Rust crates with a choice of licenses are declared incorrectly


johnbatty:
# Problem description

Many Rust crates have a choice of licenses - typically `MIT OR Apache-2.0`, following the lead of Rust itself:
https://www.rust-lang.org/policies/licenses

`clearlydefined` is often (usually) incorrectly declaring the license of these crates as `MIT AND Apache-2.0`, rather than `MIT OR Apache-2.0`.  Examples:
* [`quote 1.0.9`](https://clearlydefined.io/definitions/crate/cratesio/-/quote/1.0.9)
* [`serde 1.0.126`](https://clearlydefined.io/definitions/crate/cratesio/-/serde/1.0.126)
* [`clap 2.33.3`](https://clearlydefined.io/definitions/crate/cratesio/-/clap/2.33.3)
* [`regex 1.5.3`](https://clearlydefined.io/definitions/crate/cratesio/-/regex/1.5.3)

There are other cases that are potentially worse, where a crate with a choice of 3 licenses is again declared with `AND` rather than `OR`, e.g. `slog` and related feature crates should be [`MPL-2.0 OR MIT OR Apache-2.0`](https://github.com/slog-rs/slog/blob/master/Cargo.toml#L8), but is declared by `clearlydefined` as `MPL-2.0 AND MIT AND Apache-2.0`:
* [`slog 2.7.0`](https://clearlydefined.io/definitions/crate/cratesio/-/slog/2.7.0)

This case is worse because the declared license is more restrictive than the correct license.

# Summary analysis

The `clearlydefined` summarizer correctly identifies the package license due to extracting the declared license from `Cargo.toml`, but the `licensee` and `scancode` scanners/summarizers (a) don't understand Rust crates, so do not read licensing info from `Cargo.toml`, and (b) detect the multiple license text files on  offer and incorrectly assume that they all apply, combining their licenses using AND.

The Aggregator service combines the correct `clearlydefined` declaration with the other incorrect declarations to give an incorrect final result.

# Investigation detail

## Overview of `clearlydefined` license detection/declaration

`clearlydefined` uses multiple tools to harvest data about packages (note: looks like there's a bug where not all of these tools run on every scan):
  * `clearlydefined`
  * `fossology`
  * `licensee`
  * `scancode`
 
[...]

PR #858, opened 2021-06-22
Fix for Rust multi-license declaration, plus new Rust license tests

johnbatty:
Added some Rust multi-licensing tests to help diagnose https://github.com/clearlydefined/service/issues/856
* A couple of tests of the `clearlydefined` scanner, which show that it is correctly declaring the Rust licenses
* Another test that runs some real raw data through the multiple scanners and aggregrator.  This demonstrates that Rust multi-licenses are declared incorrectly, and is useful to run under a debugger with breakpoints to view/understand the operation.

The latter test should possibly be moved into a separate "functional" test, as it is testing real data through multiple components.  I plan to clean it up to run through multiple example data files to confirm that they are all declared as expected.

I'm not sure whether the raw data files should be in an `evidence` directory, or a `fixtures` directory, or somewhere else - I used `evidence` as there was a similar example that used this structure.

### DIFF
--- a/providers/summary/fossology.js
+++ b/providers/summary/fossology.js
@@ -92,6 +92,8 @@ class FOSSologySummarizer {
 
   _declareLicense(coordinates, result) {
     if (!result.files) return
+    // For Rust crates, leave the license declaration to the ClearlyDefined summarizer which parses Cargo.toml
+    if (get(coordinates, 'type') === 'crate') return
     // if we know this is a license file by the name of it and it has a license detected in it
     // then let's declare the license for the component
     const licenses = uniq(
--- a/providers/summary/licensee.js
+++ b/providers/summary/licensee.js
@@ -47,6 +47,8 @@ class LicenseeSummarizer {
 
   _addLicenseFromFiles(result, coordinates) {
     if (!result.files) return
+    // For Rust crates, leave the license declaration to the ClearlyDefined summarizer which parses Cargo.toml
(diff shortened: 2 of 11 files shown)


johnbatty:
@nellshamrell I have made a proposed fix for the Rust crate licensing, following our discussion.

I have modified the non-clearlydefined summarizers to avoid declaring the license for Rust crates because:
* their current logic gets it wrong (ANDing together the license types from the license files)
* they don't read `Cargo.toml` to find out what the declared license is, which has got to be the best way to get a crate license
* the `clearlydefined` summarizer does read `Cargo.toml` and declares the license from that

The tests that I have added show that this works for selection of example cases I have added.

There is a slight oddity of the SPDX parser adding unnecessary parentheses when there are 3 licenses - but I believe you've already raised this as an issue with the SPDX library: https://github.com/clearlydefined/spdx/issues/21

As we discussed, there is the (hopefully rare) possibility that the crate declares the license incorrectly, and there are files with other licenses in the repo.  However, I think that trying to deal with this case now would add significant complexity.  I'd rather fix up the existing problem first to get the licensing declared correctly for the vast majority of crates with license choices.

nellshamrell:
@johnbatty this looks fantastic! Thank you so much!

And I agree with you, based on previous work, the evidence directory seems like the best place to put the raw data files.

Approving this now, will likely deploy this tomorrow.

nellshamrell (approved):

PR #862, opened 2021-07-15
adds in architecture doc for adding go support to the ClearlyDefined …

nellshamrell:
…service

(Also adds a folder for storing architecture docs as needed)

Signed-off-by: Nell Shamrell <nellshamrell@gmail.com>

### DIFF
--- a/docs/architecture/go_components.md
+++ b/docs/architecture/go_components.md
@@ -110,6 +110,7 @@ For example: `go.uber.org/multierr v1.6.0` maps to `go/proxy.golang.org/go.uber.
 * `go.uber.org/multierr v1.6.0` maps to `go/proxy.golang.org/go.uber.org/multierr/v1.6.0`
 * `k8s.io/api v0.20.4` maps to `go/proxy.golang.org/k8s.io/api/v0.20.4`
 * `modernc.org/sqlite v1.10.8` maps to `go/proxy.golang.org/modernc.org/sqlite v1.10.8`
+* `github.com/project/package v1.2.3` maps to `go/proxy.golang.org/github.com.project/package/v1.2.3`
 
 **Examples that seem a little weird:**
 * `github.com/Azure/azure-event-hubs-go/v3 v3.2.0` maps to `go/proxy.golang.org/github.com.Azure.azure-event-hubs-go/v3/v3.2.0`
--- a/docs/architecture/go_components.md
+++ b/docs/architecture/go_components.md
@@ -1,0 +1,117 @@
+# ClearlyDefined Go Support Architecture (Service)
+
+This drafts an architecture document for adding support for Go components to ClearlyDefined.
+
+## TLDR;
+
+* Add a new "go" type to ClearlyDefined
+* Only support go modules as components (which were added in Go 1.11). [According to the Go documentation](https://blog.golang.org/using-go-modules), modules are a dependency management system that makes dependency version information explicit and easier to manage. There were other, more complicated ways of managing go dependencies prior to this, but we should only focus on modules in our first iteration. We can revisit other ways of managing go dependencies later if necessary.
+* Support "proxy.golang.org" as the first go component provider, add other providers later if needed.
+
+## Prior work
+
+@jeffmendoza previously explored how we might add support for harvesting go components to ClearlyDefined in [this Google doc](https://docs.google.com/document/d/1T2WQ_yy3k8XIHw8oMPxz9mG4ys-SVmZ65CnJLhI2geA/edit#heading=h.gjdgxs). This has been tremendously helpful as I've explored how to implement this.
+
+### No central Go repository
+
+Most (if not all) of the types of components we support (maven, gem, crate, etc.) have some sort of central package repository that components are downloaded from.
+
+Go, however, does not. Instead, go modules are pulled directly from their original source location - which is most often github.com, googlesource.com, or a few others.
+
+### Go Proxy
+
+Google does, however, run a [proxy for go modules](https://proxy.golang.org/). From the site:
+
+"The Go team is providing the following services run by Google: a module mirror for accelerating Go module downloads, an index for discovering new modules, and a global go.sum database for authenticating module content.
+
+As of Go 1.13, the go command by default downloads and authenticates modules using the Go module mirror and Go checksum database."
+
+This allows users to download module source zips and is used by default in the Go CLI tooling.
+
+It would make sense for us to use proxy.golang.org as our (first) provider for harvesting go modules.
+
+### Go paths
+
+When someone uses go modules, they have a `go.mod` file in the root of their repository. Let's look at [an example](https://github.com/influxdata/telegraf/blob/v1.19.1/go.mod):
+
+And let's look at a few of the ways that this package's dependencies are defined:
+
+```
+	cloud.google.com/go v0.56.0
+	code.cloudfoundry.org/clock v1.0.0 // indirect
+	collectd.org v0.5.0
+	github.com/Azure/azure-event-hubs-go/v3 v3.2.0
+	go.starlark.net v0.0.0-20210406145628-7a1108eaa012
+	go.uber.org/multierr v1.6.0 // indirect
+	golang.org/x/net v0.0.0-20210226172049-e18ecbb05110
+	k8s.io/api v0.20.4
+	modernc.org/sqlite v1.10.8
+```
+
+Notice the variety in paths? Google allows "vanity" package names (like `go.uber.org/multierr` - which are automatically redirected to the source by Google's proxy service.
+
+## Constructing coordinates
+
+In order to request a definition of a go component from ClearlyDefined (which will also queue up a harvest if ClearlyDefined does not have a definition for the component), a user will need to construct coordinates in the API call.
+
+### For a Maven component
+
+For example, in order to request a maven package, the API call would look like this:
+
+```bash
+curl -X GET https://clearlydefined.io/definitions/maven/mavencentral/com.apollographql.apollo3/apollo-runtime-jvm/3.0.0-dev14" -H "accept: */*"
+```
+
+In this case, the coordinates are maven/mavencentral/com.apollographql.apollo3/apollo-runtime-jvm/3.0.0-dev14
+
+Let's break this down:
+* **maven** is the **type** 
+* **mavencentral** is the **provider**
+* **com.apollographql.apollo3** is the **namespace**
+* **apollo-runtime-jvm** is the **name**
+* **3.0.0-dev14** is the **revision**
+
+When we run this API call:
+
+```bash
+curl -X GET https://clearlydefined.io/definitions/maven/mavencentral/com.apollographql.apollo3/apollo-runtime-jvm/3.0.0-dev14" -H "accept: */*"
+```
+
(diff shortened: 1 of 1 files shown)


docs/architecture/go_components.md
```diff
@@ -0,0 +1,117 @@
+# ClearlyDefined Go Support Architecture (Service)
+
+This drafts an architecture document for adding support for Go components to ClearlyDefined.
+
+## TLDR;
+
+* Add a new "go" type to ClearlyDefined
+* Only support go modules as components (which were added in Go 1.11). [According to the Go documentation](https://blog.golang.org/using-go-modules), modules are a dependency management system that makes dependency version information explicit and easier to manage. There were other, more complicated ways of managing go dependencies prior to this, but we should only focus on modules in our first iteration. We can revisit other ways of managing go dependencies later if necessary.
+* Support "proxy.golang.org" as the first go component provider, add other providers later if needed.
```
nellshamrell, line 9: This is the closet thing I can find to a central go package registry - it won't contain all packages, but we can add other providers as needed.

docs/architecture/go_components.md
```diff
@@ -0,0 +1,117 @@
...
+* **name** should be the name of the package
+* **revision** should be the component revision (note - go components seem to have multiple ways of specifying revisons - sometimes version numbers, sometimes commit hashes)
+
+Like with Maven components, we translate namespaces that have multiple slashes (like `golang.org/x`) into namepaces with periods (`golang.org.x`).
+
+For example: `go.uber.org/multierr v1.6.0` maps to `go/proxy.golang.org/go.uber.org/multierr/v1.6.0`
+
+**Examples that seem to work well:**
+
+* `collectd.org v0.5.0` maps to `go/proxy.golang.org/collectd.org/v0.5.0`
```
nellshamrell, line 106: There was some previous discussion of formatting the coordinates with the namespace as the provider - i.e. `go.uber.org/multierr v1.6.0` would map to `go/go.uber.org/-/multierr/v1.6.0`. However, this wouldn't work for packages that are imported as a single name - like `collectd.org v0.5.0` which would map to `go/-/-/collectd.org/v0.5.0`. This won't work unless we have some sort of default for the provider in cases like these.

docs/architecture/go_components.md
```diff
@@ -0,0 +1,117 @@
...
+* `go.starlark.net v0.0.0-20210406145628-7a1108eaa012` maps to `go/proxy.golang.org/go.starlark.net/v0.0.0-20210406145628-7a1108eaa012`
+* `cloud.google.com/go v0.56.0` maps to `go/proxy.golang.org/cloud.google.com/go/v0.56.0`
+* `code.cloudfoundry.org/clock v1.0.0` maps to `go/proxy.golang.org/code.cloudfoundry.org/clock/v1.0.0`
+* `go.uber.org/multierr v1.6.0` maps to `go/proxy.golang.org/go.uber.org/multierr/v1.6.0`
+* `k8s.io/api v0.20.4` maps to `go/proxy.golang.org/k8s.io/api/v0.20.4`
+* `modernc.org/sqlite v1.10.8` maps to `go/proxy.golang.org/modernc.org/sqlite v1.10.8`
+
+**Examples that seem a little weird:**
```
nellshamrell, line 114: I'm not sure what to do about these weird examples - is it worth accepting that some paths will be a little weird if most other paths will be consistent?

docs/architecture/go_components.md
jeffmendoza: >it won't contain all packages
Is this the case? I thought it is essentially a pass-through proxy to any resolvable Go package. ie: if `go get` could get it then it can be got through the proxy.

This makes me wonder if "proxy.golang.org" should be the provider or just "-" since the proxy is just a means to an end and the Go provider is essentially the internet.

docs/architecture/go_components.md
```diff
@@ -0,0 +1,117 @@
...
+
+Go's support of vanity package names makes this a little different than other components. While there are a few different approaches we could take, we should map the coordinates to what go developers know the package as.
+
+* **type** should be go
+* **provider** should be "proxy.golang.org" (we may need to add other providers in the future)
+* **namespace** should be anything that appears before the package name
+* **name** should be the name of the package
+* **revision** should be the component revision (note - go components seem to have multiple ways of specifying revisons - sometimes version numbers, sometimes commit hashes)
+
+Like with Maven components, we translate namespaces that have multiple slashes (like `golang.org/x`) into namepaces with periods (`golang.org.x`).
```
jeffmendoza, line 100: It might be worth defining what happens in the highly-unlikely case of a name collision with this conversion:
```
github.com/user/repo/pkg -> ["github.com.user.repo", "pkg"]
github.com/user.repo/pkg -> ["github.com.user.repo", "pkg"]
```

docs/architecture/go_components.md
jeffmendoza: Sorry, what is the current proposal for package names with no / ? Should it be:
`collectd.org v0.5.0` maps to `go/proxy.golang.org/-/collectd.org/v0.5.0`

docs/architecture/go_components.md
```diff
@@ -0,0 +1,117 @@
...
+* `collectd.org v0.5.0` maps to `go/proxy.golang.org/collectd.org/v0.5.0`
+* `go.starlark.net v0.0.0-20210406145628-7a1108eaa012` maps to `go/proxy.golang.org/go.starlark.net/v0.0.0-20210406145628-7a1108eaa012`
+* `cloud.google.com/go v0.56.0` maps to `go/proxy.golang.org/cloud.google.com/go/v0.56.0`
+* `code.cloudfoundry.org/clock v1.0.0` maps to `go/proxy.golang.org/code.cloudfoundry.org/clock/v1.0.0`
+* `go.uber.org/multierr v1.6.0` maps to `go/proxy.golang.org/go.uber.org/multierr/v1.6.0`
+* `k8s.io/api v0.20.4` maps to `go/proxy.golang.org/k8s.io/api/v0.20.4`
+* `modernc.org/sqlite v1.10.8` maps to `go/proxy.golang.org/modernc.org/sqlite v1.10.8`
+
+**Examples that seem a little weird:**
+* `github.com/Azure/azure-event-hubs-go/v3 v3.2.0` maps to `go/proxy.golang.org/github.com.Azure.azure-event-hubs-go/v3/v3.2.0`
```
jeffmendoza, line 115: Trailing major version as a part of the module name is going to be fairly common, but hopefully is is on the way out. It result in an awkward import in go code. It seems like putting this further up the path is cleaner and will hopefully replace it, ex: https://pkg.go.dev/github.com/google/go-github/v32/github

docs/architecture/go_components.md
```diff
@@ -0,0 +1,117 @@
...
+* `golang.org/x/net v0.0.0-20210226172049-e18ecbb05110` maps to `go/proxy.golang.org/golang.org.x/net/v0.0.0-20210226172049-e18ecbb05110`
```
jeffmendoza, line 116: This doesn't seem awkward.

docs/architecture/go_components.md
```diff
@@ -0,0 +1,117 @@
...
+**Examples that seem to work well:**
+
+* `collectd.org v0.5.0` maps to `go/proxy.golang.org/collectd.org/v0.5.0`
+* `go.starlark.net v0.0.0-20210406145628-7a1108eaa012` maps to `go/proxy.golang.org/go.starlark.net/v0.0.0-20210406145628-7a1108eaa012`
+* `cloud.google.com/go v0.56.0` maps to `go/proxy.golang.org/cloud.google.com/go/v0.56.0`
+* `code.cloudfoundry.org/clock v1.0.0` maps to `go/proxy.golang.org/code.cloudfoundry.org/clock/v1.0.0`
+* `go.uber.org/multierr v1.6.0` maps to `go/proxy.golang.org/go.uber.org/multierr/v1.6.0`
+* `k8s.io/api v0.20.4` maps to `go/proxy.golang.org/k8s.io/api/v0.20.4`
+* `modernc.org/sqlite v1.10.8` maps to `go/proxy.golang.org/modernc.org/sqlite v1.10.8`
+
```
jeffmendoza, line 113: All of these examples are domain/module-name but more common is probably github.com/org/module. Definitely make sure that this is accepted by everyone in the "works well" category. Ex:

`github.com/project/package v1.2.3` maps to `go/proxy.golang.org/github.com.project/package/v1.2.3`

docs/architecture/go_components.md
nellshamrell: According to https://proxy.golang.org/ 

"proxy.golang.org does not save all modules forever. There are a number of reasons for this, but one reason is if proxy.golang.org is not able to detect a suitable license. In this case, only a temporarily cached copy of the module will be made available, and may become unavailable if it is removed from the original source and becomes outdated. The checksums will still remain in the checksum database regardless of whether or not they have become unavailable in the mirror."

docs/architecture/go_components.md
nellshamrell: ah! That's a type on my part. `go/proxy.golang.org/-/collectd.org/v0.5.0` is correct!

nellshamrell: Fixed!

docs/architecture/go_components.md
nellshamrell: Oy....that's a good point.

docs/architecture/go_components.md
nellshamrell: Maybe I had been staring at it too long...

docs/architecture/go_components.md
nellshamrell: Gotcha. If it's hopefully on the way out, I think we can accept that it will be a little weird for older components, but less weird with new components.

docs/architecture/go_components.md
nellshamrell: Good point, I'll add that in.

nellshamrell: Done!

docs/architecture/go_components.md
jeffmendoza: Ah ok, I guess it is client-side in the go tool that grabs a package directly if it doesn't exist on the proxy. From the Go users point of view, it is transparent.

jeffmendoza: I think this might need to be influenced by the crawler design, if you can use `go get` to get the package, you may wish to not use a "provider", but if you will only access the proxy directly, then it makes sense.

jeffmendoza (approved):

docs/architecture/go_components.md
nellshamrell: I did some experimenting with accessing the proxy directly - these URLs work (and should fit our purposes)

https://proxy.golang.org/rsc.io/quote/@v/list
https://proxy.golang.org/rsc.io/quote/@v/v1.3.0.mod
https://proxy.golang.org/rsc.io/quote/@v/v1.3.0.info
https://proxy.golang.org/rsc.io/quote/@v/v1.3.0.zip

docs/architecture/go_components.md
nellshamrell: Going to open a larger issue for this, since this affects Maven and GitLab components (at least) as well.

nellshamrell: Done! #863

nellshamrell:
Discussed this in today's community meeting and it was approved :) Going to go ahead and merge this.

PR #864, opened 2021-07-22
updates go architecture doc based on what was found in implementation

nellshamrell:
I found that original proposed structure for coordinates wasn't going to work well with the vanity import paths/urls that Go allows for. This updates the architecture to better accommodate those paths.

Signed-off-by: Nell Shamrell <nellshamrell@gmail.com>

### DIFF
--- a/docs/architecture/go_components.md
+++ b/docs/architecture/go_components.md
@@ -6,7 +6,7 @@ This drafts an architecture document for adding support for Go components to Cle
 
 * Add a new "go" type to ClearlyDefined
 * Only support go modules as components (which were added in Go 1.11). [According to the Go documentation](https://blog.golang.org/using-go-modules), modules are a dependency management system that makes dependency version information explicit and easier to manage. There were other, more complicated ways of managing go dependencies prior to this, but we should only focus on modules in our first iteration. We can revisit other ways of managing go dependencies later if necessary.
-* Support "proxy.golang.org" as the first go component provider, add other providers later if needed.
+* Use the import path as the provider for the component. 
 
 ## Prior work
 
@@ -28,7 +28,7 @@ As of Go 1.13, the go command by default downloads and authenticates modules usi
 
 This allows users to download module source zips and is used by default in the Go CLI tooling.
 
-It would make sense for us to use proxy.golang.org as our (first) provider for harvesting go modules.
+It would make sense for us to use proxy.golang.org to download zips of go components (this would be done in the crawler).
 
 ### Go paths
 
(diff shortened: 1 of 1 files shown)


docs/architecture/go_components.md
```diff
@@ -92,27 +92,21 @@ A Go component's coordinates will also need to consist of a type, provider, name
...
-**Examples that seem a little weird:**
-* `github.com/Azure/azure-event-hubs-go/v3 v3.2.0` maps to `go/proxy.golang.org/github.com.Azure.azure-event-hubs-go/v3/v3.2.0`
-* `golang.org/x/net v0.0.0-20210226172049-e18ecbb05110` maps to `go/proxy.golang.org/golang.org.x/net/v0.0.0-20210226172049-e18ecbb05110`
+* `collectd.org v0.5.0` maps to `go/-/-/collectd.org/v0.5.0`
+* `go.starlark.net v0.0.0-20210406145628-7a1108eaa012` maps to `go/-/-/go.starlark.net/v0.0.0-20210406145628-7a1108eaa012`
+* `cloud.google.com/go v0.56.0` maps to `go/cloud.google.com/-/go/v0.56.0`
+* `code.cloudfoundry.org/clock v1.0.0` maps to `go/code.cloudfoundry.org/-/clock/v1.0.0`
+* `go.uber.org/multierr v1.6.0` maps to `go/go.uber.org/-/multierr/v1.6.0`
+* `github.com/project/package v1.2.3` maps to `go/github.com/project/package/v1.2.3`
+* `golang.org/x/net v0.0.0-20210226172049-e18ecbb05110` maps to `go/golang.org.x/net/v0.0.0-20210226172049-e18ecbb05110`
```
jeffmendoza, line 110: Should be `go/golang.org/x/net/v0.0.0-20210226172049-e18ecbb05110` i think

jeffmendoza:
Just a note: This isn't a result of vanity import paths. A Go module can live anywhere on the web, a module of `mydomain.com/package` could very well be the literal git repo url, and not a redirect.

Modules can still have multiple directories in the path between the host and final name of the package. ex: `domain.com/long/path/go-package`. The slash in "long/path" will need to be encoded to be a valid ClearlyDefined "namespace". 

Also, periods are valid in url's as well `domain.com/project.xyz/v2/package`. "project.xyz/v2" will need to encoded in a way that can be deterministically un-encoded.

docs/architecture/go_components.md
nellshamrell: ah- typo!

nellshamrell:
@jeffmendoza- ty so much for your input on this! I can definitely see your point, there - we may have no choice but to do some url encoding in the case of go components - I don't currently see any other way around it.

PR #865, opened 2021-07-26
Go architecture take 3

nellshamrell:
Take 3 of the Go coordinate architecture - this time using URL encoding for slashes in import paths.


disulliv (approved):
Looks good to me. The URL encoding is a fantastic idea to get around the issue with slashes.

nellshamrell:
> Looks good to me. The URL encoding is a fantastic idea to get around the issue with slashes.

That idea came from @jeffmendoza :)

PR #872, opened 2021-09-24
Add logs to analysis >100s calls

MichaelTsengLZ:
1. `/definitions` API sometimes is slow when no existing definition.
2. `/notice` API sometimes is slow.


nellshamrell (approved):
This is outstanding! Ty!

PR #873, opened 2021-09-24
Improve perf logs by adding coordinates

MichaelTsengLZ:


nellshamrell (approved):

nellshamrell:
Checked this in our dev environment and it looks good! Deploying to prod.

PR #878, opened 2021-10-01
Cache curation-data source code file tree to improve perf

MichaelTsengLZ:
When the `GET /definitions` request is as slow as 100s, there are many times that logs show around 1 minutes run time when getting the `curations-data` repo source code file tree. After dig into the `geit` npm package, I found that in order to get the file tree, there are one Github calls and 836 iterations when processing file blobs. What's more, the number of iterations will grow with the growth of curation-data repo. While the curation tree is not changed very often, cache it would help improve the perf of those slow API calls.

### DIFF
--- a/providers/curation/process.js
+++ b/providers/curation/process.js
@@ -7,6 +7,9 @@ async function work(once) {
   try {
     let message = await queue.dequeue()
     if (!get(message, 'data.pull_request') || !get(message, 'data.action')) return
+    // Wait for ten seconds because GitHub use eventual consistency so that
+    // later may not able to get PRs when event happened.
+    await new Promise(resolve => setTimeout(resolve, 10 * 1000))
     const pr = message.data.pull_request
     const action = message.data.action
     switch (action) {
--- a/providers/curation/process.js
+++ b/providers/curation/process.js
@@ -7,20 +7,21 @@ async function work(once) {
   try {
     let message = await queue.dequeue()
     if (!get(message, 'data.pull_request') || !get(message, 'data.action')) return
-    // Wait for ten seconds because GitHub use eventual consistency so that
-    // later may not able to get PRs when event happened.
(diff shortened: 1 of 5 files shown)


test/providers/curation/processTest.js
```diff
@@ -7,7 +7,8 @@ const memoryQueue = require('../../../providers/queueing/memoryQueue')
 const sinon = require('sinon')
 
 describe('Curation queue processing', () => {
-  it('handles opened message', async () => {
+  it('handles opened message', async function () {
+    this.timeout(12000)
```
nellshamrell, line 11: Can you tell me more about why you added time outs on these tests? I'm sure it's for a good reason, but I would love to understand this better.

routes/webhook.js
```diff
@@ -31,12 +31,16 @@ async function handleGitHubCall(request, response) {
       case 'opened':
       case 'reopened':
       case 'synchronize': {
+        // Wait for ten seconds because GitHub use eventual consistency so that
```
MichaelTsengLZ, line 34: I added a 10s delay here. And the default test timeout is 2000ms.

providers/curation/process.js
```diff
@@ -12,12 +12,16 @@ async function work(once) {
     switch (action) {
       case 'opened':
       case 'synchronize': {
+        // Wait for ten seconds because GitHub use eventual consistency so that
+        // later may not able to get PRs when event happened.
+        await new Promise(resolve => setTimeout(resolve, 10 * 1000))
```
MichaelTsengLZ, line 17: Same here.

test/providers/curation/processTest.js
MichaelTsengLZ: I added some delay here with comments. https://github.com/clearlydefined/service/pull/878/files#r720403603. This is some experience from the CLA-Assistance.

nellshamrell (approved):

PR #885, opened 2021-11-09
adds some logging to debug issues we are seeing in the staging enviro…

nellshamrell:
…nment and production environment

### DIFF
--- a/app.js
+++ b/app.js
@@ -155,6 +155,7 @@ function createApp(config) {
 
   // catch 404 and forward to error handler
   const requestHandler = (req, res, next) => {
+    logger.info('Error when handling a request', { request: req, response: res })
     const err = new Error('Not Found')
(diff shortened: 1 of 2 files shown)


jeffwilcox (approved):

app.js
```diff
@@ -155,6 +155,7 @@ function createApp(config) {
 
   // catch 404 and forward to error handler
   const requestHandler = (req, res, next) => {
+    logger.info('Error when handling a request', { request: req, response: res })
```
MichaelTsengLZ, line 158: Not very sure whether the logger serialized the `{ request: req, response: res}`. If yes, I think it's good. If not, I think app insights will only show `{request: [Object], response: [Object] }`.

nellshamrell: ty! will check

Issue #884, opened 2021-11-04
Issue with requesting definitions for go components with url encoding in namespaces


nellshamrell:
Looks like there is a bug in production with go components that did not show up in development - I think because we have more dependencies in the production environment.

Requests for definitions without url encoding in the namespace, such as https://clearlydefined.io/definitions/go/golang/rsc.io/quote/v1.5.2 are working just fine

However, requests for definitions with url encoding such as https://clearlydefined.io/definitions/go/golang/github.com%2fgorilla/mux/v1.7.3 are not showing definition information.

nellshamrell:
Some notes:

These three requests (which are done as part of the definition request in the UI) initially return 404s from the disk cache,  but later succeed with 204s.

* https://api.clearlydefined.io/definitions/go/golang/github.com%2fgorilla/mux/v1.7.3?expand=prs&matchCasing=false
* https://api.clearlydefined.io/curations/go/golang/github.com%2fgorilla/mux/v1.7.3?expand=prs
* https://api.clearlydefined.io/suggestions/go/golang/github.com%2fgorilla/mux/v1.7.3

nellshamrell:
It also appears that https://clearlydefined.io/definitions/go/golang/github.com%2fgorilla/mux/v1.7.3 is not triggering a request to the harvest

nellshamrell:
I believe the prod instance is handling the url encoding/decoding differently from how it was being handled locally. Continuing to troubleshoot.

nellshamrell:
Created a harvest directly with:

```bash
curl -X POST "https://dev-api.clearlydefined.io/harvest" -H "accept: */*" -H "Content-Type: application/json" -d "[{\"tool\":\"package\",\"coordinates\":\"go/golang/github.com%2fgorilla/mux/v1.7.3\"}]"
```

The harvest appears to have run fine (looking at the azure logs, all the tools (ClearlyDefined, Licensee, and Scancode) ran successfully)

nellshamrell:
Yet we still aren't able to see the definition with:

https://clearlydefined.io/definitions/go/golang/github.com%2fgorilla/mux/v1.7.3

or 

```bash
 curl https://dev-api.clearlydefined.io/definitions/go/golang/github.com%2fgorilla/mux/v1.7.3

[...]

PR #886, opened 2021-11-11
Fix go namespace slashes

nellshamrell:
# Fix go namespace slashes

## Background

This is a fix for #884 

When go support was deployed to staging and production, I discovered that when someone requests a definition with url encoded slashes, such as:

```
definitions/go/golang/github.com%2fgorilla/mux/v1.7.3
```

The load balancer decodes the slashes before the request is passed to the service, which means the coordinates requested are:

```
definitions/go/golang/github.com/gorilla/mux/v1.7.3
```

Which results in a 404 not found, as there are more parameters in the url than the route expect.

## Implementation

This fix is a bit brute force, but it does work. It allows for up to three additional parameters in the URL and, for go components only, strings the relevant arguments together, then parses the revision, name, and namespace out of that string.

### DIFF
--- a/routes/definitions.js
+++ b/routes/definitions.js
@@ -26,12 +26,14 @@ async function getDefinition(request, response) {
   const log = logger()
   log.info('getDefinition route hit', { ts: new Date().toISOString(), requestParams: request.params })
 
+  let coordinates
+
   // Painful way of handling go namespaces with multiple slashes
   // Unfortunately, it seems the best option without doing a massive
   // rearchitecture of the entire coordinate system
-  if (request.params.type == "go" && request.params.provider == "golang") {
+  if (request.params.type == 'go' && request.params.provider == 'golang') {
     let namespaceNameRevision = utils.parseNamespaceNameRevision(request)
-    let splitString = namespaceNameRevision.split("/")
+    let splitString = namespaceNameRevision.split('/')
 
     // Pull off the last part of the string as the revision
     const revision = splitString.pop()
--- a/routes/definitions.js
+++ b/routes/definitions.js
@@ -31,7 +31,7 @@ async function getDefinition(request, response) {
   // Painful way of handling go namespaces with multiple slashes
   // Unfortunately, it seems the best option without doing a massive
   // rearchitecture of the entire coordinate system
-  if (request.params.type == 'go' && request.params.provider == 'golang') {
+  if (request.params.type === 'go' && request.params.provider === 'golang') {
     let namespaceNameRevision = utils.parseNamespaceNameRevision(request)
     let splitString = namespaceNameRevision.split('/')
 
--- a/routes/definitions.js
+++ b/routes/definitions.js
@@ -13,12 +13,60 @@ const logger = require('../providers/logging/logger')
 // API for serving consumers and API
 router.get('/:type/:provider/:namespace/:name/:revision/pr/:pr', asyncMiddleware(getDefinition))
 router.get('/:type/:provider/:namespace/:name/:revision', asyncMiddleware(getDefinition))
-router.get('/:type/:provider/:namespace/:name/:revision/:extra', asyncMiddleware(getDefinition))
+
+// When a request for a component with slashes in the namespace comes in
+// They are initially encoded with url encoding like this go/golang/github.com%2fgorilla/mux/v1.7.3
+// However, when it comes through a load balancer, the load balancer sometimes decodes the encoding to
+// go/golang/github.com/gorilla/mux/v1.7.3
+// which causes routing errors unless we allow for additional fields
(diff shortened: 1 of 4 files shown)


routes/definitions.js
```diff
@@ -26,12 +26,14 @@ async function getDefinition(request, response) {
   const log = logger()
   log.info('getDefinition route hit', { ts: new Date().toISOString(), requestParams: request.params })
 
+  let coordinates
+
   // Painful way of handling go namespaces with multiple slashes
   // Unfortunately, it seems the best option without doing a massive
   // rearchitecture of the entire coordinate system
-  if (request.params.type == "go" && request.params.provider == "golang") {
+  if (request.params.type == 'go' && request.params.provider == 'golang') {
```
MichaelTsengLZ, line 34: Possibly some lint rule is not setup very well.

MichaelTsengLZ:
Feels a little brute force and we only support the go component with up to three slashes. Could we replace slash in the namespace with another char, like _ or whitespace,  and then call encodeURIComponent()?

nellshamrell:
@MichaelTsengLZ it is, indeed, brute force. I tried a few experiments with replacing the slashes with other characters. The problem is that there any character that can be used in a URL can be used in a go namespace (including underscores). 

We could consider something like whitespace, but I feel that would make the API more difficult to use (it's not intuitive to include whitespace in an API call).

I feel this approach introduces the fewest compromises and we can add more "extra" arguments to support more slashes in go namespaces.

nellshamrell:
Much of the discussion around what to do about slashes in go namespaces can be found in these three pull requests:

https://github.com/clearlydefined/service/pull/862
https://github.com/clearlydefined/service/pull/864
https://github.com/clearlydefined/service/pull/865

routes/definitions.js
jeffwilcox: Michael, do you mean that these should be triple equals to be the best JavaScript it can be? `===`
(Agree if so)

jeffwilcox (approved):
Looks good! One minor note.

jeffwilcox:
As to the brute force and other aspect... sorry, I don't have enough background on ClearlyDefined to help here, but I trust Nell and the design she has in mind for the implementation.

routes/definitions.js
MichaelTsengLZ: Good catch. the == should be ===. What I mean is the double quote can be caught by lint rules. I think == could also be caught in lint rules.

MichaelTsengLZ:
> @MichaelTsengLZ it is, indeed, brute force. I tried a few experiments with replacing the slashes with other characters. The problem is that there any character that can be used in a URL can be used in a go namespace (including underscores, hyphens, periods, etc.).
> 
> We could consider something like whitespace, but I feel that would make the API more difficult to use (it's not intuitive to include whitespace in an API call). I do think we would run into trouble when the load balancer decoded the encoding before passing the url to the service.
> 
> I feel this approach introduces the fewest compromises and we can add more "extra" arguments to support more slashes in go namespaces.

Whitespace will be encoded as %20 in the URL while slash will be encoded as %2F. Therefore, I don't think it will complicate the URL that much. But I agree this is also very hacky.

nellshamrell:
@MichaelTsengLZ I tested how the API behaves when it receives an API call with whitespace in it, it results in it not being able to find the route (similar to what was previously happening when the slashes were decoded). We would need to add in special handling for the whitespaces similar to what we are doing with slashes here:

(The 404 not found error occurs when the express app cannot find a route that corresponds to the request)
```
curl http://localhost:4000/definitions/go/golang/github.com containerd/containerd/v1.5.4
{"error":{"code":"404","message":"An error has occurred","innererror":{"status":404,"name":"Error","message":"Not Found","stack":"Error: Not Found\n    at requestHandler (/opt/service/app.js:159:17)\n    at Layer.handle [as handle_request] (/opt/service/node_modules/express/lib/router/layer.js:95:5)\n    at trim_prefix (/opt/service/node_modules/express/lib/router/index.js:317:13)\n    at /opt/service/node_modules/express/lib/router/index.js:284:7\n    at Function.process_params (/opt/service/node_modules/express/lib/router/index.js:335:12)\n    at next (/opt/service/node_modules/express/lib/router/index.js:275:10)\n    at /opt/service/node_modules/express/lib/router/index.js:635:15\n    at next (/opt/service/node_modules/express/lib/router/index.js:260:14)\n    at Function.handle (/opt/service/node_modules/express/lib/router/index.js:174:3)\n    at router (/opt/service/node_modules/express/lib/router/index.js:47:12)"}}}curl: (6) Could not resolve host: containerd
```

nellshamrell:
I have corrected the "==" to "===". That does, indeed, make it the best Javascript it can be :)

PR #889, opened 2021-11-17
Removed redundant code

qtomlinson:
@nellshamrell As per our discussion, summarizeService.summarizeAll only takes two arguments, so the result of getFilter() is not currently used.  Removed getFilter related code.


nellshamrell (approved):
Thank you!

PR #892, opened 2021-11-29
Fix NPE retrieving definition with force set to true

qtomlinson:
Need to pass in the coordinates.


qtomlinson:
@nellshamrell a minor fix.

nellshamrell (approved):
So THAT'S why the force flag hasn't been working. Ty!

Issue #860, opened 2021-07-01
PyPi components should treat - and _ as the same


disulliv:
Information found here: https://github.com/package-url/purl-spec/blame/180c46d266c45aa2bd81a2038af3f78e87bb4a25/README.rst#L679

For example, these 2 links resolve to the same thing:
https://pypi.org/project/backports.ssl_match_hostname/
https://pypi.org/project/Backports.ssl-match-hostname/

However, in Clearly Defined, these components are not equivalent:
https://clearlydefined.io/definitions/pypi/pypi/-/backports.ssl_match_hostname/3.7.0.1
https://clearlydefined.io/definitions/pypi/pypi/-/backports.ssl-match-hostname/3.7.0.1

One way to go about this is for all PyPi components, automatically resolve the component name as all lowercase and convert all _ to -'s to standardize.

nellshamrell:
This would be a good starter issue.

One idea for implementation:

Whenever a request comes in for a pypi definition (it would come into the getDefinition route](https://github.com/clearlydefined/service/blob/e2478a1626ebe6cdc13cedefd9afcf48b61b4af9/routes/definitions.js#L17) and the name of the component has an underscore (_), automatically convert the underscore into a hyphen (-).

qtomlinson:
https://www.python.org/dev/peps/pep-0503/#normalized-names mentioned that "." should also be treated same as -, and _.

nellshamrell:
This was fixed with #893

PR #891, opened 2021-11-19
Upgrade eslint to ^7.5

qtomlinson:
eslint@^7.5 supports optional chanining.  Update eslint to avoid optional chaining error.

PR #893, opened 2021-11-29
Convert compatible pypi package names to the ones used in pypi

qtomlinson:
Pypi package management system treats ".", "-", and "_" as the same. For example, the following three urls yield the same component, backports.ssl_match_hostname.
https://pypi.org/project/backports.ssl.match.hostname/
https://pypi.org/project/backports_ssl_match_hostname/
https://pypi.org/project/backports-ssl-match-hostname/

During crawling, package coordinates internally stored in the harvest store are consistent with the package management repository.  Harvested information can only be retrieved based on the same standardized coordinates.

To allow retrieval of pypi package information via compatible names, added converting input package names to the ones used in the python package index. This conversion is done at all retrieval api entry points, and by querying pypi index (same as in crawler).  The query result is cached for later reuse and set to expire after configured life span.

Tasks: https://github.com/clearlydefined/service/issues/860


qtomlinson:
@nellshamrell This pull request is dependent on https://github.com/clearlydefined/service/pull/891

nellshamrell:
hmm...just checked out this branch locally and the tests passed fine. Seeing if there is something up in the build system.

nellshamrell:
Actually, I had the wrong branch checked out. Trying again locally.

nellshamrell:
yep, looks like it's failing locally as well

qtomlinson:
Thanks for the feedback.  I will try to set up another clean branch and have a look.

nellshamrell:
Looks like the failure started in [this commit](https://github.com/clearlydefined/service/pull/893/commits/3dfbcfa192180e791cb43f687f89e516a484f985)

nellshamrell:
I think I know what it is :) 

"test/lib/entityCoordintates.js" should be "test/lib/entityCoordinates.js"

(No worries! I do stuff like that all the time!)

qtomlinson:
Thanks for catching the naming mistake.  I will fix that.

nellshamrell:
Still have a failure - but I think I know what's up. Confirming.

lib/coordinatesMapper.js
```diff
@@ -0,0 +1,29 @@
...
+
+class CoordinatesMapper {
+
+  constructor(mappers = defaultMappers, cache = defaultcache) {
+    this.mappers = mappers
+    this.cache = cache
+  }
+
+  async map(coordinates) {
+    const mapper = this.mappers[coordinates?.provider]
```
nellshamrell, line 19: There's something going on with this line that's causing the `SyntaxError: Unexpected token '.'` (which I only found by running the mocha tests in debug mode in VS Code)

nellshamrell:
You can isolate the error by running just the coordinatesMapper test with `npx nyc mocha test/lib/coordinatesMapper.js`

nellshamrell:
I think it's still not able to process the optional chaining operator when running tests, even though eslint is updated

nellshamrell:
Ah, that operator is a feature of Node 14. Checking what version of node we are using...

nellshamrell:
Yep, we're using Node 12 in our Docker builds. We can look into updating to Node 14 (might be risky, but likely worth doing sometime soon) or rewrite the code to be compatible with Node 12.

lib/coordinatesMapper.js
qtomlinson: I only see this error when parserOptions.ecmaVersion is changed to 9 in .eslintrc.json.  This should be addressed by my eslint upgrade commit.  Does eslint in build pipeline use config in formats other than json?

qtomlinson:
I will change to avoid using optional chaining in the mean time.

qtomlinson:
Just noticed Node.js installed in our build pipeline is 12.x as well.  See azure-pipelines.yml

nellshamrell:
Update to Node 14 is complete! Reviewing this again.

nellshamrell:
This overall looks great! I'm going to do some final testing tomorrow then should be able to deploy it :) Thank you so much for this outstanding work @qtomlinson!

qtomlinson:
@nellshamrell Thanks for the feedback! A couple of points to be discussed:
1.	Naming of Class: CoordinatesMapper, it maps/converts the user specified coordinates to the standard one used in the package repository.  Do you have suggestion for naming?
2.	Package location for the added classes: CoordinatesMapper and PypiCoordinatesMapper.  CoordinatesMapper is responsible for caching and delegate to the provider specific classes for coordinates retrieval.  The classes are currently in /lib.  Is there a better code location for them?
3.	The default for life span of mapped coordinates in cache is currently set to be 1 day.   Should this default changed to a longer period?

nellshamrell:
Hi @qtomlinson!

1. I think "CoordinatesMapper" is fine as a name
2. Being in /lib is fine
3. The cache being one day is also fine

Great work!

nellshamrell (approved):
This is great! I will deploy this tomorrow :)

PR #898, opened 2021-12-06
Use pypi link for source

nellshamrell:
A community member brought these two definitions to my attention:

* https://clearlydefined.io/definitions/pypi/pypi/-/python-openstackclient/5.5.1
* https://clearlydefined.io/definitions/pypi/pypi/-/zuul/3.3.0

Because a source url is not found when these definitions are harvested, they have very low scores.

This is in part because pypi does not include source urls in its metadata (at least, not in a consistent way). 

This pull request adds a default source location for pypi definitions - the package's page on the central pypi registry (which allows for downloading of the source code and (in the project's README) often also has a link to the package's source code repo. We do something similar with go packages, since the go ecosystem also does not have a consistent way to identify source urls.

Currently, the ClearlyDefined crawler does search for a source on GitHub at harvest. If a pypi package's source is stored publicly on GitHub, ClearlyDefined can often find it. This pull request does not change that behavior - if ClearlyDefined finds a source repo on GitHub, that will overwrite the pypi page as the source url.

Ideally, the ClearlyDefined crawler would be able to find source repos in other places (such as GitLab, BitBucket, etc.). I've [opened an issue on the crawler repo](https://github.com/clearlydefined/crawler/issues/448) documenting the need for this.

Additionally, in the ClearlyDefined UI as it is now, there is no way to add a source url unless it is a GitHub repo. I've [opened an issue on the website repo](https://github.com/clearlydefined/website/issues/921) documenting the need for other options.

### DIFF
--- a/test/lib/util.js
+++ b/test/lib/util.js
@@ -716,4 +716,19 @@ describe('Utils buildSourceUrl', () => {
       expect(result).to.eq('https://pkg.go.dev/clearlydefined/foo/service@v1.2.3')
     })
   })
+
+  it('returns the correct pypi source url', () => {
+    const args = {
+      type: 'pypi',
+      provider: 'pypi',
+      namespace: '-',
+      name: 'zuul',
+      revision: '3.3.0'
+    }
+
+    const coordinates = utils.toEntityCoordinatesFromArgs(args)
+    const result = utils.buildSourceUrl(coordinates)
+
+    expect(result).to.eq('https://pypi.org/project/zuul/3.3.0/')
+  })
 })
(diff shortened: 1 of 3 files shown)


test/lib/util.js
```diff
@@ -605,3 +605,130 @@ describe('Utils getLicenseLocations', () => {
     })
   })
 })
+
+describe('Utils buildSourceUrl', () => {
```
nellshamrell, line 609: While I was in here adding a test case for pypi packages, I figured I would add in some test cases for existing behavior as well.

disulliv (approved):

PR #903, opened 2022-01-06
Add a different rate limit for batch API requests

nellshamrell:
Adds a second configurable API limit to the ClearlyDefined Service.

Some of the performance problems we have seen with the ClearlyDefined infrastructure are due to heavy use of our "batch" API calls - calls which submit a list of definitions, curations, or notices. These require processing of all items submitted to the API and, when these are used at the level we saw in December, can cause performance availability issues for the service as a whole.

These requests include:

* [POST /definitions](https://github.com/clearlydefined/service/blob/719e78bdf315101d1473132adf57900845c10092/routes/definitions.js#L110) (when listing definitions)
* [POST /curations](https://github.com/clearlydefined/service/blob/719e78bdf315101d1473132adf57900845c10092/routes/curations.js#L57) (when listing curations)
* [POST /notices](https://github.com/clearlydefined/service/blob/719e78bdf315101d1473132adf57900845c10092/routes/notices.js#L11)

NOTE: there are multiple POST routes for both definitions and curations. I have confirmed that the batch limit only applies to the desired urls.

When this pull request is deployed, we will need to add two more environmental variables to the config for the Azure App Service which runs the ClearlyDefined Service
* BATCH_RATE_LIMIT_WINDOW 
* BATCH_RATE_LIMIT_MAX


disulliv (approved):

nellshamrell:
@MichaelTsengLZ that sounds reasonable, I'll look into implementing that tomorrow.

nellshamrell:
rate-limit-redis is, so far, not playing particularly nicely with Azure Cache for Redis. Trying a few more things, but we might need to consider a different option if those don't work.

MichaelTsengLZ:
> rate-limit-redis is, so far, not playing particularly nicely with Azure Cache for Redis. Trying a few more things, but we might need to consider a different option if those don't work.

How about something like this. Create an Azure Redis `ioredis` instance and inject the whole client in?

image

nellshamrell:
@MichaelTsengLZ this is ready for review again :)

MichaelTsengLZ (approved):
Ship it. 🚢

nellshamrell:
I will ship it first thing tomorrow :)

nellshamrell:
I have added these two configuration variables to both our dev and prod app services:
* BATCH_RATE_LIMIT_MAX
* BATCH_RATE_LIMIT_WINDOW

Issue #901, opened 2022-01-04
Inconsistency when querying the same curation via GET or POST API


sschuberth:
As initially [reported on Discord](https://discordapp.com/channels/414106994996281365/419968097391345664/927869323597475871), I'm getting inconsistent results when querying the same curation via GET or POST API from the development server.

A single GET request returns `Apache-1.0` as a declared license, see for `npm/npmjs/@nestjs/platform-express/6.2.3`, see

![image](https://user-images.githubusercontent.com/349154/148122619-d10d19c4-82b8-42ab-a1b6-f3a7518f2fd1.png)

But a bulk POST request (with a single coordinate) returns `Apache-2.0` for the same component, see

![image](https://user-images.githubusercontent.com/349154/148122719-50fbdfcf-2f3e-463c-ae1e-e5fa9cd1badf.png)

This looks to me like a bug in the service.

sschuberth:
I'm seeing this again when requesting a curation for `maven/mavencentral/io.netty/netty-buffer/4.1.34.Final`, but what I'm getting back is the curation for `maven/mavencentral/io.netty/netty-buffer/4.1.52.Final`. That sounds to me like a rather serious data problem. Can anyone please have a look at this?

sschuberth:
@capfei @nellshamrell any update on this? This became a blocker for us to use ClearlyDefined at all 😞

sschuberth:
I'm also seeing some strange behavior on the *production* server that might be related: When requesting a batch of curations for

```
[
  "maven/mavencentral/javax.servlet/javax.servlet-api/3.1.0"
]
```

The response is

```
{
  "maven/mavencentral/javax.servlet/javax.servlet-api/3.1.0": {
    "curations": {
      "maven/mavencentral/javax.servlet/javax.servlet-api/3.0.1": {
        "licensed": {
          "declared": "CDDL-1.1 OR GPL-2.0-only WITH Classpath-exception-2.0"
        }
      },
      "maven/mavencentral/javax.servlet/javax.servlet-api/3.1.0": {
        "described": {
          "sourceLocation": {
            "name": "servlet-spec",
            "namespace": "javaee",
            "provider": "github",
            "revision": "bf383f0b9274365c512b31d2e7cd29224e405515",
            "type": "git",
            "url": "https://github.com/javaee/servlet-spec/commit/bf383f0b9274365c512b31d2e7cd29224e405515"
          }
        },
        "files": [
          {
            "attributions": [
[...]

PR #925, opened 2022-02-16
Retrieve closed pull requests from oldest to newest during syncAllContributions

qtomlinson:
This is to ensure when the merged pull requests are used to update
curation db sequentially, older data will be correctly overwritten
by newer data.

Tasks: https://github.com/clearlydefined/service/issues/901


nellshamrell (approved):
Ty! This is super helpful!

PR #930, opened 2022-03-08
Added parsing date with SQL format

qtomlinson:
Service side of fix for release dates for pods.
Task: https://github.com/clearlydefined/crawler/issues/450


qtomlinson:
@jeffwilcox Ready for review.  Service side of the fix for release dates for pods.  Fix is also needed in PodExtract in crawler. See https://github.com/clearlydefined/crawler/pull/469.

PR #934, opened 2022-04-12
Fix described.urls.version for pod

qtomlinson:
Task: https://github.com/clearlydefined/website/issues/963


qtomlinson:
@disulliv ready for review.  This is to fix cocoapod: "the version number does not have a link to the component like the other types do" in https://github.com/clearlydefined/service/pull/934.

disulliv (approved):

PR #941, opened 2022-06-23
Propose to change NOASSERTION to OTHER for Google PATENTS file

MichaelTsengLZ:
All Go components start with "golang.org/x" and have this [PATENT](https://cs.opensource.google/go/x/tools/+/master:PATENTS) file, Scancode could detect it as `google-patent-license-golang` file. Instead of convert `google-patent-license-golang` to NOASSERTION, I propose to change it to OTHER since it's a patent file.

### DIFF
--- a/lib/scancodeMap.js
+++ b/lib/scancodeMap.js
@@ -391,5 +391,6 @@ module.exports = new Map([
   ['zlib', 'Zlib'],
   ['zpl-2.0', 'ZPL-2.0'],
   ['zpl-1.1', 'ZPL-1.1'],
-  ['zpl-2.1', 'ZPL-2.1']
+  ['zpl-2.1', 'ZPL-2.1'],
+  ['google-patent-license-golang', 'OTHER']
 ])


qtomlinson:
This looks reasonable to me.  @bduranc Could you please comment on categorizing `google-patent-license-golang` to OTHER license?

bduranc:
This is an interesting proposal. Changing these to "OTHER" would probably be the eventual outcome anyway since we can't map these cases to a valid SPDX expression.

Sounds reasonable to me. But only as long as we can ensure these Patent License cases are the real deal and not false-positives.

Let's bring it up on the next group call and see what the others have to say.

If we do proceed this way, I think it would also be best to find a way to track these changes somehow. Perhaps, if possible, we could create one large curation PR that has all these changes in bulk or something similar.

MichaelTsengLZ:
> This is an interesting proposal. Changing these to "OTHER" would probably be the eventual outcome anyway since we can't map these cases to a valid SPDX expression.
> 
> Sounds reasonable to me. But only as long as we can ensure these Patent License cases are the real deal and not false-positives.
> 
> Let's bring it up on the next group call and see what the others have to say.
> 
> If we do proceed this way, I think it would also be best to find a way to track these changes somehow. Perhaps, if possible, we could create one large curation PR that has all these changes in bulk or something similar.

What do you mean by saying create one large curation PR? I think if we need to review the PATENT files, then I think it's better to just add those PATENT file name detected by Scancode one by one to the `lib/scancodeMap.js` file.

bduranc:
@MichaelTsengLZ : Sorry, I guess there is a bit of confusion on my end or some aspect I don't understand. 
Is this something we plan to do retro-actively, or just for any new packages scanned?

I was more just referring to the fact that we should at least have a log of what we change (I guess this itself comes from the perhaps wrong assumption that we'd be doing retro-active changes to already harvest components), that is all.

MichaelTsengLZ:
> @MichaelTsengLZ : Sorry, I guess there is a bit of confusion on my end or some aspect I don't understand. Is this something we plan to do retro-actively, or just for any new packages scanned?
> 
> I was more just referring to the fact that we should at least have a log of what we change (I guess this itself comes from the perhaps wrong assumption that we'd be doing retro-active changes to already harvest components), that is all.

It will work for new packages scanned. For those existing components with PATENT files, I think we could have a big curate PR to fix those, as you suggest. Or trigger reprocesses (not recrawl) to fix those.

MichaelTsengLZ:
@qtomlinson @bduranc Looks like the next meeting is July 18, which is far away. If changing all PATENT files from NOASSERTION to OTHER is a little bit risky. How about just changing `google-patent-license-golang` PATENT file from NOASEERTION to OTHER? 

I'm in the middle of developing a feature. And need to improve the coverage of Go components before deploying the feature.

qtomlinson:
> just changing `google-patent-license-golang` PATENT file from NOASEERTION to OTHER?

I am ok with it if @bduranc agrees.

bduranc:
@qtomlinson and @MichaelTsengLZ :

> > just changing `google-patent-license-golang` PATENT file from NOASEERTION to OTHER?
> 
> I am ok with it if @bduranc agrees.

I believe it's always best to start small with changes like this. The proposal as stated here sounds fine to me.

However, in the spirit of transparency, I think we should still give the others an opportunity to chime in somehow.

Let's at least get @capfei , @AE49 , @ariel11 's thoughts on this.

qtomlinson (approved):

PR #943, opened 2022-07-09
Fix curation issue for Go components with %2f in namespace

MichaelTsengLZ:
1. When Go components with %2f in the namespace, the curation YAML file path of the component in the [Git repo](https://github.com/clearlydefined/curated-data) could not be found.

For example, this component `go/golang/github.com%2Fcoreos/etcd/v3.3.10+incompatible` will generate path `curations/go/golang/github.com%2Fcoreos/etcd/v3.3.10+incompatible`
However, the actual file path is `go/golang/github.com/coreos/etcd/v3.3.10+incompatible` when fetched by `geit` package. We should replace `%2f or %2F` with `/` in the path.

2. Fix a small bug. `_cleanCurationTree()` requires a number or string but `_cleanCurationTree(pr)` pr is an object.

### DIFF
--- a/providers/curation/github.js
+++ b/providers/curation/github.js
@@ -15,6 +15,7 @@ const logger = require('../logging/logger')
 const semver = require('semver')
 const { LicenseMatcher } = require('../../lib/licenseMatcher')
 const { Cache } = require('memory-cache')
+const { deCodeSlashes } = require('../../lib/utils')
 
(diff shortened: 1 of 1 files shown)


disulliv (approved):

qtomlinson (approved):
may be worthwhile to add a unit test (see githubPRs.js)

PR #944, opened 2022-07-13
Fixed coordinates parsing for some go components in preview definitions API

qtomlinson:
The existing preview definition post API:
https://api.clearlydefined.io/api-docs/#/definitions/post_definitions__type___provider___namespace___name___revision_
does not work well for certain go components on the dev and production
systems.  For example, the following call failed on dev and production
deployments:
Post to, https://dev-api.clearlydefined.io/definitions/go/golang/github.com%2fdgrijalva/jwt-go/v3.2.0+incompatible?matchCasing=false&preview=true
Post body, {described: {releaseDate: "2022-06-04"}}

In those systems, the Cloudfare reportedly translates %2f
in the URL to /. Therefore, the existing preview definition post API
is not matched, resulting in 404 status.

The coordinates are the identifiers for the definitions resource.  Fixed parsing of the coordinates to allow for '/' present in the namespace for go components.

Task: https://github.com/clearlydefined/website/issues/920

### DIFF
--- a/routes/definitions.js
+++ b/routes/definitions.js
@@ -12,7 +12,7 @@ const pathToParams = require('../middleware/pathToParams')
 
 // Gets the definition for a component with any applicable patches. This is the main
 // API for serving consumers and API
-router.get('/:type/:provider/:namespace/:name/:revision/pr/:pr', asyncMiddleware(getDefinition))
+router.get('/*/pr/:pr', pathToParams, asyncMiddleware(getDefinition))
 router.get('/:type/:provider/:namespace/:name/:revision', asyncMiddleware(getDefinition))
 
 // When a request for a component with slashes in the namespace comes in
@@ -144,7 +144,7 @@ function adaptResultKeys(result, requestedKeys, coordinatesLookup, matchCase) {
 
 // Previews the definition for a component aggregated and with the POST'd curation applied.
 // Typically used by a UI to preview the effect of a patch
-router.post('/*', [pathToParams, asyncMiddleware(previewWithCoordinatesParams)])
+router.post('/*', pathToParams, asyncMiddleware(previewWithCoordinatesParams))
 async function previewWithCoordinatesParams(request, response) {
   if (!request.query.preview)
     return response.status(400).send('Only valid for previews. Use the "preview" query parameter')
--- a/routes/definitions.js
+++ b/routes/definitions.js
@@ -8,6 +8,7 @@ const utils = require('../lib/utils')
 const EntityCoordinates = require('../lib/entityCoordinates')
 const validator = require('../schemas/validator')
 const logger = require('../providers/logging/logger')
+const pathToParams = require('../middleware/pathToParams')
 
 // Gets the definition for a component with any applicable patches. This is the main
 // API for serving consumers and API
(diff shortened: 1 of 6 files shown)


qtomlinson:
@MichaelTsengLZ @disulliv ready for review :)

routes/definitions.js
```diff
@@ -95,15 +95,32 @@ async function reload(request, response) {
...
+  const result = await definitionService.compute(coordinates, curation)
+  response.status(200).send(result)
+}
+
+// Previews the definition for a component aggregated and with the given curation applied.
+// Typically used by a UI to preview the effect of a patch
+// The request body contains:
+// - coordinates: component coordinates as a string
+// - curation: the curation to be applied
+router.patch('/', asyncMiddleware(previewWithCoordinates))
```
MichaelTsengLZ, line 118: Instead of using the patch, can we use post? `patch` is basically for updates. If conflict with other routes, we could rename the route to be `/preview`.

MichaelTsengLZ:
BTW, this alone could not solve the issue that the UI could not create curation for Go components. In `providers/curation/github.js`, we probably need to add a new version of API like `definitions-1.0.0` to put `coordinates` in the query parameter instead of query path.

routes/definitions.js
qtomlinson: I originally did it as a Post call to definitions/preview; however, after reviewing the API [swagger-ui](https://api.clearlydefined.io/api-docs/#), we don't tend to add any path (e.g. /preview) in api endpoints.  In general, Patch is for partial update, and Post is for creating new resources (see https://www.w3schools.com/tags/ref_httpmethods.asp).  That is why I put it as a patch call.

If you prefer to have it as a post call,  I can change it back.

qtomlinson:
> BTW, this alone may not solve all the issues in the UI for Go components.

Agreed.  In an attempt to keep commits small, this commit focuses on what is observed in https://github.com/clearlydefined/website/issues/920.

routes/definitions.js
qtomlinson: As per suggestion, updated to a post call to definitions/preview.  See https://github.com/clearlydefined/service/compare/9d340fd6ca13243a8386b31b9143f6a496a8ef07..81433237d5456b8ee650f6e0104283c3a87580b2

qtomlinson:
> we probably need to add a new version of API like `definitions-1.0.0` to put `coordinates` in the query parameter instead of the query path for all those get routes, like `get('/:type/:provider/:namespace/:name/:revision/pr/:pr')`

The `get('/:type/:provider/:namespace/:name/:revision/pr/:pr')` api seems to fit well in the newly introduced preview post API.  Instead of the curation object, caller pass in the pr number.  That is 
Post to definitions/preview
Post body: 
{
"coordinates": "go/golang/github.com%2fdgrijalva/jwt-go/v3.2.0+incompatible",
"pr": prNumberToApply
}
What do you think?

routes/definitions.js
qtomlinson: Another thought, to align with coordinates as query parameters (as in definitions-1.0.0), can potentially put coordinates as query params: patch /definitions?coordinates=zzzz, request body contains the same curation body currently in the build (prior to change)  What do you think?

qtomlinson: After reading about REST API designs, the coordinates here are the identifiers for a specific definition resource, and therefore, belongs to the url.  Fix parsing coordinates in the path instead.  See Route paths https://expressjs.com/en/guide/routing.html

Issue #670, opened 2019-11-21
Inconsistent number of Cocoapods definitions in the stats and fetchable through the API


pombredanne:
Using the API I could only get 365 definitions from: https://api.clearlydefined.io/definitions?type=pod&provider=cocoapods
and then following the "continuationToken" to

- https://api.clearlydefined.io/definitions?type=pod&provider=cocoapods&continuationToken=cG9kL2NvY29hcG9kcy8tL3N3aWZ0aG9vay8yLjAuMA==
- https://api.clearlydefined.io/definitions?type=pod&provider=cocoapods&continuationToken=cG9kL2NvY29hcG9kcy8tL3l3ZWJ2aWV3LzAuMTIuMA==
- https://api.clearlydefined.io/definitions?type=pod&provider=cocoapods&continuationToken=cG9kL2NvY29hcG9kcy8tL3ppcHphcC84LjEuMQ==

This last URL returns an empty continuationToken meaning that's the end of pagination.
In contrast this page https://clearlydefined.io/stats lists 7847 pods and not only 365

pombredanne:
@jeffmcaffer FYI this is one of the bugs that impairs mirroring too

nellshamrell:
I was able to confirm this:

I ran an initial API call, then followed the continuation tokens like this:
```bash
curl  https://api.clearlydefined.io/definitions?type=pod&provider=cocoapods

curl  https://api.clearlydefined.io/definitions?type=pod&provider=cocoapods&continuationToken=cG9kL2NvY29hcG9kcy8tL29jdG9raXQvMC4xLjE=

curl https://api.clearlydefined.io/definitions?type=pod&provider=cocoapods&continuationToken=cG9kL2NvY29hcG9kcy8tL3VpdGFibGV2aWV3K3B1bGx0b3pvb21pbnRhYmxlaGVhZGVydmlldy8xLjAuMA==

curl https://api.clearlydefined.io/definitions?type=pod&provider=cocoapods&continuationToken=cG9kL2NvY29hcG9kcy8tL3lvdWJvcmFsaWIvNi4yLjM=

curl https://api.clearlydefined.io/definitions?type=pod&provider=cocoapods&continuationToken=cG9kL2NvY29hcG9kcy8tL3pncGFyYWxsZWx2aWV3LzAuMS42
```

I saved the json returned by each of those requests to a few files, then ran this script to parse them and count the components returned (not the most elegant thing I've ever written, but it worked).

```javascript
[...]

PR #965, opened 2023-01-20
Fixed incorrect continuation token when searching with sort

qtomlinson:
The continuation token was generated based on _mongo.partitionKey. This is correct when _mongo.partitionKey is the only sort field.  When there are other sort fields in the query, the sequence of records are influenced by those fields. Continuation token based on only _mongo.partitionKey value does not produce the correct results for next pages.

There is a npm package on "cursor based pagination": https://www.npmjs.com/package/mongo-cursor-pagination The library supports one "paginatedField".  The ClearlyDefined api supports sort resulting in multiple sort fields. For example: sort on namespace sorts on three fields: ['coordinates.namespace', 'coordinates.name', 'coordinates.revision'].  So the library is not sufficient for CD's need.

To generate the correct next page, the continuation token needs to record all the values for sort fields. As an extension of existing logic, _mongo.partitionKey will always be included to the sort fields as a tie breaker. Also added logic to generate pagination query based on multiple sort values.

Task: https://github.com/clearlydefined/service/issues/670

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,9 +1,1025 @@
 {
   "name": "service",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@aws-crypto/ie11-detection": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/ie11-detection/-/ie11-detection-3.0.0.tgz",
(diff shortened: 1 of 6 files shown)


qtomlinson:
@mpcen ready for review

mpcen (approved):

qtomlinson:
@mpcen This commit is for correctness only.  There may be a performance impact which may be addressed by adding combination indexes (e.g. described.releaseDate + _mongo.partitionKey) to the definition DB.

mpcen:
@qtomlinson i think the perf hit is fine as we're starting a push for perf improvements now.

qtomlinson:
@mpcen rebased and ready.

Issue #846, opened 2021-06-01
License file not detected after scanning


capfei:
Package license is is named correctly and it at the root, but scanning did not pick it.

License file not detected with only two scanners used: https://clearlydefined.io/definitions/nuget/nuget/-/ReportGenerator/4.0.5-rc3

This version had three scanners used and the license file was detected: https://clearlydefined.io/definitions/nuget/nuget/-/ReportGenerator/2.1.1

capfei:
https://clearlydefined.io/definitions/pod/cocoapods/-/FirebaseInstanceID/7.2.0

ariel11:
Partially harvested package missed "LICENSE" in "Files" section  - https://clearlydefined.io/definitions/pod/cocoapods/-/Flipper-Boost-iOSX/1.76.0.1.11

ariel11:
An example of all scanners being run but the LICENSE.html in "FIles" was not marked "NOASSERTION"  - https://clearlydefined.io/definitions/npm/npmjs/@ag-grid-enterprise/core/23.1.1

ariel11:
The scanners also missed at the file level (example:  server-side-row-model.cjs.js - https://clearlydefined.io/file/6af929081ba05472bf83e5d8fe13bdbc84f598ce261ea87a4775c10fcc7d4223) the reference to "@license Commercial."  ClearlyDefined marked that file with "MIT," which is accurate - there's references to MIT on that file.  However, I would have expected the tooling to put "MIT AND NOASSERTION" for that file given the @license Commercial reference.  Here's the definition page: https://clearlydefined.io/definitions/npm/npmjs/@ag-grid-enterprise/server-side-row-model/23.1.1.

ariel11:
Another example - on this one (https://clearlydefined.io/definitions/nuget/nuget/-/RequireJS/2.1.14), scanners did not pick up license info on the files (e.g. https://clearlydefined.io/file/7a10cbce6ec24bc6769d7a8de5f99620c1635dbead60712225404dde13d424f9).  That file should have said "NOASSERTION" because it has license info on the file.
![image](https://user-images.githubusercontent.com/17935946/121947110-87fffe00-cd0a-11eb-8fd6-9ba920923d4b.png)
Instead, it's just blank
![image](https://user-images.githubusercontent.com/17935946/121947163-977f4700-cd0a-11eb-8749-b56cd88a440f.png)

capfei:
[...]

PR #968, opened 2023-01-26
Fix NoAssertion for crate component

qtomlinson:
For crate components, the license declared by clearlydefined summarizer is the preferred choice.  See Commit "Fix handling of Rust crates with license choices".

In the problematic case of
https://clearlydefined.io/definitions/crate/cratesio/-/mpmc/0.1.6, scancode detected that the license is BSD-2-Clause-Views.  The component's license retrieved from the package repository is "non-standard". This is converted to NoAssertion in the clearlydefined summary, and later set to be the package's license.

The solution is to move the logic of overriding crate license information to aggregator and only override when the license declared by clearlydefined summary is not "NOASSERTION".

Task: https://github.com/clearlydefined/service/issues/846#issuecomment-1405725749

### DIFF
--- a/business/aggregator.js
+++ b/business/aggregator.js
@@ -32,6 +32,7 @@
 const { getLatestVersion, mergeDefinitions } = require('../lib/utils')
 const { flattenDeep, get, set, intersectionBy } = require('lodash')
 const logger = require('../providers/logging/logger')
+const { setIfValue } = require('../lib/utils')
 
 class AggregationService {
   constructor(options) {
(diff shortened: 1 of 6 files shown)


qtomlinson:
@mpcen

mpcen:
Do we know if this "non-standard" license occurs elsewhere? If so, it might be worth contributing the license to SPDX

qtomlinson:
"non-standard" does not show up in https://spdx.org/licenses/

mpcen (approved):

PR #974, opened 2023-02-01
Fix NoAssertion for some composer components

qtomlinson:
For composer components, files are packaged under one single sub directory in the root directory. The license file contained in the sub directory is not picked up.  Solve by traversing into the subdirectory to detect license file.

Task: https://github.com/clearlydefined/service/issues/846#issuecomment-1411265574


qtomlinson:
@mpcen ready for review

mpcen (approved):

PR #999, opened 2023-07-17
Fix ScanCodeSummarizer

qtomlinson:
Fix a few cases where Scancode results provide license information.
1. In addition to content.summary.packages[0].declared_license, check content.summary.packages[0].license_expression
2. Concatenate the license only when its score >= 80, to be consistent with summarizing license info for files.
3. Add META/INF directory as license location for maven packages.

See commit messages for detailed explanation and test cases


qtomlinson:
@mpcen ready for review

qtomlinson:
@mpcen @jeffwilcox rebased and ready for review.  Any feedback would be very much appreciated!

providers/summary/scancode.js
```diff
@@ -178,6 +189,10 @@ class ScanCodeSummarizer {
       .filter(e => e)
   }
 
+  _isMatchedLicense(license) {
+    return license.score >= 80
```
jeffwilcox, line 193: where is this threshold documented/defined for the community understanding - is that a scancode native value that means "definitely the match"?

jeffwilcox: [I read the more detailed commit messages]

Are there any situations where this >= 80 threshold might produce an inaccurate resulting license that we could point to a risk on this?

qtomlinson: Thanks for the feedback!
The exiting logic in scancode summarizer includes licenses with lower scores in the declared license, but lists only licenses scoring above 80 for the files.  For example, the declared licenses for https://clearlydefined.io/definitions/composer/packagist/colinmollenhour/cache-backend-redis/1.14.4 are BSD-3-Clause AND GPL-2.0-only.  However discovered license is only BSD-3-Clause (GPL-2.0-only is filtered out due to lower score).  This inconsistency causes confusion.  My attempt here is to make declared license and discovered license consistent and remove licenses scoring less than 80 from the declared license.

As to what is the appropriate score threshold to avoid risk of incorrect license:
- in versions prior to 3.0.0, licenses with score >= 90 are considered as declared license (`_getLicenseByFileName`)
- Since version 3.0.0, is_license_text flag from scancode is used.  The “is_license_text” flag is set to true for files that contain mostly license texts and notices.  Based on [documentation](https://github.com/nexB/scancode-toolkit/blob/6d4ea7f10cf0d92f1ce922ff46e85e2e903ff4e8/docs/source/cli-reference/scan-options-post.rst#--is-license-text-option_), "mostly means over 90% of the content of the file."
According to this documentation (prior to v32 update, 30.1.0 is currently used in the crawler), the license objects for each detected license in the files has "is_license_text" attribute as well.  I will adapt my solution to use the is_license_text flag on the detected license instead.

qtomlinson: Upon further research, there are four flags determined at each license match level: "is_license_text", "is_license_notice", "is_license_tag", "is_license_reference".   This is also the order of precedence.  Details see https://scancode-results-analyzer-fork-ayan.readthedocs.io/en/latest/how-analysis-is-performed/cases-incorrect-scans.html.  is_license_text alone is not enough to determine whether it is the correct license match.  Also see more discussion at  https://github.com/nexB/scancode-toolkit/issues/3082

I will refactor and reuse the logic for scancode prior to 3.0.0 to determine declared license by scoring >= 90.  This way is more consistent within CD's implementation.  @jeffwilcox What do you think?

qtomlinson: > Are there any situations where this >= 80 threshold might produce an inaccurate resulting license that we could point to a risk on this?

@jeffwilcox Without any fixes, there is currently a risk of inaccurate resulting license.  Details see [issue](https://github.com/clearlydefined/service/issues/1010).  File licenses (as discovered) are always filtered based on score (>= 80).  For declared licenses, there is a filtering based on score (>=90) in place for ScanCode data with versions prior to 3.0.0.  This filtering is missing for later versioned ScanCode data, resulting license matched with lower scores shows up in "declared", but filtered out in "discovered".  Including the score filtering (>=90) will ensure more accurate license information, and also consistency in handling different versions of ScanCode data.

qtomlinson:
@elrayle for review

pombredanne:
Feel free to pull me and @AyanSinhaMahapatra in. We maintain ScanCode.
Note that  filtering based on license score is likely to miss several detected licenses.

AyanSinhaMahapatra:
> where is this threshold documented/defined for the community understanding - is that a scancode native value that means "definitely the match"?

See https://scancode-toolkit.readthedocs.io/en/stable/reference/license-detection-reference.html for reference, we now have LicenseDetection objects (with tons of upgrades in license detection heuristics) which have a `detection_log` optionally which has info on license match accuracy. See https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/detection.py#L98 for possible values, there is some docs/docstrings but scattered, I'll improve on the docs here too. Note that this is similar to https://scancode-results-analyzer-fork-ayan.readthedocs.io/en/latest/how-analysis-is-performed/selecting-incorrect-unique.html#file-regions-with-incorrect-scans but has many improvements, as this was a prototype, which was improved a lot before being merged in scancode proper.

Also note that we do much better package and license summarization on a codebase-level https://scancode-toolkit.readthedocs.io/en/stable/cli-reference/scan-options-post.html#summary-option based on legalese/key files which would be useful also.

qtomlinson:
@pombredanne @AyanSinhaMahapatra It is great to have feedback from the experts!  Thanks for the insight and pointers on the new ScanCode data format.  The references are extremely helpful for our ScanCode upgrade.  I have included the related comments in the [upgrade issue](https://github.com/clearlydefined/service/issues/1006).

The current issue at hand deals with ScanCode output data from version 30.1.0. How to make declared license consistent with discovered licenses ([issue](https://github.com/clearlydefined/service/issues/1010))?

Options:
1. rescan the package when ScanCode is upgraded, or
2. see whether the harvested data has enough information where a license can be correctly derived without triggering a re-harvest.
Option #2 is the cheaper one, and is being explored here.

With the introduction of is_license_text on the file level, file.is_license_text is used in CD to detect a license file.  When there are multiple license matches for a license file, historically, licenses[i].score has been used as a filter (>=80).  For example, for colinmollenhour-Cm_Cache_Backend_Redis-6661e3f/LICENSE, there are two license matches: 
	- BSD-3-Clause (score 95.71) and 
	- GPL 2.0 (score 11.72).  
After filtering based on score, only BSD-3-Clause is displayed as the license for the file, and later in discovered licenses, in clearlydefined UI, 

For the declared license, the filtering based on score is missing for recent versions of ScanCode data, resulting both BSD-3-Clause and GPL 2.0 are declared but only BSD-3-Clause is discovered.  This is confusing to end users.

Without resorting to filtering based on a score threshold, using licenses[i].matched_rule[j].is_license_text was also explored.  It is found that in some cases, when licenseFile.is_license_text === true, licenses[i].matched_rule[j].is_license_text can be false for all matched licenses (is_license_notice matched, see test/fixtures/scancode/3.0.2/commons-clause.json).  Using licenses[i].matched_rule[j].is_license_notice is not good either.  For colinmollenhour-Cm_Cache_Backend_Redis-6661e3f/LICENSE, GPL 2.0 is a match by is_license_notice === true with score of 11.72.  So it seems that score is a consistent measure of license match in data versioned <= 30.1.0.  That is why filtering based on score >=90 for declared license is proposed, with filtering score >=80 as discovered license already in place.  

Any alternatives?

qtomlinson:
> Note that filtering based on license score is likely to miss several detected licenses.

Yes. Filtering (for consistency) does increase the likelihood of NOASSERTION here.  Those cases need curation or re-harvested using the updated ScanCode.

providers/summary/scancode.js
```diff
@@ -33,8 +34,11 @@ class ScanCodeSummarizer {
     const result = {}
     this.addDescribedInfo(result, harvested)
     let declaredLicense = this._readDeclaredLicense(scancodeVersion, harvested)
-    if (!declaredLicense || declaredLicense === 'NOASSERTION') {
-      declaredLicense = this._getDeclaredLicense(scancodeVersion, harvested, coordinates)
+    if (!isDeclaredLicense(declaredLicense)) {
+      declaredLicense = this._readLicenseExpression(harvested) || declaredLicense
```
qtomlinson, line 38: In the latest ScanCode format, "the codebase level packages has been also updated: license_expression -> declared_license_expression, also with it’s SPDX version, declared_license -> extracted_license_statement" from [doc](https://scancode-toolkit.readthedocs.io/en/stable/reference/license-detection-reference.html#chnagelog-summary).  Here in CD with 30.1.0 ScanCode data, declared_license is used to parse package level license information.  In addition, package level license_expression (->declared_license_expression in the new format) can be utilized to provide license information.  Let me know if you have any feedback.

PR #1002, opened 2023-08-18
Pin exact dependencies

JamieMagee:
For applications it's considered a best practice to pin exact versions of dependencies.

This change removes the semantic version range specifiers and pins each dependency to the maximum allowed by the range specifier. This makes the implicit updates explicit. For example:

- `^1.0.0` will have been upgraded to the maximum **minor** version available. Equivalent to `1.0.0 <= x < 2.0.0`
- `~1.0.0` will have been upgraded to the maximum **patch** version available. Equivalent to `1.0.0 <= x < 1.1.0`


jeffwilcox (approved):

JamieMagee:
Looks like there might be some breaking changes, despite the fact that npm packages are supposed to follow semantic versioning.

I'll break this down into smaller changes and merge it piece-by-piece instead.

Issue #957, opened 2022-11-07
Query with a well-formed ID fails


waynebeaton:
I encountered a handful of `pypi` libraries today for which the API throws a 500 error without explanation.

For example, all of these:

```
curl -X GET "https://api.clearlydefined.io/definitions/pypi/pypi/-/slip.dbus/0.6.4" -H  "accept: */*"
curl -X GET "https://api.clearlydefined.io/definitions/pypi/pypi/-/python-dmidecode/3.12.2" -H  "accept: */*"
curl -X GET "https://api.clearlydefined.io/definitions/pypi/pypi/-/blivet-gui/2.3.0" -H  "accept: */*"
```

Answer this result:
```
{"error":{"code":"500","message":"An error has occurred","innererror":{}}}
```

My expectation is that I should get something back.

For completeness, I'm actually using the API to submit many IDs at the same time, and -- if any one of them is bad (for whatever reason) -- I get a 500 on the entire lot. I have an algorithm that does a divide and conquer to grab data for the valid ones and identify the problematic ones, but that runs the risk of butting up against the API limit which is the entire reason why I'm trying to batch things.

For example, two of the entries in the following query are valid and return valid results when the third (`slip-dbus`) is removed.

```
curl -X POST "https://api.clearlydefined.io/definitions" -H  "accept: application/json" -H  "Content-Type: application/json" -d "[\"pypi/pypi/-/sos/4.3\",\"pypi/pypi/-/slip.dbus/0.6.4\",\"pypi/pypi/-/spdx-tools/0.7.0a3\"]"
```

lumaxis:
Confirmed that this is fixed by #1016 in our staging environment – fix should be rolled out soon!

PR #1016, opened 2023-11-08
Handle NotFound when mapping pypi coordinates

qtomlinson:
Task: https://github.com/clearlydefined/service/issues/957


qtomlinson:
@jeffwilcox @JamieMagee for review

PR #1021, opened 2024-01-12
add github action to build and deploy the dev CD/service

elrayle:
* Partially addresses #1026 

### Description

This PR adds a GitHub action to perform the deploy of service to the dev service app in Azure.  To support debugging, it also adds the sha of the current deploy to the status endpoint.

_NOTE: As part of this work, Azure auth credentials were created following this [documentation](https://github.com/Azure/login#login-with-a-service-principal-secret)._

### Follow-on Work

* create a bot user (e.g. clearlydefinedopts-bot and update credentials to use this bot
* permanently disable Azure DevOpts pipeline for dev service
* setup prod service to use a GH action to complete Issue #1026
* #1027

### DIFF
--- a/.github/workflows/build_and_deploy_dev.yml
+++ b/.github/workflows/build_and_deploy_dev.yml
@@ -70,6 +70,11 @@ jobs:
                     "name": "DOCKER_REGISTRY_SERVER_PASSWORD",
                     "value": "${{ secrets.REGISTRY_PASSWORD }}",
                     "slotSetting": false
+                },
+                {
+                    "name": "BUILD_SHA",
+                    "value": "${{ github.sha }}",
+                    "slotSetting": false
                 }
             ]
 
--- a/.github/workflows/build_and_deploy_dev.yml
+++ b/.github/workflows/build_and_deploy_dev.yml
@@ -7,8 +7,7 @@ name: Build and Deploy to dev service app
 on:
   workflow_dispatch:
   push:
-    branches: [elr/gh-action-dev-deploy]
-    # branches: [master]
+    branches: [master]
 
 # There are secrets and environment variables that need to be set that control what is pushed to
 # ghcr and Azure.
--- a/.github/workflows/build_and_deploy_dev.yml
+++ b/.github/workflows/build_and_deploy_dev.yml
@@ -19,7 +19,7 @@ on:
 # Environment Variables:
 #   APPLICATION_TYPE:   type of application that is being deployed; used to add a label to the Docker image (values: api | web | worker)
 #   AZURE_WEBAPP_NAME:  name of the Azure WebApp being deployed
-#   CODE_ENVIRONMENT:   environment that the code is being deployed to; used to add a label to the Docker image (values: dev | prod)
+#   DEPLOY_ENVIRONMENT:   environment that the code is being deployed to; used to add a label to the Docker image (values: dev | prod)
 #   DEPLOY_DOCKER_TAG:  the tag used for deploying a specific Docker image to Azure.  For dev, use the `github.sha`.  For production, use the SEMVER
 #                       version of the release.  Make sure to add this tag to the `DOCKER_TAGS` in the `Build and push Docker image` step.
 #   DOCKER_IMAGE_NAME:  name of the Docker image that is being built and pushed to ghcr.io.
@@ -27,7 +27,7 @@ on:
 env:
   APPLICATION_TYPE: api
   AZURE_WEBAPP_NAME: clearlydefined-api-dev
-  CODE_ENVIRONMENT: dev
+  DEPLOY_ENVIRONMENT: dev
   DEPLOY_DOCKER_TAG: ${{ github.sha }}
   DOCKER_IMAGE_NAME: ghcr.io/${{ github.repository_owner }}/${{ github.repository }}-dev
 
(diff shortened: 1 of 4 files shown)


routes/index.js
```diff
@@ -4,7 +4,11 @@ const express = require('express')
 const router = express.Router()
 
 router.get('/', function(req, res) {
-  res.status(200).send({ status: 'OK' })
+  const sha = require('child_process')
+    .execSync('git rev-parse HEAD')
+    .toString().trim()
+  const msg = `{ "status": "OK", "sha": "${sha}" }`
```
qtomlinson, line 10: execSync can throw, so probably need to wrap in a try-catch block

this endpoint can potentially be the k8s health check. In Azure app service, there is also a similar concept. In the service log, this seemed to be called every 5 minutes (likely as health check). Retrieving the commit hash every time may not be performant.  Is it possible to cache this hash or store this commit hash in an environment variable?

qtomlinson (commented):
Nice work! Really appreciated the background information in the PR.

routes/index.js
elrayle: Updated the deploy to pass the sha as a config.  It now gets the sha once during setup from the configs.

elrayle:
@qtomlinson Thanks for the review.  I've updated the code in the last two commits to address your comments.

.github/workflows/build_and_deploy_dev.yml
```diff
@@ -0,0 +1,86 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }} # token created when the action launched (short lived)
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          push: true
+          file: Dockerfile
+          tags: |
+            ${{ env.DOCKER_IMAGE_NAME }}:latest
```
RomanIakovlev, line 37: I'm a bit on the fence regarding whether we need 'latest' tag, not sure what good use it brings. At least for deployment I think it is an anti-pattern, see my further comment on this below.

.github/workflows/build_and_deploy_dev.yml
```diff
@@ -0,0 +1,86 @@
...
+
+      - name: Set DOCKER configs in Azure web app
+        uses: Azure/appservice-settings@v1
+        with:
+          app-name: ${{ env.AZURE_WEBAPP_NAME }}
+          app-settings-json: |
+            [
+              {
+                    "name": "DOCKER_CUSTOM_IMAGE_NAME",
+                    "value": "${{ env.DOCKER_IMAGE_NAME }}:latest }}",
```
RomanIakovlev, line 56: I think it is problematic to use the 'latest' tag for deployment. If we'll have to debug the deployment, it will not be immediately clear which exactly tag was 'latest' during the deployment time. I think it's better to use a fixed image version here, which you luckily have access to, because you use git sha sum for it.

elrayle: Thanks for the feedback.  What you are saying makes sense.  I will remove `:latest` tag and only use the sha for development.  For production, the plan was already to only use the semver version.

.github/workflows/build_and_deploy_dev.yml
```diff
@@ -0,0 +1,105 @@
...
+# ghcr and Azure.
+#
+# Secrets:
+#   AZURE_CREDENTIALS:                service principal that has access to the Azure WebApp
+#   AZURE_WEBAPP_PUBLISH_PROFILE_DEV: publish profile for the Azure WebApp  NOTE: The name of the secret changes.  For dev, it ends in `_DEV`.  Production does not have an extension.
+#
+# Environment Variables:
+#   APPLICATION_TYPE:   type of application that is being deployed; used to add a label to the Docker image (values: api | web | worker)
+#   AZURE_WEBAPP_NAME:  name of the Azure WebApp being deployed
+#   CODE_ENVIRONMENT:   environment that the code is being deployed to; used to add a label to the Docker image (values: dev | prod)
```
qtomlinson, line 22: DEPLOY_ENVIRONMENT? or DEPLOYMENT_ENVIRONMENT?

.github/workflows/build_and_deploy_dev.yml
```diff
@@ -0,0 +1,105 @@
...
+        uses: docker/login-action@v3.0.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }} # user that kicked off the action
+          password: ${{ secrets.GITHUB_TOKEN }} # token created when the action launched (short lived)
+
+      - name: Build and push Docker image
+        env:
+          DOCKER_TAGS: |
+            ${{ env.DOCKER_IMAGE_NAME }}:${{ github.sha }}
```
qtomlinson, line 51: probably use EPLOY_DOCKER_TAG here.  For production, this would be semver.

qtomlinson (approved):
Great documentation!  Minor fixes see comments.

.github/workflows/build_and_deploy_dev.yml
elrayle: I've been going back and forth on whether in production, there should be sha and semver tags or just semver tag.  It doesn't have to be resolved for this PR, but I'd like to get opinions on this.  Theoretically, if you have the semver, you can get the sha from the release tag in GitHub.  If we choose to only have a single tag for both dev and prod, it simplifies the action script.

elrayle:
I've been going back and forth on whether in production, there should be sha and semver tags or just semver tag. It doesn't have to be resolved for this PR, but I'd like to get opinions on this. Theoretically, if you have the semver, you can get the sha from the release tag in GitHub. If we choose to only have a single tag for both dev and prod, it simplifies the action.

PR #1022, opened 2024-01-17
Fix previously broken test

lumaxis:
Instead of triggering the code path at https://github.com/clearlydefined/service/blob/4ae722e9eaad2bada374df687c0a5dad9e1efe18/providers/summary/scancode.js#L82-L83, this test previously just tested its own `throw new Error` line. This is now fixed and the expected error message and fixture adjusted to make the test actually work.


elrayle (approved):
Nice to see this move toward consistency in case.  And great fix on the test.

PR #1030, opened 2024-02-02
add GitHub action for deploy of production service

elrayle:
This action is very similar to the [dev action](https://github.com/clearlydefined/service/blob/master/.github/workflows/build_and_deploy_dev.yml).  The main differences are:

* deploy on release
* deploy to US prod and EU prod

### DIFF
--- a/.github/workflows/build_and_deploy_prod.yml
+++ b/.github/workflows/build_and_deploy_prod.yml
@@ -1,0 +1,149 @@
+# This workflow will build a docker image, push it to ghcr.io, and deploy it to an Azure WebApp.
+name: Build and Deploy to prod service app
+
+# Update the triggers based on the environment that is being deployed to.
+#   Triggers for dev deployments:  1) manually triggered, 2) push to branch `master`
+#   Triggers for prod deployments: 1) manually triggered, 2) release [published]
+on:
+  workflow_dispatch:
+  release:
+    types: [published]
+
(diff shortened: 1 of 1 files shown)


qtomlinson (approved):

.github/workflows/build_and_deploy_prod.yml
```diff
@@ -0,0 +1,149 @@
...
+                    "slotSetting": false
+                },
+                {
+                    "name": "DOCKER_REGISTRY_SERVER_PASSWORD",
+                    "value": "${{ secrets.REGISTRY_PASSWORD }}",
+                    "slotSetting": false
+                },
+                {
+                    "name": "BUILD_SHA",
+                    "value": "${{ github.sha }}",
```
qtomlinson, line 98: May want to consider passing in env.DEPLOY_DOCKER_TAG, that is the semVer for production and github.sha for dev.   Same on line 138.

elrayle: We can pass in the version in addition to the sha.  I want the sha always available because it is the source of truth about what code has been deployed.

qtomlinson: sounds good!

elrayle:
Closing.  The action is being merged into prod instead of master.  It will be back ported to master.

PR #1043, opened 2024-02-09
deprecate config TRIMMED_DEFINITION_MONGO_COLLECTION_NAME

elrayle:
Deprecated config: TRIMMED_DEFINITION_MONGO_COLLECTION_NAME
Replacement config: DEFINITION_MONGO_TRIMMED_COLLECTION_NAME

Config TRIMMED_DEFINITION_MONGO_COLLECTION_NAME was removed in PR #987.  To avoid a major version change in the next release, the config is being added back in and deprecated.

New processing:
* if the old config has a value, a deprecation message is written to the console
* if the new config is specified, its value is used and the old one is ignored if it exists
* if only the old config is specified, its value is used
* otherwise, if neither are specified, the default is used

### DIFF
--- a/providers/stores/mongoConfig.js
+++ b/providers/stores/mongoConfig.js
@@ -17,9 +17,13 @@ function definitionPaged(options) {
 }
 
 function definitionTrimmed(options) {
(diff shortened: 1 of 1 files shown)


elrayle:
This should have been set to merge into prod branch.

PR #1044, opened 2024-02-09
prepare for release v1.1.0 by updating the version

elrayle:
* MERGE AFTER PR #1046

Normally, setting the version happens in master just before it is merged into prod in preparation of a release.  Since the code to be released is already in prod branch and master branch has newer commits, this needs to be added directly to prod.  Once merged into prod branch, it will be backported to master.

Used `npm version 1.1.0` to generate.  ([documentation](https://docs.npmjs.com/cli/v8/commands/npm-version)]

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "service",
-  "version": "0.1.1",
+  "version": "1.1.0",
   "description": "Service side of clearlydefined.io.",
   "scripts": {
     "test": "nyc mocha --exit \"test/**/*.js\" && eslint .",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "service",
-  "version": "0.1.1",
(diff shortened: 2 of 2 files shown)


qtomlinson (commented):
Need to update package-lock.json to be in sync.  Can use [npm version](https://docs.npmjs.com/cli/v8/commands/npm-version) to do it as well.

qtomlinson (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "service",
-  "version": "0.1.1",
+  "version": "1.1.0",
```
qtomlinson, line 3: Need to update package-lock.json to be in sync. Can use [npm version](https://docs.npmjs.com/cli/v8/commands/npm-version) to do it.

elrayle: Thanks.  That worked great.

PR #1045, opened 2024-02-09
deprecate config TRIMMED_DEFINITION_MONGO_COLLECTION_NAME

elrayle:
**NOTE: This is merging into prod branch to allow release of v1.1.0 as a minor release.**. It will be backported to master branch.

----

Deprecated config: TRIMMED_DEFINITION_MONGO_COLLECTION_NAME
Replacement config: DEFINITION_MONGO_TRIMMED_COLLECTION_NAME

Config TRIMMED_DEFINITION_MONGO_COLLECTION_NAME was removed in PR https://github.com/clearlydefined/service/pull/987. To avoid a major version change in the next release, the config is being added back in and deprecated.

New processing:

- if the old config has a value, a deprecation message is written to the console
- if the new config is specified, its value is used and the old one is ignored if it exists
- if only the old config is specified, its value is used
- otherwise, if neither are specified, the default is used


elrayle:
DO NOT MERGE The branch was created from master, so it can't be merged into prod.

PR #1065, opened 2024-02-29
Restrict PyPI coordinates name in PypiCoordinatesMapper

qtomlinson:
In PyPI, '/' is invalid in a package name (see https://peps.python.org/pep-0508/#names).  Add a check to restrict the coordinates name prior to querying the PyPI API.

### DIFF
--- a/lib/pypiCoordinatesMapper.js
+++ b/lib/pypiCoordinatesMapper.js
@@ -16,16 +16,15 @@ class PypiCoordinatesMapper {
   }
 
   _shouldResolve(coordinates) {
-    return coordinates.name.includes('.') ||
-      coordinates.name.includes('_') ||
-      coordinates.name.includes('-')
+    if (typeof coordinates.name !== 'string' || coordinates.name.includes('/')) return false
+    return coordinates.name.includes('.') || coordinates.name.includes('_') || coordinates.name.includes('-')
   }
 
   async _resolve(coordinates) {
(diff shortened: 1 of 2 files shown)


lib/pypiCoordinatesMapper.js
```diff
@@ -16,16 +16,15 @@ class PypiCoordinatesMapper {
   }
 
   _shouldResolve(coordinates) {
-    return coordinates.name.includes('.') ||
-      coordinates.name.includes('_') ||
-      coordinates.name.includes('-')
+    if (typeof coordinates.name !== 'string' || coordinates.name.includes('/')) return false
+    return coordinates.name.includes('.') || coordinates.name.includes('_') || coordinates.name.includes('-')
```
lumaxis, line 20: How would you like something like this?

```suggestion
    if (typeof coordinates.name !== 'string' || coordinates.name.includes('/')) {
        return false;
    }

    const validSeparatorCharacters = ['.', '_', '-'];
    return validSeparatorCharacters.some(char => coordinates.name.includes(char));
}
```

lumaxis (approved):
Had one suggestion but looks good! Thanks especially for adding tests 🙂

lib/pypiCoordinatesMapper.js
qtomlinson: Tried using `equivalentSeparators.some(char => coordinates.name.includes(char))`, but the code scanner does not like it.  Will leave this as is.

PR #1066, opened 2024-02-29
Use URL object to generate request URL

lumaxis:
This is a test to check if this satisfies a GitHub code scanning warning.


qtomlinson (approved):
Thank you!

PR #1072, opened 2024-03-11
Use dependency injection in PypiCoordinatesMapper

qtomlinson:
This is to fix the CodeQL scanning failure and make stubbing network call easier in unit tests.

### DIFF
--- a/lib/pypiCoordinatesMapper.js
+++ b/lib/pypiCoordinatesMapper.js
@@ -5,8 +5,9 @@ const requestPromise = require('request-promise-native')
 const EntityCoordinates = require('./entityCoordinates')
 
 class PypiCoordinatesMapper {
-  constructor() {
+  constructor(fetch = requestPromise) {
     this.baseUrl = 'https://pypi.python.org'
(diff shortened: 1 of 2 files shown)


lib/pypiCoordinatesMapper.js
```diff
@@ -21,19 +22,16 @@ class PypiCoordinatesMapper {
   }
 
   async _resolve(coordinates) {
+    if (coordinates.name === '..') return null
```
lumaxis, line 25: I was wondering if we want a more generalized function to sanitize coordinates? But also fine as is 👍🏼

lumaxis (approved):

lib/pypiCoordinatesMapper.js
qtomlinson: I thought about this as well: using URL object and the check `/../` in the final url can be refactored into a function.  I will do the refactoring as a separate PR.  The check for `/` in coordinates.name is specific to coordinates, which maybe harder to refactor.

Issue #1084, opened 2024-04-03
Precedence is not preserved when joining Scancode license expressions


qtomlinson:
### Expected:
When joining license expressions with AND: 
- 'MIT OR Apache-2.0', 
- 'GPL'.  

Precedence should be preserved in the result.  The joined expression should be: `GPL AND (MIT OR Apache-2.0)`

### Observed:
Precedence is not preserved in the result.  The observed result is `GPL AND MIT OR Apache-2.0`

### Reproducing Steps:
- In test case "creates expressions from license expressions" @[service/test/summary/scancode.js](https://github.com/clearlydefined/service/blob/89555eb2a172d4c804d1f4541377e300de77ca63/test/summary/scancode.js#L201), given `new Set(['MIT OR Apache-2.0', 'GPL'])`, the expected joined expression should be `GPL AND (MIT OR Apache-2.0)`

- Another test case is at "summarizes and ignore license_expression containing NOASSERTION in version 30.1.0 of ScanCode" @ [service/test/providers/summary/scancode.js](https://github.com/clearlydefined/service/blob/89555eb2a172d4c804d1f4541377e300de77ca63/test/providers/summary/scancode.js#L110), the expected `result.licensed.declared` should be `BSD-2-Clause AND (CC0-1.0 OR BSD-2-Clause)`.  

This relates to https://github.com/clearlydefined/service/pull/1056#discussion_r1527021349

PR #1087, opened 2024-04-04
Update _joinExpressions to preserve precedence

lumaxis:
Fixes https://github.com/clearlydefined/service/issues/1084


qtomlinson (commented):
Yes.  This would be my first thought of fixing this as well.  But looking at the test cases of stringify in service/test/lib/spdx.js, I have 2nd thought: the license expressions joined this way may not be the simplified form. stringify in the SPDX module has the ability to add and omit () based on precedence (e.g, MIT AND BSD-3-Clause WITH GCC-exception-3.1 OR CC-BY-4.0 AND Apache-2.0).  Options to leverage this may include: 
1. avoid calling stringify the expression nodes in _createExpressionFromLicense, construct an AND expression node when joining all expression nodes, and finally use SPDX.stringify to serialize the AND expression node.  This however means that the internals of expression node need to be used to construct AND node, which is not ideal
2. Alternatively, in _joinExpressions, can add () to each license expression unconditionally, use AND to join them, and then return SPDX.normalize(joinedExpressionString).   SPDX.normalize calls stringify internally, so the expression after normalization is the simplified form without unnecessary ().  This approach is less efficient in that normalize parses the expression string again before stringify. But the advantage is that it is simple and delegates the knowledge of operators, precedence, and simplifying () to @clearlydefined/spdx.

Any thoughts?

lumaxis:
As we discussed last week, I agree and would definitely prefer to let the library handle this logic. Actually, my preference would be to either figure out if `@clearlydefined/spdx` can already do this through `merge`/`flatten`/`expand` somehow or alternatively, make changes to the library so that it can handle joining expressions for us.

lumaxis:
@qtomlinson Do you think you might have time to carry this forward while I'm on vacation? I think that would be tremendously helpful! 🙏🏼

qtomlinson:
@lumaxis sounds good.  I have updated the PR based on what we discussed during the dev sync meeting.  Please review and provide feedback.  I can also rebase this PR after the ScanCode upgrade PRs are merged.

elrayle (approved):
LGTM... are there any outstanding questions from previous discussions?  If not, let's ship it.  🚢

PR #1091, opened 2024-04-10
fix: deploy falsely reports SUCCESS when it failed

elrayle:
Multiple issues are fixed in this PR:
* If the webapp publish profile secret is empty, webapps-deploy v3 doesn’t deploy but reports the deploy as passing.  See Azure/webapps-deploy [Issue #404](https://github.com/Azure/webapps-deploy/issues/404).
* Configs were set before running the deploy.  This means that the config values in Azure are updated even if the deploy fails.
* The action was runnable by anyone with write access.  That is too broad for production.

To avoid these known issues, the PR updates to...
* check that all required secrets are set before proceeding
* only update configs if the deploy passes
* call the reusable workflow that checks if the user has access to deploy

### DIFF
--- a/.github/workflows/build_and_deploy_prod.yml
+++ b/.github/workflows/build_and_deploy_prod.yml
@@ -12,10 +12,12 @@ on:
 # There are secrets and environment variables that need to be set that control what is pushed to
 # ghcr and Azure.
 #
+# Org Secrets:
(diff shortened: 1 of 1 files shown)


elrayle:
This is superseded by the work in:

* #1092 
* https://github.com/clearlydefined/operations/pull/71

The new PRs combines the dev and deploy workflows into a single shared workflow in operations.

Issue #1103, opened 2024-04-22
Missing source location for sourcearchive packages


qtomlinson:
For a git source package, e.g. https://clearlydefined.io/definitions/git/github/ratatui-org/ratatui/bcf43688ec4a13825307aef88f3cdcd007b32641, the source location is populated.

In comparison, for a sourcearchive package, 
https://clearlydefined.io/definitions/sourcearchive/mavencentral/com.azure/azure-storage-blob/12.20.0
the source location is not populated.

The behaviors between these two types of source packages are inconsistent.

PR #1108, opened 2024-04-24
Add source location in definitions for sourcearchive packages

qtomlinson:
- definitionServiceTest.js is getting big. Separate the aggregator related tests to aggregatorTests.js.
- For a package containing source, there is existing logic to build source location in its definition.  Add the source location for sourcearchive packages as well.
- Add unit tests for all cases of building source location

Task: https://github.com/clearlydefined/service/issues/1103


lumaxis (approved):
Nice! Clean and easy 🙂

qtomlinson:
@lumaxis @elrayle This fixes future definition results.  How about for the previously computed definitions that are currently in storage?  We can discuss this at our dev sync meeting this week.

PR #1172, opened 2024-08-12
Fixed origins api for pypi components throwing 500 error when invalid package is provided

yashkohli88:
Below are the tasks completed as a part of this fix -

- [x]  Code updated to introduce try-catch block which checks if the error status code is 404, it returns an empty response. It throws error in other scenarios.
- [x] A separate function has been created to refactor the existing code which contained repeated set of statements, and to perform unit testing on the changes.
- [x] encodeURIComponent has also been introduced where we are creating URL to prevent injection.
- [x]  New Test cases added for happy path and two error scenarios with status code 404 and 400 to check the response. 
- [x]  Fixture added for the response of the positive testcase.

List of testcases added in the origins.js file related to the changes -

- Return proper response when valid package is provided as input
- Return blank response when invalid package is provided as input
- Throws error response when error other than invalid package (status 404) is returned from package manager response.

This fix is related to [Issue 1182](https://github.com/clearlydefined/service/issues/1182)


elrayle (approved):
Nice to see 500 errors being addressed.  Thanks!

elrayle:
@yashkohli88 The PR is marked as a work in progress.  Is this ready for review?  _NOTE: There are some conflicts to resolve._

Issue #1167, opened 2024-07-31
Origins API for Maven throwing 500 Error


yashkohli88:
#### Expected Behavior:
The origins API for Maven should return a response with a 404 error code and a proper 'Not Found' error message when the search string is not found. 

#### Observed:
The origins API for Maven is currently throwing an unhandled error with status code 500 when the search string is not found.

![image](https://github.com/user-attachments/assets/65dde000-5f85-4d8c-9c9a-c55bdda59f47)

#### Steps to Reproduce:
1. Hit the endpoint `https://dev-api.clearlydefined.io/origins/maven/124` with a GET request where "124" is the string being searched.
2. Observe the response, error code and error message.

#### Additional Information: 
The statement on [line 53 of the originMaven.js](https://github.com/clearlydefined/service/blob/c9419ab13d4e821da878404ecb83ad6ab0a71430/routes/originMaven.js#L53) file should check if the `spellcheck.suggestions `exist. As response from the URL https://search.maven.org/solrsearch/select?q=124&rows=100&wt=json does not contain suggestions section when the search string results are not found.

PR #1176, opened 2024-08-14
Fixed origins api for maven components throwing 500 error when invalid group id is provided

yashkohli88:
Below are the tasks completed as a part of this fix -

- [x]  Code updated to check if the value exist and test case added to verify the changes.
- [x]  New Test cases added for happy paths and all the error scenarios.
- [x]  Fixtures added for the input and one of the response.
- [x] getResponse function reads the fixture having response from [maven repo](https://search.maven.org/solrsearch/select?q=123&rows=100&wt=json) for each query (group id or artefact id) converts it to json object and then it is passed to getSuggestions function from originMaven.js file.

List of testcases added in the origins.js file related to the changes -

- Return suggestions when incomplete group id is provided as input
- Return list of artefacts when complete group id is provided as input
- Return blank response when group id is invalid and suggestions are not present
- Return blank response when group id and artefact id are invalid and suggestions are not present

This fix is related to [Issue 1167](https://github.com/clearlydefined/service/issues/1167)

### DIFF
--- a/routes/originMaven.js
+++ b/routes/originMaven.js
@@ -50,12 +50,13 @@ function getSuggestions(answer, group) {
     return docs.map(item => {
       return { id: item.id }
     })
-  const suggestions = answer.spellcheck.suggestions[1]
+  const suggestions = answer.spellcheck?.suggestions?.[1]
   const result = suggestions ? suggestions.suggestion : []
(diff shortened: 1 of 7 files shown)


elrayle (approved):
LGTM!  Thanks for getting in all the thorough testing.

PR #1201, opened 2024-10-15
Ljones140/reduce logging

ljones140:
# What

- Reduces high amount of logging to App Insights by reducing our busiest logs to debug level.
- Log level will default to `info` can be changed using new env var `APPINSIGHTS_EXPORT_LOG_LEVEL`
- Can be changed via App Service config without a code change

Logging library docs https://github.com/bragma/winston-azure-application-insights/blob/cfffb48e8761072a6d8aa0e9a122b46f42d26d49/README.md

# Why

App Insights is our biggest cost

Kusto query for working out busiest logs
```
traces
| summarize messageCount=count() by bin(timestamp, 1h), message
| order by messageCount
```

# Testing

Tested in dev setting `APPINSIGHTS_EXPORT_LOG_LEVEL` to `debug` and logs appeared
Set to `info` and the `debug` level logs did not appear

### DIFF
--- a/business/definitionService.js
+++ b/business/definitionService.js
@@ -70,6 +70,7 @@ class DefinitionService {
     const existing = await this._cacheExistingAside(coordinates, force)
     let result
     if (get(existing, '_meta.schemaVersion') === currentSchema) {
+      // Log line used for /status page insights
       this.logger.info('computed definition available', { coordinates: coordinates.toString() })
       result = existing
     } else result = force ? await this.computeAndStore(coordinates) : await this.computeStoreAndCurate(coordinates)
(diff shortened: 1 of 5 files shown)


providers/logging/winstonConfig.js
```diff
@@ -11,7 +11,7 @@ function factory(options) {
   const realOptions = options || {
     key: config.get('APPINSIGHTS_INSTRUMENTATIONKEY'),
     echo: false,
-    level: 'info'
+    level: config.get('APPINSIGHTS_EXPORT_LOG_LEVEL') || 'info'
```
elrayle, line 14: Nice addition to be able to set the log level from the configs.  Quite handy.  ❤️

business/definitionService.js
```diff
@@ -70,6 +70,7 @@ class DefinitionService {
     const existing = await this._cacheExistingAside(coordinates, force)
     let result
     if (get(existing, '_meta.schemaVersion') === currentSchema) {
+      // Log line used for /status page insights
```
elrayle, line 73: Excellent adding the comments about which log lines are used on the `/status` page.  That is handy information to have.

elrayle (approved):
Thanks for reducing logging and being a good steward of our resources.  I appreciate that you included the query in the description.  Nice to know how to identify costly logging if we need to evaluate again in the future.

providers/logging/winstonConfig.js
ljones140: Thanks, I will add to the infrastructure docs

PR #1203, opened 2024-10-16
Update to @clearlydefined/spdx v0.1.8

qtomlinson:
- Update tests for v0.1.8, to reflect https://github.com/clearlydefined/spdx/pull/30
- Also added more tests
- This is a interim step before moving to use GitHub packages.

### DIFF
--- a/test/lib/spdx.js
+++ b/test/lib/spdx.js
@@ -82,7 +82,7 @@ describe('SPDX utility functions', () => {
             right: { left: { license: 'BSD-3-Clause' }, conjunction: 'and', right: { license: 'Unlicense' } }
           }
         },
-        'MIT OR (BSD-2-Clause OR BSD-3-Clause AND Unlicense)'
+        'MIT OR BSD-2-Clause OR BSD-3-Clause AND Unlicense'
       ],
       [
         {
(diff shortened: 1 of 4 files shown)


qtomlinson:
@ljones140 I [experimented](https://github.com/qtomlinson/service/commit/616a1352670b3435fb3284551dfb28c1aeacdfb7) with using GitHub packages and found that an environment variable GITHUB_TOKEN may be required for authentication to install GitHub packages.  See [documentation](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-npm-registry).  Referring to GitHub repo with tag for dependency is an interim step to consume the changes.

ljones140:
> @ljones140 I [experimented](https://github.com/qtomlinson/service/commit/616a1352670b3435fb3284551dfb28c1aeacdfb7) with using GitHub packages and found that an environment variable GITHUB_TOKEN may be required for authentication to install GitHub packages. See [documentation](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-npm-registry). Referring to GitHub repo with tag for dependency is an interim step to consume the changes.

Hey @qtomlinson 

I don't think we will need a `GITHUB_TOKEN` as this package is public, so we don't need authn to read.

We do need to tell NPM to use GitHub as the package repository for `@clearlydefined` via the `.npmrc`.
`@clearlydefined:registry=https://npm.pkg.github.com`

I've branched off this and made a commit here which does this https://github.com/clearlydefined/service/commit/43092e915876a708b72de6942170e67adeac9109

test/lib/spdx.js
```diff
@@ -217,13 +217,17 @@ describe('SPDX utility functions', () => {
       'NOASSERTION': 'NOASSERTION',
       'See license': 'NOASSERTION',
       'MIT OR Apache-2.0': 'MIT OR Apache-2.0',
+      'MIT or Apache-2.0': 'NOASSERTION',
```
elrayle, line 220: Is this what we expect or want?  Shouldn't operators be case insensitive?

elrayle (approved):
There is one new test that I am surprised maps to NOASSERTION simply because of the case of the operator.  This is not a blocker for this PR since it is only showing how the code actually works now.  But I'd be inclined to open an issue requesting case-insensitivity for operators.

test/lib/spdx.js
qtomlinson: According to SPDX specification, operator has to be upper cased, whereas, the license identifiers are case insensitive.  So our code is behaving correctly.  I added the case here as documentation and for my future fix for curation validation.

PR #1204, opened 2024-10-18
Fix GitHub package registry lookup

ljones140:
Update `.npmrc` so that `@clearlydefined` packages come from GitHub package registry


elrayle (approved):
Thanks for the update and quick fix to version identification.

PR #1209, opened 2024-10-23
Update deployment pipelines 3 1 1

ljones140:
Updates deployment pipeline.

Important that prod is updated as we will now send App Version and Build Sha to docker image when built.
This will be returned on GET `https://api.clearlydefined.io/`

### DIFF
--- a/.github/workflows/build-and-deploy-prod.yml
+++ b/.github/workflows/build-and-deploy-prod.yml
@@ -21,8 +21,8 @@ jobs:
           path: package-lock.json
 
   build-and-deploy-prod:
-        needs: upload-package-lock-json
-        uses: clearlydefined/operations/.github/workflows/app-build-and-deploy.yml@v3.1.0
+    needs: upload-package-lock-json
+    uses: clearlydefined/operations/.github/workflows/app-build-and-deploy.yml@v3.1.0
     secrets: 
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-deploy-prod.yml
```diff
@@ -7,8 +7,21 @@ on:
...
+
+      - name: Upload package-lock.json
+        uses: actions/upload-artifact@v4
+        with:
+          name: package-lock.json
+          path: package-lock.json
+
   build-and-deploy-prod:
-    uses: clearlydefined/operations/.github/workflows/app-build-and-deploy.yml@v1.1.0
+    uses: clearlydefined/operations/.github/workflows/app-build-and-deploy.yml@v3.1.0
```
elrayle, line 24:
```suggestion
        needs: upload-package-lock-json
        uses: clearlydefined/operations/.github/workflows/app-build-and-deploy.yml@v3.1.0
```

elrayle (commented):
One suggestion, otherwise looks good.  Thanks for adding the artifact!

.github/workflows/build-and-deploy-prod.yml
ljones140: Done, thanks

elrayle (approved):
Looks good.  Thanks for bringing this workflow in line with the others.

Issue #1181, opened 2024-08-20
Failure calling curation patch API with certain license expressions


qtomlinson:
**Expected:**
should be able to call patch /curate with payload include valid license expressions

**Observed:**
Error message: 
`{"errors": [[{"message": "Invalid license in curation","error": "2.25.1 licensed.declared with value \"GPL-2.0-Only AND GPL-3.0-Or-Later AND LGPL-2.1-Only AND MPL-1.1\" is not SPDX compliant"}]],"patchesInError": [{"coordinates": {"type": "maven","provider": "mavencentral","namespace": "org.glassfish.jersey.core","name": "jersey-client"},"revisions": {"2.25.1": {"licensed": {"declared": "GPL-2.0-Only AND GPL-3.0-Or-Later AND LGPL-2.1-Only AND MPL-1.1"}}}}]}`

**Steps to reproduce:**
`curl -X PATCH "https://dev-api.clearlydefined.io/curations" -H "accept: application/json" -H "Content-Type: application/json" -d "{\"contributionInfo\":{\"summary\":\"test\",\"details\":\"test details\",\"resolution\":\"test resolution\",\"type\":\"Missing\"},\"patches\":[{\"coordinates\":{\"type\":\"maven\",\"provider\":\"mavencentral\",\"namespace\":\"org.glassfish.jersey.core\",\"name\":\"jersey-client\"},\"revisions\":{\"2.25.1\":{\"licensed\":{\"declared\":\"GPL-2.0-Only AND GPL-3.0-Or-Later AND LGPL-2.1-Only AND MPL-1.1\"}}}}]}"`

yashkohli88:
I tried debugging the curations API and found that the license name should be exactly same as the identifier present in the [spdx list](https://spdx.org/licenses/) as service is using the spdx parser to identify the license.

Then one more thing to take care of while curating complex license expression is to add brackets in some of the expressions. One of such case is described below.

Below is the license expression used to curate these three licenses successfully.
Here is the PR for the curation which I tried on my local system and personal repo - https://github.com/yashkohli88/curated-data-dev/pull/4
License expression - GPL-2.0-only OR (LGPL-2.1-only OR MPL-1.1)

The difference in the parsed license expression and the input license expression is at the [Line 79](https://github.com/clearlydefined/service/blob/c9419ab13d4e821da878404ecb83ad6ab0a71430/lib/curation.js#L79)

Below image with breakpoint on the same line confirms.

[...]

PR #1223, opened 2024-11-01
Fix curation validation

qtomlinson:
When the curated license contains NOASSERTION (e.g., MIT AND NOASSERTION), the curation should be invalid.  In addition, when the curated license is not normalized, include the normalized license expression in the error message.

In the future, we can consider ignoring the cases of SPDX identifiers as per [SPDX specification](https://spdx.github.io/spdx-spec/v2-draft/SPDX-license-expressions/#d2-case-sensitivity).  For example, "MIT, Mit and mIt should all be treated as the same identifier".

### DIFF
--- a/lib/curation.js
+++ b/lib/curation.js
@@ -77,8 +77,10 @@ class Curation {
 
     sourceLicenseList.forEach(({ source, license }) => {
       const parsed = SPDX.normalize(license)
-      if (parsed !== license || parsed === 'NOASSERTION') {
+      if (!parsed || parsed.includes('NOASSERTION')) {
         errors.push(`${source} with value "${license}" is not SPDX compliant`)
+      } else if (parsed !== license) {
+        errors.push(`${source} with value "${license}" is not normalized. Suggest using "${parsed}"`)
       }
     })
 
--- a/test/lib/curation.js
+++ b/test/lib/curation.js
@@ -89,20 +89,83 @@ describe('Curations', () => {
     expect(curation.errors[0].error.message).to.equal('Licensed object can only contain declared')
   })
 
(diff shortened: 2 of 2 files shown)


qtomlinson:
This is related to https://github.com/clearlydefined/service/issues/1181

test/lib/curation.js
```diff
@@ -89,20 +89,83 @@ describe('Curations', () => {
...
-  })
-
-  it('should identify invalid file licenses (not SPDX valid)', () => {
-    const content = getFixture('curation-invalid.11.yaml')
-    const curation = new Curation(content)
-    expect(curation.isValid).to.be.false
-    expect(curation.errors[0].error).to.equal(
-      '/foo in 4.17.4 files with value "mit and apache-2.0" is not SPDX compliant'
-    )
+  describe('declared licenses', () => {
```
elrayle, line 92: I appreciate all the tests.  Makes it very clear what is happening.

lib/curation.js
```diff
@@ -77,8 +77,10 @@ class Curation {
 
     sourceLicenseList.forEach(({ source, license }) => {
       const parsed = SPDX.normalize(license)
-      if (parsed !== license || parsed === 'NOASSERTION') {
+      if (!parsed || parsed.includes('NOASSERTION')) {
         errors.push(`${source} with value "${license}" is not SPDX compliant`)
+      } else if (parsed !== license) {
+        errors.push(`${source} with value "${license}" is not normalized. Suggest using "${parsed}"`)
```
elrayle, line 83: This is excellent feedback to the creator of the curation.

elrayle (approved):
Thanks for making it easier for the curator to understand, not just that something went wrong, but what went wrong.

PR #1232, opened 2024-11-13
Optimize recomputing definition

qtomlinson:
- Fix utils/getLatestVersion and add unit tests
- Allow injection of logger in AbstractAzBlobStore and AbstractFileStore
- Introduce getAllLatest to harvest stores, AzHarvestBlobStore and FileHarvestStore, to retrieve the latest versions of harvest tool results.  It will fall back to getAll results in case any error in calculating the latest files. Use getAllLatest in definition computation.
- Add an integration test to ensure that the definition computed using the latest harvest results are the same as the definition computed using all the harvest results.
- Add unit tests

Timing of computing definition for npm/npmjs/-/debug/3.1.0:
before optimization: retrieve blob: 3.2 sec (without 2.9.0+b1), calculating summary: 0.004 sec, curation: 1.2 sec
after optimization: retrieve blob: 1.3 sec, calculating summary: 0.02 sec, curation: 1.2 sec

Task: https://github.com/clearlydefined/service/issues/1237


elrayle (approved):
Excellent PR.  Always happy to see performance improvements.

providers/stores/azblobHarvestStore.js
```diff
@@ -46,34 +46,40 @@ class AzHarvestBlobStore extends AbstractAzBlobStore {
...
+  async getAll(coordinates) {
     // Note that here we are assuming the number of blobs will be small-ish (<10) and
     // a) all fit in memory reasonably, and
     // b) fit in one list call (i.e., <5000)
-    const list = new Promise((resolve, reject) =>
+    const files = await this._getAllFiles(coordinates)
+    return await this._getContent(files)
+  }
+
+  _getAllFiles(coordinates) {
```
elrayle, line 57: Suggesting name change here and in fileHarvestStore to: `_getListOfAllFiles(coordinates)`

providers/stores/azblobHarvestStore.js
```diff
@@ -85,6 +91,35 @@ class AzHarvestBlobStore extends AbstractAzBlobStore {
...
+
+  /**
+   * Get the latest version of each tool output for the given coordinates. The coordinates must be all the way down
+   * to a revision.
+   * @param {EntityCoordinates} coordinates - The component revision to report on
+   * @returns {Promise} A promise that resolves to an object with a property for each tool and tool version
+   *
+   */
+  async getAllLatest(coordinates) {
+    const allFiles = await this._getAllFiles(coordinates)
```
elrayle, line 103: Suggesting name change here and in fileHarvestStore to: 
```suggestion
    const allFilesList = await this._getListOfAllFiles(coordinates)
```

PR #1242, opened 2024-11-27
Add the ability to queue definition recomputation

qtomlinson:
**Performance Degradation During Schema Version Upgrade**

During the process of upgrading the definition schema version, performance degradation has been observed. Currently, schema version updates are handled by recomputing the definition on the fly, which causes slower performance. These updates are critical for data correctness and to introduce the LicenseRef marker.

**Proposed Solution:**

1. **Current Implementation (versionCheck):**
- Checks the schema version and recomputes the definition on demand if it is stale.
- This approach is currently in production.

2. **Preferred Approach (upgradeQueue):**
- When a request is received, return the existing definition.
- If the schema has changed, trigger a recompute after returning the result.
- Use a queue to handle the upgrade, similar to how the service computes definitions for harvested components.
- This allows the system to function more smoothly without immediate performance degradation.
- The tech steering committee has agreed on this approach (discussed on Nov 26).

3. **Switching Methods:**
- Once the majority of the migration is complete, the system can switch back to calculating on the fly via configuration.

**PR Details:**

This PR introduces two implementations for the upgrade handler:
- **versionCheck:** Current implementation in production.
- **upgradeQueue:** Preferred approach as described above.

**Environment Variables:**
- **DEFINITION_UPGRADE_PROVIDER:** Controls the upgrade behavior (versionCheck, upgradeQueue) with `versionCheck` as the default.
- **DEFINITION_UPGRADE_QUEUE_PROVIDER:** Controls the queuing of upgrades (memory, azure) with `memory` as the default.
- **DEFINITION_UPGRADE_QUEUE_CONNECTION_STRING:** Connection string to the Azure Storage Queue.
- If not present, `HARVEST_AZBLOB_CONNECTION_STRING` is used for connection information.
- **DEFINITION_UPGRADE_QUEUE_NAME:** Name of the upgrade queue, defaulting to `definitions-upgrade`.

### DIFF
--- a/providers/queueing/memoryQueue.js
+++ b/providers/queueing/memoryQueue.js
@@ -9,6 +9,7 @@ class MemoryQueue {
     this.logger = logger()
     this.data = []
     this.messageId = 0
+    this.decoder = options.decoder
   }
 
   async initialize() {}
@@ -33,9 +34,13 @@ class MemoryQueue {
     const message = this.data[0]
     if (!message) return null
     this.data[0].dequeueCount++
-    if (message.dequeueCount <= 5) return Promise.resolve({ original: message, data: JSON.parse(message.messageText) })
+    if (message.dequeueCount <= 5) return Promise.resolve({ original: message, data: this._parseData(message) })
     await this.delete({ original: message })
     return this.dequeue()
+  }
+
+  _parseData({ messageText }) {
+    return JSON.parse(this.decoder(messageText))
   }
 
   /** Similar to dequeue() but returns an array instead. See AzureStorageQueue.dequeueMultiple() */
@@ -59,4 +64,12 @@ class MemoryQueue {
   }
 }
 
-module.exports = () => new MemoryQueue()
+const factory = (opts = {}) => {
+  const defaultOpts = {
+    decoder: text => text
+  }
+  const mergedOpts = { ...defaultOpts, ...opts }
+  return new MemoryQueue(mergedOpts)
+}
+
+module.exports = factory
--- a/providers/queueing/memoryQueue.js
+++ b/providers/queueing/memoryQueue.js
@@ -40,7 +40,8 @@ class MemoryQueue {
 
   /** Similar to dequeue() but returns an array instead. See AzureStorageQueue.dequeueMultiple() */
   async dequeueMultiple() {
-    return [await this.dequeue()]
+    const message = await this.dequeue()
+    return message ? [message] : []
   }
 
   /**
--- a/providers/upgrade/defUpgradeQueue.js
+++ b/providers/upgrade/defUpgradeQueue.js
@@ -1,22 +1,28 @@
 // (c) Copyright 2024, SAP SE and ClearlyDefined contributors. Licensed under the MIT license.
 // SPDX-License-Identifier: MIT
 
-const logger = require('../logging/logger')
 const { lt } = require('semver')
 const { DefinitionVersionChecker } = require('./defVersionCheck')
 const EntityCoordinates = require('../../lib/entityCoordinates')
-const { setup } = require('./process')
+const setup = require('./process')
 
 class DefinitionQueueUpgrader extends DefinitionVersionChecker {
   async validate(definition) {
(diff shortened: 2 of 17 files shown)


providers/upgrade/defUpgradeQueue.js
```diff
@@ -0,0 +1,43 @@
...
+
+    await this._queueUpgrade(definition)
+    return definition
+  }
+
+  async _queueUpgrade(definition) {
+    if (!this._upgrade) throw new Error('Upgrade queue is not set')
+    const message = this._constructMessage(definition)
+    await this._upgrade.queue(JSON.stringify(message))
+    this.logger.debug('Queued for definition upgrade ', {
```
elrayle, line 22: I've heard this referred to as `recompute` which I see in the code in other places. I don't see `upgrade` in other parts of the code.  Is there a reason to name the queue and method `upgrade` instead of `recompute`.

elrayle (approved):
I looked at this really close and it looks good to me.  Probably should get someone with more js experience to review it too.

providers/upgrade/defUpgradeQueue.js
qtomlinson: This recompute happens only when schema version is update/upgraded. I am open to renaming it if it improves the readability.

providers/queueing/memoryQueue.js
```diff
@@ -33,14 +34,19 @@ class MemoryQueue {
     const message = this.data[0]
     if (!message) return null
     this.data[0].dequeueCount++
-    if (message.dequeueCount <= 5) return Promise.resolve({ original: message, data: JSON.parse(message.messageText) })
+    if (message.dequeueCount <= 5) return Promise.resolve({ original: message, data: this._parseData(message) })
```
mpcen, line 37: This was the `messageText` property of the `message` object before. Was this intentional?

jeffmendoza (approved):

providers/upgrade/defUpgradeQueue.js
```diff
@@ -0,0 +1,43 @@
...
+const EntityCoordinates = require('../../lib/entityCoordinates')
+const { setup } = require('./process')
+
+class DefinitionQueueUpgrader extends DefinitionVersionChecker {
+  async validate(definition) {
+    if (!definition) return
+    const result = await super.validate(definition)
+    if (result) return result
+
+    await this._queueUpgrade(definition)
```
mpcen, line 14: Is it worth capturing the possible exception here? Maybe wrapping this in a try/catch

providers/upgrade/process.js
```diff
@@ -0,0 +1,90 @@
...
+    if (!result) {
+      await this._definitionService.computeStoreAndCurate(coordinates)
+      this.logger.info(`Handled definition update for ${coordinates.toString()}`)
+    } else {
+      this.logger.debug(`Skipped definition update for ${coordinates.toString()}`)
+    }
+  }
+}
+
+let queueHandler
```
mpcen, line 81: Can we move `queueHandler` and `defUpgrader` into `setup`'s  scope or is there a specific reason why these are scoped higher?

mpcen (approved):
Overall everything looks good. Left minor comments/questions.

providers/queueing/memoryQueue.js
qtomlinson: `message` is destructured to extract the `messageText` property on line 42

providers/upgrade/process.js
qtomlinson: Good point!

providers/upgrade/defUpgradeQueue.js
qtomlinson: Definitely! Added logging in this._queueUpgrade.

qtomlinson:
Upon deployment, settings to turn on queuing: 
```
DEFINITION_UPGRADE_PROVIDER=upgradeQueue
DEFINITION_UPGRADE_QUEUE_PROVIDER=azure
```

PR #1275, opened 2025-01-17
Update pypi base url

qtomlinson:
Based on docs at https://packaging.python.org/en/latest/guides/migrating-to-pypi-org/, "the default interface for browsing packages is pypi.org. The domain pypi.python.org now redirects to pypi.org, and may be disabled sometime in the future."

Also see pypi json api documentation at https://docs.pypi.org/api/json/

Update the urls for the pypi json api in originPypi and pypiCoordinatesMapper to reflect this.


elrayle (approved):
Looks good.  Nice catch of a change to package manager.

PR #1334, opened 2025-07-04
Additional types

JamieMagee:
The majority of these types were straightforward to add, and should be self-explanatory to review. However, there were a few places where I had to workaround some issues. I've left comments at the relevant places in the code.

Other than that, I also:
- Added `@tsconfig/strictest`
- disabled `strictNullChecks` as it's not really possible to enforce in JavaScript
- Added custom types for:
  - `@clearlydefined/spdx` as it doesn't ship with them (I will contribute then back to [clearlydefined/spdx](https://github.com/clearlydefined/spdx/) afterwards)
  - `painless-config` as the types it ships with are incorrect and the upstream repository is archived
  - `winston-azure-application-insights` as it doesn't ship with them


lib/fetch.js
```diff
@@ -31,23 +67,40 @@ function buildRequestOptions(request) {
...
+ * @returns {Promise<any | FetchResponse>} Promise that resolves to the response data or full response object
+ * @throws {FetchError} When the request fails or returns an error status code
+ */
 async function callFetch(request, axiosInstance = axios) {
   try {
     const options = buildRequestOptions(request)
     // @ts-ignore
     const response = await axiosInstance(options)
     if (!request.resolveWithFullResponse) return response.data
+    // @ts-ignore - Adding custom properties to response object
```
JamieMagee, line 85: Done to avoid

```
Property 'statusCode' does not exist on type 'AxiosResponse<any, any>'.ts(2339)
```

lib/fetch.js
```diff
@@ -31,23 +67,40 @@ function buildRequestOptions(request) {
...
     response.statusCode = response.status
+    // @ts-ignore - Adding custom properties to response object
```
JamieMagee, line 87: Done to avoid

```
Property 'statusMessage' does not exist on type 'AxiosResponse<any, any>'.ts(2339)
```

lib/fetch.js
```diff
@@ -31,23 +67,40 @@ function buildRequestOptions(request) {
...
     response.statusMessage = response.statusText
     return response
   } catch (error) {
+    // @ts-ignore - Adding statusCode property to error
```
JamieMagee, line 91: Done to avoid

```
Property 'statusCode' does not exist on type 'AxiosResponse<any, any>'.ts(2339)
```

lib/pypiCoordinatesMapper.js
```diff
@@ -29,7 +99,7 @@ class PypiCoordinatesMapper {
       const answer = await this._fetch({ url, method: 'GET', json: true })
       return answer?.info?.name && { name: answer.info.name }
     } catch (error) {
-      if (error.statusCode === 404) return null
+      if (/** @type {FetchError} */ (error).statusCode === 404) return null
```
JamieMagee, line 102: Done to avoid

```
Property 'statusCode' does not exist on type 'unknown'.ts(2339)
```

lib/resultCoordinates.js
```diff
@@ -26,24 +43,54 @@ class ResultCoordinates {
...
+   *
+   * @param {string | null | undefined} urn - URN string in format
+   *   "scheme:type:provider:namespace:name:revision:revision:tool:tool:toolVersion"
+   * @returns {ResultCoordinates | null} New ResultCoordinates instance or null if urn is invalid
+   */
   static fromUrn(urn) {
     if (!urn) return null
     // eslint-disable-next-line no-unused-vars
-    const [scheme, type, provider, namespace, name, revToken, revision, toolToken, tool, toolVersion] = urn.split(':')
+    const [, type, provider, namespace, name, , revision, , tool, toolVersion] = urn.split(':')
```
JamieMagee, line 56: Done to avoid

```
'...' is declared but its value is never read.ts(6133)
```

types/clearlydefined-spdx.d.ts
```diff
@@ -0,0 +1,177 @@
+/**
+ * TypeScript type definitions for @clearlydefined/spdx package
+ *
+ * This module provides utilities for parsing, normalizing, and manipulating SPDX license expressions. It supports
+ * operations like parsing expressions into ASTs, converting back to strings, checking license satisfaction, and merging
+ * license expressions.
+ */
```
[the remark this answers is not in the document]
JamieMagee: Done https://github.com/clearlydefined/spdx/pull/37

qtomlinson (commented):
Thank you for the detailed and excellent work! Several of my comments focus on the declaration of private properties or methods in different parts of the code. Let me know if you need any clarification.

lib/fetch.js
qtomlinson: As a future separate PR, we can potentially adapt all callers to use the equivalent fields in AxiosResponse to avoid adding custom properties to the response object

lib/condaRepoAccess.d.ts
```diff
@@ -0,0 +1,130 @@
...
+/** Package search result */
+export interface CondaPackageMatch {
+  /** Package identifier */
+  id: string
+}
+
+/** Main class for accessing Conda repository data */
+declare class CondaRepoAccess {
+  /** Cache instance for storing fetched data */
+  private cache: ICache
```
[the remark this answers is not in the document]
JamieMagee: It's not necessary, but I added types for private fields and methods for a few reasons:
- Future-proofing for a TypeScript migration
- Better tooling and IDE integration
- I was already here, and it wasn't much additional effort 😅

lib/entityCoordinates.d.ts
```diff
@@ -11,7 +11,7 @@ export interface EntityCoordinatesSpec {
 }
 
 /** Represents entity coordinates for a software component */
-declare class EntityCoordinates {
+export declare class EntityCoordinates {
   /** The type of the entity (e.g., 'npm', 'maven', 'git') */
   type?: string
```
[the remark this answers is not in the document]
JamieMagee: It does, yes. Good catch.

lib/mockInsights.js
```diff
@@ -3,27 +3,72 @@
...
+ *
+ * @typedef {import('applicationinsights').Contracts.ExceptionTelemetry} ExceptionTelemetry
+ *
+ * @typedef {import('applicationinsights').Contracts.TraceTelemetry} TraceTelemetry
+ */
+
+/**
+ * Mock implementation of Application Insights client for testing and development. This class provides a lightweight
+ * alternative to the full Application Insights client, allowing for console-based logging of telemetry data during
+ * development and testing.
```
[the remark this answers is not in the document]
JamieMagee: Thanks for the explanation. I'll update the doc.

lib/gradleCoordinatesMapper.d.ts
```diff
@@ -0,0 +1,152 @@
...
+  /**
+   * Determines if the given coordinates should be resolved.
+   *
+   * Resolution is needed when there's no namespace but there is a name, indicating this is a Gradle plugin that needs
+   * to be mapped to its Maven coordinates.
+   *
+   * @param coordinates - The coordinates to check
+   * @returns True if the coordinates should be resolved, false otherwise
+   */
+  private _shouldResolve(coordinates: GradleCoordinates): boolean
```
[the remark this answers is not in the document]
JamieMagee: It's not necessary, but I added types for private fields and methods for a few reasons:
- Future-proofing for a TypeScript migration
- Better tooling and IDE integration
- I was already here, and it wasn't much additional effort 😅

lib/rateLimit.js
```diff
@@ -65,40 +173,112 @@ class RedisBasedRateLimiter extends RateLimiter {
...
     return (request, response, next) => {
       this._createMiddleware()
+        .then(middleware => {
+          if (middleware) {
+            middleware(request, response, next)
+          } else {
+            next(new Error('Failed to create middleware'))
+          }
+        })
         .catch(error => next(error))
```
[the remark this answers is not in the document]
JamieMagee: With the original implementation I get the following TypeScript error:

```
This expression is not callable.
  Not all constituents of type 'void | RateLimitMiddleware' are callable.
    Type 'void' has no call signatures.ts(2349)
```

I'll revert it back, and add an `if` statement to guard against this. Something like:

```ts
this._createMiddleware()
  .catch(error => next(error))
  .then(middleware => {
    if (middleware) {
      middleware(request, response, next)
    }
  })
```

lib/fetch.js
JamieMagee: Yeah, I agree. My approach at the moment is to avoid touching the underlying JavaScript code as much as is possible. Changing underlying code can come in a follow-up wave.

qtomlinson (approved):
@JamieMagee Thank you for incorporating the review comments.

Declaring private fields or methods within the types file can reduce readability. For instance, the GradleCoordinatesMapper class contains only two public methods:
```
map(coordinates: GradleCoordinates): Promise<EntityCoordinates | null>
getMavenMetadata(pluginId: string): Promise<string | FetchResponse<string> | null>
```
This is easier to understand compared to a scenario where there are 2 public and 8 private methods.

Nonetheless, this is not a critical issue. I appreciate the detailed and excellent changes you've made!

PR #1356, opened 2025-10-01
prettier and eslint driven changes so those tests will pass now. 223+ files impacted but looks like non-important changes

JustinGOSSES:
Saw the prettier test failed on my last PR so submitting this PR to catch up such that all files past the prettier test now.

- Prettier and eslint tests will pass now. 
- 223+ files impacted but looks like non-important changes


JustinGOSSES:
Never mind, looks like someone already fixed those issues in master branch since last checked and so only 1 change. Will delete this PR.

PR #1302, opened 2025-04-23
Optimize Queueing and Definition Retrieval for Component Requested For Harvest

qtomlinson:
#### Background:
Stats from 2025-03-16 to 2025-04-17 for the "get definition" process:
- Averaging 17 million "definition not available" counts daily (with 8.5 million definition computations triggering harvest requests).
- Averaging 250,000 distinct coordinates with "definition not available" daily.
image
image

#### Problems Identified:
1. **Duplicate Harvest Requests**:
    - **Queue Growth**: The queue expands due to redundant harvest requests.
    - **Scanner Tool Failures**: The crawler retries a failed request up to 5 times. The additional repeated queuing of the failed components forces the crawler to fail-retry more than necessary, thereby hindering overall progress.
  
2. **Duplicate Definition Computations**:
    - Definition computations for unavailable definitions involve 4 network calls, adding load on internal dependencies (stores and queue).

#### Solution:
**Introducing CacheBasedHarvester**:
- **Caching Mechanism**: Cache harvest coordinates to mark them as 'in-progress'. 
- Cache entries expire daily.
- Cached entries are removed when a definition is successfully recomputed after crawler tool results are available.

**Benefits**:
- Reduced queueing of duplicated harvest requests.
- Optimized definition computation: By tracking in-progress harvest coordinates, definition computations can return placeholders without triggering network calls. This reduces daily definition computations for those "definition not available" cases from 8.5 million to 250,000 (one per unique coordinate).

#### Impact:
- Provisioned cache capacity needs confirmation to ensure sufficient resources for the additional caching.

#### Future Work:
- Transitioning definition computations during definition retrieval to the background to improve the responsiveness of API calls.

PR #1361, opened 2025-11-12
Cache computed empty definition

qtomlinson:
Investigations have shown that empty definitions are repeatedly requested and computed for certain coordinates. PR #1302 optimized 'definition not available' cases, reducing network calls and processing time for each case. However, this increased the concurrency of definition blob retrievals, potentially raising blob store latency. To address this, this PR implements caching of empty definitions to minimize repeated retrievals from the definition blob store.

### DIFF
--- a/routes/notices.js
+++ b/routes/notices.js
@@ -9,7 +9,6 @@ const validator = require('../schemas/validator')
 const logger = require('../providers/logging/logger')
 const bodyParser = require('body-parser')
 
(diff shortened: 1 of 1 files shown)


jeffmendoza (approved):

qtomlinson:
@pombredanne @jamesiri This is the cache work that I mentioned during our dev sync meeting

PR #1362, opened 2025-11-12
Return Blank Response for Versionless Coordinate

yashkohli88:
### Summary

This PR enhances the `DefinitionService` by ensuring that API requests for coordinates without a specified revision (versionless calls) return a blank (undefined) response. Previously, such requests would incorrectly return a definition object, even when a revision was not present. This update brings consistency to both API behavior and testing, and ensures that only versioned coordinates yield valid definitions.

### Changes in This PR

- **business/definitionService.js**
    - Updated the `get()` method to immediately return undefined if the coordinate lacks a revision.
- **test/business/definitionServiceTest.js**
    - Added and updated test cases:
        - Verifies that calling `get()` with a versionless coordinate returns undefined.
        - Confirms that calling `get()` with a coordinate that includes a revision returns the expected definition.

### DIFF
--- a/.github/workflows/build-and-deploy-dev.yml
+++ b/.github/workflows/build-and-deploy-dev.yml
@@ -12,10 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v4.1.1
+        uses: actions/checkout@v5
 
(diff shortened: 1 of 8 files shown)


yashkohli88:
Dev testing  - https://github.com/yashkohli88/operations/actions/runs/19624186657/job/56189927349

PR #1366, opened 2025-11-17
update to node 24 packages

brettfo:
Hi all, I'm Brett and I'm a software engineer at Microsoft.  We use ClearlyDefined a lot internally and I was asked by my management to update some things in this repo.  I figured I should start with something small and hopefully easy to review.  Some of the node-adjacent packages used referenced either node 20 or outdated Docker images so this is an attempt to rectify that.

My plan for future PRs is to update various dependencies as required along with improving test coverage where possible.

I'm open to any and all feedback and if you'd like the feedback to be private, you can email me directly at `brettfo@microsoft.com`.

### DIFF
--- a/lib/scancodeMap.d.ts
+++ b/lib/scancodeMap.d.ts
@@ -1,0 +1,28 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+
+/**
+ * A scancode license key (e.g., 'mit', 'apache-2.0', 'gpl-3.0')
+ */
+export type ScancodeKey = string
+
+/**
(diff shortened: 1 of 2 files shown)


test/providers/store/mongoDefinitionPagination.js
```diff
@@ -32,8 +32,11 @@ const shouldPaginateSearchCorrectly = function () {
     })
 
     after('cleanup database', async function () {
-      await mongoStore.collection.drop()
-      await mongoStore.close()
+      if (mongoStore) {
+        await mongoStore.collection.drop()
```
brettfo: I found during some local testing that the `before()` function didn't always complete before `after()` which would mean the variable `mongoStore` was still `undefined` and this line would throw.

pombredanne (approved):
Thanks. Looking great.

PR #1340, opened 2025-07-15
feat: update nodejs and other packages

elaine-mattos:
Hi dear clearlydefined community,

My team is currently evaluating the ClearlyDefined API for integration into our compliance workflows. While testing in our fork, we noticed several deprecated dependencies and decided to modernize the setup.

This PR upgrades the codebase to support Node.js ~~20~~ 24 and updates a range of dependencies. Along the way, we made necessary refactors due to breaking changes, and introduced improvements in initialization and error handling.

We're still getting familiar with the codebase, so please feel free to comment or guide us if we’ve misunderstood anything. Also, if this PR feels too large, we’re happy to split it into smaller pieces.

## Some details
  - Node and Dependency updates:
    - Upgraded project to Node.js ~~20~~ 24;
    - Updated multiple dependencies to their latest compatible versions.  
  - Express 5 migration:
    - Removed `express-init` in favor of direct async initialization; 
    - Updated middleware and route initialization to match Express 5 syntax;
    - Changed `app.options('*', cors())` to `app.options('*splat', cors())` per new path syntax.
  - GitHub & GitLab API usage:
    - Updated all usages of `@octokit/rest` and `@gitbeaker/node` to match new APIs and import styles;
    - Replaced deprecated authentication and client initialization patterns.
  - YAML parsing:
     -  Replaced deprecated `yaml.safeLoad` and `yaml.safeDump` with their respectives `yaml.load`and `yaml.dump`.
  - Logging: 
     - Introduced `LOGGER_LOG_TO_CONSOLE` environment variable to control log output to the console;
     - Introduced the `ENABLE_REST_VALIDATION_ERRORS` environment variable to control REST input validation verbosity. When enabled, `ajv` will report all validation errors (`allErrors: true`), which is useful for debugging but may trigger deep object traversal warnings. [More info](https://codeql.github.com/codeql-query-help/javascript/js-resource-exhaustion-from-deep-object-traversal/)
     - Improved debug and error logs for critical operations.
     
Let us know your thoughts! ^.^

PR #1383, opened 2025-12-17
feat(cache): add backward compatibility for pako 1.x to 2.x migration

elaine-mattos:
Hi,

as part of the Node.js and package updates effort (#1340), here is the follow-up PR that adds  backward compatibility to handle Redis cache entries compressed with both pako 1.x (binary string) and pako 2.x (base64) formats, preventing cache invalidation during the upgrade.

## Summary

Adds backward compatibility to handle Redis cache entries compressed with both pako 1.x (binary string) and pako 2.x (base64) formats, preventing cache invalidation during the upgrade.

## Changes

- **Format detection in `redis.js`**: Automatically detects whether cached data is in old binary string format (pako 1.x) or new base64 format (pako 2.x)
- **Backward compatibility**: Existing cache entries remain readable after upgrading to pako 2.x
- **Test coverage**: Added comprehensive tests for both `def_*` (definitions) and `hrv_*` (harvest) key patterns

## Testing

- Unit tests verify reading both old (pako 1.x binary) and new (pako 2.x base64) formats
- Tested with local Redis container using keys written with pako 1.x and read with pako 2.x
- Tests use both `pako@1.0.8` and `pako@2.1.0` to prove cross-version compatibility

## Benefits

- Zero cache invalidation on deployment
- Seamless migration from pako 1.x to 2.x
- Both formats coexist during transition period

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -102,6 +102,7 @@
         "node-mocks-http": "1.17.2",
         "nodemon": "3.1.11",
         "nyc": "17.1.0",
+        "pako-1": "npm:pako@^1.0.8",
         "prettier": "3.8.1",
         "proxyquire": "2.1.3",
         "sinon": "21.0.1",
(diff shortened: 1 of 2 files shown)


providers/caching/redisConfig.js
```diff
@@ -23,7 +23,7 @@ function serviceFactory(options) {
   const realOptions = options || {
     service: config.get('CACHING_REDIS_SERVICE'),
     apiKey: config.get('CACHING_REDIS_API_KEY'),
-    port: Number(config.get('CACHING_REDIS_PORT')) || 6380
+    port: Number(config.get('CACHING_REDIS_PORT'))
```
JamieMagee: Was the removal of `|| 6380` here intentional? `redis.js` has it in the function signature default, and `config.js` has added the fallback.

PR #1394, opened 2026-01-14
feat(types): add TypeScript definitions for remaining lib/ modules

JamieMagee:
## Summary

This PR adds TypeScript type definitions for the remaining modules in the `lib/` directory, completing the TypeScript coverage for this folder as part of the ongoing migration outlined in [docs/typescript-migration.md](docs/typescript-migration.md).

## New Type Definition Files

- `lib/utils.d.ts` - Core utility functions for coordinates, dates, license handling, and definition merging
- `lib/curation.d.ts` - Curation class and related types
- `lib/licenseMatcher.d.ts` - License matching policies and comparison types
- `lib/heapLogger.d.ts` - Heap logging configuration types

## JavaScript Code Changes

Several JavaScript files required modifications to satisfy TypeScript's type checking:

### lib/curation.js
- Add explicit `return undefined` in `load()` catch block
- Add `@ts-ignore` for `data.revisions` access after validation

### lib/heapLogger.js
- Wrap `heapStatsInterval` with `Number()` for setInterval compatibility

### lib/licenseMatcher.js
- Use union type for `_addFileToMap` propName parameter
- Add explicit undefined return and type cast in `getLatestToolHarvest`
- Cast `m.value` to string in `NugetHarvestLicenseMatchStrategy.compare`

### lib/utils.js
- Add `@ts-ignore` for `extend` module (no types available)
- Add `@ts-ignore` for Date subtraction in `compareDates`
- Add `@ts-ignore` for dynamic `sourcearchive` property assignment
- Cast `_mergeExcept` result to `FileEntry` in `_mergeFile`
- Fix `mergeDefinitions` return type to `void`

### schemas/validator.js
- Use bracket notation for `process.env` access
- Add `.default` to `ajv-formats` and `ajv-errors` imports
- Add `.json` extension to schema imports
- Cast `ajvOptions` to satisfy strict typing

### tsconfig.json
- Enable `resolveJsonModule` for JSON schema imports
- Add `schemas/validator.js` to include list

## Testing

- `npm run tsc` passes with no errors
- All existing tests should continue to pass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,7 +53,7 @@
         "semver": "7.7.2",
         "serialize-error": "^12.0.0",
         "spdx-expression-parse": "github:clearlydefined/spdx-expression-parse.js#fork",
-        "spdx-license-list": "^6.10.0",
+        "spdx-license-list": "^6.11.0",
         "swagger-ui-express": "^5.0.1",
         "throat": "^6.0.2",
         "tiny-attribution-generator": "0.1.3",
@@ -10001,9 +10001,9 @@
       "license": "CC0-1.0"
     },
     "node_modules/spdx-license-list": {
-      "version": "6.10.0",
-      "resolved": "https://registry.npmjs.org/spdx-license-list/-/spdx-license-list-6.10.0.tgz",
-      "integrity": "sha512-wF3RhDFoqdu14d1Prv6c8aNU0FSRuSFJpNjWeygIZcNZEwPxp7I5/Hwo8j6lSkBKWAIkSQrKefrC5N0lvOP0Gw==",
+      "version": "6.11.0",
+      "resolved": "https://registry.npmjs.org/spdx-license-list/-/spdx-license-list-6.11.0.tgz",
+      "integrity": "sha512-p5ICd51dSnh7zIMtPgbB9ShBg3HMT77OeI6WVhrFFvxa5KIFYNcqxD4joAE+n1zZ7wlJdEkrOMwC75JUMMmsJA==",
       "license": "CC0-1.0",
       "engines": {
         "node": ">=8"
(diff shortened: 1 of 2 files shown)


lib/curation.d.ts
```diff
@@ -0,0 +1,115 @@
...
+  attributions?: string[]
+}
+
+/** Revision-level curation data */
+export interface CurationRevision {
+  licensed?: {
+    declared?: string
+  }
+  described?: {
+    releaseDate?: string
```
qtomlinson: We have described.facets? and described.projectWebsite?, that can be curated

lib/utils.d.ts
```diff
@@ -0,0 +1,317 @@
...
+/** Described section of a definition */
+export interface DescribedSection {
+  releaseDate?: string
+  sourceLocation?: SourceLocationSpec
+  facets?: Record<string, string[]>
+  hashes?: Record<string, string>
+  files?: number
+  tools?: string[]
+  toolScore?: ScoreSection
+  score?: ScoreSection
```
qtomlinson: urls?: Record<string, string>

qtomlinson (approved):
Thanks for your contribution!

lib/utils.js
```diff
@@ -178,64 +279,109 @@ function extractLicenseFromLicenseUrl(licenseUrl) {
...
+ * Merges the given definition onto the base definition
+ * @param {Definition | null | undefined} base - Base definition to merge onto
+ * @param {Partial<Definition> | null | undefined} proposed - Proposed changes to merge
+ * @param {boolean} [override] - If true, proposed values override rather than merge
+ * @returns {void}
+ */
 function mergeDefinitions(base, proposed, override) {
   if (!proposed) return
-  if (!base) return proposed
+  if (!base) return
```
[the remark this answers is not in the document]
JamieMagee: I think this might actually have been a bug in the original implementation. If we have `proposed` and `base`, then base is modified in-place and this method is a `void`. However, if we have `proposed`, but not `base` then we return a value.

We don't actually have any users of this function that use the return value, so it is effectively a void method anyway.

lib/utils.d.ts
JamieMagee: Added

PR #1400, opened 2026-01-17
feat(types): add TypeScript definitions for summary providers

JamieMagee:
Adds TypeScript type definition files (`.d.ts`) and JSDoc annotations to all summary provider modules, following the gradual TypeScript migration strategy documented in `docs/typescript-migration.md`.

PR #1403, opened 2026-01-20
feat(types): add TypeScript definitions for business layer

JamieMagee:
## Summary

Add TypeScript type definitions for all business layer modules, continuing the gradual TypeScript migration strategy.

**Stacked on:** #1400

JavaScript code changes for TypeScript compatibility:

| File | Change | Reason |
|------|--------|--------|
| `aggregator.js` | `result.files?.length \|\| 0` | Handle undefined files array safely |
| `definitionService.js` | `return undefined` instead of implicit return | Explicit return type for type safety |
| `definitionService.js` | `/** @type {Error} */ (error)` casts in catch blocks | TypeScript needs explicit error types |
| `definitionService.js` | Inline type annotations for `throat()` callbacks | Arrow function parameters need types |
| `definitionService.js` | `@ts-ignore` on `_getDefinitionCoordinates` | Unused method kept for API compatibility |
| `noticeService.js` | `/** @type {string} */ (file.token)` | Token could be undefined per types |
| `suggestionService.js` | `hasSuggested = hasSuggested \|\|` instead of `\|=` | TypeScript doesn't allow bitwise ops on booleans |
| `statsService.js` | Type casts on `reduce()` accumulators | Empty object needs explicit type |
| `providers/summary/index.d.ts` | Added `BaseSummarizer` interface and index signature | Enable dynamic access to summarizers by string key |


business/definitionService.d.ts
```diff
@@ -0,0 +1,492 @@
...
+  tools?: string[]
+  /** Tool-derived score */
+  toolScore?: DescribedScore
+  /** Final score after curation */
+  score?: DescribedScore
+  /** Project website URL */
+  projectWebsite?: string
+  /** Facet definitions */
+  facets?: Record<string, string[]>
+}
```
qtomlinson: other possibilities in described section:
- hashes?: Record<string, string>
- urls?: Record<string, string>
- files?: number
- issueTracker?: string
```
    "urls": {
      "registry": "https://crates.io/crates/abscissa_derive",
      "version": "https://crates.io/crates/abscissa_derive/0.8.2",
      "download": "https://crates.io/api/v1/crates/abscissa_derive/0.8.2/download"
    },
    "hashes": {
      "sha1": "d8bd61d2fa814508a60a182fba0962d27230372e",
      "sha256": "08d914621d2ef4da433fe01907e323ee3f2807738d392d5a34c287b381f87fe2"
    },
    "files": 9,
```

business/definitionService.d.ts
```diff
@@ -0,0 +1,492 @@
...
+/** Licensed section of a definition */
+export interface DefinitionLicensed {
+  /** Declared license expression */
+  declared?: string
+  /** Tool-derived score */
+  toolScore?: LicensedScore
+  /** Final score after curation */
+  score?: LicensedScore
+  /** Facet information */
+  facets?: Record<string, FacetInfo>
```
qtomlinson: facets look good as is.  
In the future, is it worthwhile to introduce an enum to capture the ` fixed set of canonical facets` in this [doc](https://docs.clearlydefined.io/docs/resources/defined#facets)?

qtomlinson (approved):
LGTM!

business/definitionService.d.ts
JamieMagee: Yeah, good idea. I'll leave this as-is for now, and tackle that in a later pass.

PR #1439, opened 2026-01-27
Fix error handling and Buffer parsing in store implementations

JamieMagee:
- Cast catch block errors to `any` so we can check `error.code` and `error.statusCode`
- Call `Buffer.toString()` before `JSON.parse()` in fileAttachmentStore (was passing Buffer directly)
- Rename unused parameters with underscore prefix (`_result`, `_coordinates`) instead of eslint-disable comments
- Wrap `forEach` callback in braces to avoid returning the result of `delete` expression
- Use `Type` suffix for typedef imports to avoid shadowing `require()` constants

## Why these changes

The existing code worked fine at runtime but had patterns that confused the type checker. For example, catching an error and immediately accessing `error.code` fails because caught values are `unknown` by default. The Buffer issue was a real bug waiting to happen on some Node versions.


qtomlinson (approved):
Thanks for putting in the typing.  Comments are only for future enhancements.  
In the future, we may want to extract a common interface for DefinitionStore and have mongo, blob, dispatch and file definition store implementing it.

PR #1451, opened 2026-02-11
Fix service start error

qtomlinson:
The following error prevented api service start on the dev deployment: Error connecting to redis: RangeError [ERR_SOCKET_BAD_PORT]: Port should be >= 0 and < 65536. Received type number (NaN)

Add the default port setting back in the factory function for redis cache

### DIFF
--- a/bin/config.js
+++ b/bin/config.js
@@ -69,10 +69,7 @@ module.exports = {
     service: loadFactory(config.get('AUTH_PROVIDER') || 'github', 'auth')
   },
   caching: {
-    service: loadFactory(config.get('CACHING_PROVIDER') || 'memory', 'caching'),
-    caching_redis_service: config.get('CACHING_REDIS_SERVICE'),
-    caching_redis_api_key: config.get('CACHING_REDIS_API_KEY'),
-    caching_redis_port: config.get('CACHING_REDIS_PORT') || 6380
+    service: loadFactory(config.get('CACHING_PROVIDER') || 'memory', 'caching')
   },
   endpoints: {
     service: config.get('SERVICE_ENDPOINT') || 'http://localhost:4000',


bin/config.js
```diff
@@ -72,7 +72,7 @@ module.exports = {
     service: loadFactory(config.get('CACHING_PROVIDER') || 'memory', 'caching'),
     caching_redis_service: config.get('CACHING_REDIS_SERVICE'),
     caching_redis_api_key: config.get('CACHING_REDIS_API_KEY'),
-    caching_redis_port: config.get('CACHING_REDIS_PORT') || 6380
+    caching_redis_port: config.get('CACHING_REDIS_PORT')
```
JamieMagee: This property (along with `caching_redis_service` and `caching_redis_api_key`) isn't consumed anywhere — the factory in `redisConfig.js` reads from `painless-config` directly. With the `|| 6380` default removed, someone wiring `config.caching` into the factory later would get `undefined` for port.

Not a problem right now, but might be worth either removing these dead properties or adding a comment explaining why they exist.

providers/caching/redisConfig.js
```diff
@@ -23,7 +23,7 @@ function serviceFactory(options) {
   const realOptions = options || {
     service: config.get('CACHING_REDIS_SERVICE'),
     apiKey: config.get('CACHING_REDIS_API_KEY'),
-    port: Number(config.get('CACHING_REDIS_PORT'))
+    port: Number(config.get('CACHING_REDIS_PORT')) || 6380
```
JamieMagee: Minor: `Number("0")` produces `0`, which is falsy, so `|| 6380` would silently override it. Port 0 isn't a realistic Redis configuration so this probably doesn't matter, but the fallback behavior is technically surprising for any falsy-but-valid value.

JamieMagee (commented):
Fix looks correct. Two minor comments.

PR #1453, opened 2026-02-11
Add JSDoc type annotations and .d.ts files for route handlers

JamieMagee:
## Summary

- Add JSDoc `@param`/`@typedef` annotations to all route handler functions and companion `.d.ts` declaration files so TypeScript consumers get type info.
- Add a `tsconfig.json` for JS type-checking configuration.
- Widen the `AsyncMiddlewareFunction` return type from `Promise<void>` to `Promise<any>` to match actual usage (handlers return `response.send(...)` which yields `Response`).

## JS behavior changes

These are small and shouldn't affect production HTTP behavior, but are worth calling out:

**`routes/definitions-1.0.0.js`** — The `force` query param check dropped the `=== true` (boolean) comparison, keeping only `=== 'true'` (string). Express query values are always strings, so the boolean branch was dead code in normal requests. If any test or middleware sets the value as an actual boolean, that path would stop matching.

**`routes/definitions.js`** — Same treatment for `matchCasing` (dropped `=== false`). Also removed the `Array.isArray` guard around `request.get('User-Agent')` since Express always returns a single string for that header. Caught errors are now cast to `Error` before accessing `.message`.

**`routes/webhook.js`** — `return info(request, response, ...)` was split into `info(...); return`. Functionally equivalent today, but slightly more fragile to future edits since removing the `return` would cause a double-response. The unused `request` parameter in `info()` was renamed to `_request`.


routes/curations.js
```diff
@@ -86,33 +116,41 @@ async function updateCurations(request, response) {
   }
 
   const normalizedPatches = await Promise.all(
-    request.body.patches.map(async entry => {
+    request.body.patches.map(async (/** @type {any} */ entry) => {
```
qtomlinson: nit: should the entry be string | CurationData to be consistent with Curation's constructor. similarly on line 107.

routes/curations.js
```diff
@@ -122,12 +160,18 @@ async function reprocessMergedCurations(request, respond) {
   }
   const normalizedCoordinatesList = await Promise.all(coordinatesArray.map(utils.toNormalizedEntityCoordinates))
   const result = await curationService.reprocessMergedCurations(normalizedCoordinatesList)
-  respond.status(200).send(result)
+  return respond.status(200).send(result)
 }
 
+/** @type {any} */
 let curationService
+/** @type {any} */
```
qtomlinson: nit: Logger?

routes/definitions-1.0.0.js
```diff
@@ -33,13 +42,18 @@ async function getDefinition(req, resp) {
...
+/**
+ * @param {any} definition
+ * @param {boolean} [testFlag]
+ */
 function setup(definition, testFlag = false) {
   definitionService = definition
 
   if (testFlag) {
-    router._getDefinition = getDefinition
+    (router)._getDefinition = getDefinition
```
qtomlinson:
```suggestion
    const _router = /** @type {any} */ (router)
    _router._getDefinition = getDefinition
```

qtomlinson (approved):

PR #1483, opened 2026-03-10
Add type definitions for providers/search

JamieMagee:
Adds `.d.ts` files for `providers/search/` — `abstractSearch`, `memory`, `azureSearch`, and `azureConfig` — and wires them into `tsconfig.json`.

`AbstractSearch` is a base class with stub methods that `MemorySearch` and `AzureSearch` override. The types formalize that interface. The JS files get JSDoc annotations so `tsc` can check them, including `@override` tags required by the strictest config.

Most of the JSDoc changes are just adding `@param {any}` to silence implicit-any errors on internal methods. The only real fix was correcting `@returns {String[]}` to `@returns {Promise<string[]>}` on the async methods — the old JSDoc was wrong.

No behavior changes. `tsc` clean, tests pass.

### DIFF
--- a/providers/search/azureSearch.js
+++ b/providers/search/azureSearch.js
@@ -472,4 +472,4 @@ class AzureSearch extends AbstractSearch {
   }
 }
 
-module.exports = /** @param {any} options */ options => new AzureSearch(options)
+module.exports = /** @param {import('./azureSearch').AzureSearchOptions} options */ options => new AzureSearch(options)
--- a/providers/search/memory.js
+++ b/providers/search/memory.js
@@ -81,4 +81,4 @@ class MemorySearch extends AbstractSearch {
   }
 }
 
-module.exports = /** @param {any} options */ options => new MemorySearch(options)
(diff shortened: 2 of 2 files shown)


providers/search/azureSearch.js
```diff
@@ -448,4 +472,4 @@ class AzureSearch extends AbstractSearch {
   }
 }
 
-module.exports = options => new AzureSearch(options)
+module.exports = /** @param {any} options */ options => new AzureSearch(options)
```
qtomlinson: AzureSearchOptions instead of any?

providers/search/memory.js
```diff
@@ -24,35 +29,56 @@ class MemorySearch extends AbstractSearch {
...
+   * @param {any} coordinates
+   * @override
+   */
   delete(coordinates) {
     delete this.index[coordinates.toString()]
   }
 }
 
-module.exports = options => new MemorySearch(options)
+module.exports = /** @param {any} options */ options => new MemorySearch(options)
```
qtomlinson: SearchOptions instead of any to be consistent with d.ts?

PR #1484, opened 2026-03-11
Add type definitions for providers/harvest

JamieMagee:
Adds .d.ts files for the six untyped modules in providers/harvest/ and fixes three type issues in the existing cacheBasedCrawler.d.ts.

The `CachingService` interface was a duplicate of `ICache` from providers/caching, so it's gone. `HarvestEntry` was missing `tool` and `policy` fields that both crawlers actually use. `HarvestCallItem` had a `[key: string]: any` index signature instead of real properties.

### DIFF
--- a/providers/harvest/crawlerQueueConfig.d.ts
+++ b/providers/harvest/crawlerQueueConfig.d.ts
@@ -1,16 +1,12 @@
 // Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
 // SPDX-License-Identifier: MIT
 
-import { IQueue } from '../queueing'
 import { ICache } from '../caching'
 import { CacheBasedHarvester } from './cacheBasedCrawler'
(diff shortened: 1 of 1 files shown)


providers/harvest/crawlerQueueConfig.d.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+
+import { IQueue } from '../queueing'
+import { ICache } from '../caching'
+import { CacheBasedHarvester } from './cacheBasedCrawler'
+
+/** Configuration options for the queue-based crawler service factory */
+export interface CrawlerQueueConfigOptions {
```
qtomlinson: interface CrawlerQueueConfigOptions extends CrawlerQueueOptions? Then later and normal does not need to be declared again here.

PR #1485, opened 2026-03-11
Add type definitions for providers/curation

JamieMagee:
Adds `.d.ts` files for all 7 modules in `providers/curation/` plus a type stub for `geit` (which ships no types). A new `index.d.ts` holds the shared interfaces (`ICurationStore`, `GitHubCurationOptions`, `GitHubPR`, `CurationListResult`, etc.) so both store implementations and `GitHubCurationService` can reference them.

Tried to avoid `any` where possible. The main exceptions are old Octokit v16 methods (`hasNextPage`, `getNextPage`) that don't exist on current types, `catch` blocks (TypeScript forces `unknown`), and `yaml.load` returns.

No behavior changes.

### DIFF
--- a/providers/curation/github.js
+++ b/providers/curation/github.js
@@ -12,6 +12,7 @@
 /** @typedef {import('.').CurationDefinitionService} CurationDefinitionService */
 /** @typedef {import('.').CurationHarvestStore} CurationHarvestStore */
 /** @typedef {import('.').GitHubPR} GitHubPR */
+/** @typedef {import('./github').GitHubUserInfo} GitHubUserInfo */
 /** @typedef {import('.').ContributionInfo} ContributionInfo */
 /** @typedef {import('.').CurationPatch} CurationPatch */
 /** @typedef {import('.').CurationPatchEntry} CurationPatchEntry */
@@ -353,7 +354,7 @@ class GitHubCurationService {
   /**
    * @param {GitHubClient | null} userGithub
    * @param {GitHubClient} serviceGithub
-   * @param {{ login?: string, name?: string, email?: string }} info
+   * @param {GitHubUserInfo} info
    * @param {CurationPatchEntry} patch
    * @param {string} branch
    */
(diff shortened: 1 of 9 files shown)


providers/curation/githubConfig.d.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @param cache - Cache instance
+ * @param harvestStore - Harvest store for license matching
+ * @returns A configured GitHubCurationService instance
+ */
+declare function github(
+  options: GitHubCurationOptions | null | undefined,
+  store: ICurationStore,
+  endpoints: Endpoints,
+  cache: ICache,
+  harvestStore: {
```
qtomlinson: what about `harvestStore: CurationHarvestStore` ?

providers/curation/github.js
```diff
@@ -530,6 +621,12 @@ class GitHubCurationService {
     return curations.map(c => c.data)
   }
 
+  /**
+   * @param {GitHubClient | null} userGithub
+   * @param {GitHubClient} serviceGithub
+   * @param {{ login?: string, name?: string, email?: string }} info
```
qtomlinson: similar to GitHubUserInfo?

JamieMagee: Same fix as above.

providers/curation/index.d.ts
```diff
@@ -0,0 +1,194 @@
...
+  body: string
+  created_at: string
+  updated_at: string
+  closed_at: string | null
+  merged_at: string | null
+  merge_commit_sha: string | null
+  user: { login: string }
+  head: { sha: string; repo: { id?: number } }
+  base: { sha: string; repo: { id?: number } }
+}
```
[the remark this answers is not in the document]
JamieMagee: Done. `ContributionPR` now extends `GitHubPR` (omitting `html_url` and `head`/`base`, which it redefines with slightly different repo optionality).

providers/curation/githubConfig.d.ts
JamieMagee: Good call, swapped it in.

providers/curation/github.js
```diff
@@ -288,6 +350,13 @@ class GitHubCurationService {
     return yaml.dump(result, { sortKeys: true, lineWidth: 150 })
   }
 
+  /**
+   * @param {GitHubClient | null} userGithub
+   * @param {GitHubClient} serviceGithub
+   * @param {{ login?: string, name?: string, email?: string }} info
```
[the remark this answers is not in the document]
JamieMagee: Yep, switched both `_writePatch` and `_addOrUpdate` to use `GitHubUserInfo`. Made its properties optional since the callers pass `ContributionInfo` where those fields are optional.

PR #1488, opened 2026-03-12
Type the remaining uncovered JS files

JamieMagee:
Every source JS file now has a `.d.ts` and is in `tsconfig.json`. Four were missing: `app.js`, `bin/config.js`, `schemas/validator.js`, and `providers/index.js` (had a `.d.ts` but the `.js` was never added to tsconfig).

New `.d.ts` files for `app.js`, `bin/config.js`, and `schemas/validator.js`. Type stubs for `request-id` and `express-routes-versioning` since neither has `@types/` packages.

`bin/config.d.ts` is the most involved one. The factory-loaded config is typed with the specific interfaces each consumer expects (`HarvestStore`, `IQueue`, `CurationService`, etc.) so `app.js` gets real type checking on all the service wiring, not just `any` everywhere.

Also fixes two tsconfig bugs: `providers/index.js` was missing from the include list, and `middleware/querystring.js` was listed but the file does not exist.

Fixes a wrong return type on `AggregationService.process` in `definitionService.d.ts` — it was typed as returning a `Promise` but the method is synchronous.

JSDoc annotations in `app.js` and `bin/config.js` cover parameter types, the error handler, and the dynamic `app.init` assignment. None of the new or changed code uses `any`, `@ts-ignore`, or `@ts-expect-error`.

`tsc` passes clean. Tests unaffected (same pre-existing fetch-related failures).


JamieMagee (commented):
Deleted `querystring.d.ts` and removed it from tsconfig in 3e7abe1. The middleware was removed in #1388.

`initializers` and `Initializable` are different things. `initializers` is an array of thunks that call `.initialize()` on providers. `Initializable` is the interface those providers implement.

On `ProviderFactory<T>` reuse: `ProviderFactory` takes a single optional arg and returns `any`. The specific signatures here avoid the `any` and handle multi-arg factories like `curation.service(options, store, endpoints, cache, harvestStore)`. Worth consolidating if we clean up `ProviderFactory` itself later.

PR #1495, opened 2026-03-17
Support delaying definition recompute to the background

qtomlinson:
Based on Application Insights logging, "definition not available" events occur at 7K–10K per hour (data window: 12–18 hours). These definitions are currently computed on demand, increasing response time.

This PR adds a delayed compute path: when a definition is missing from the blob store, the service immediately returns an empty placeholder (a schema-valid definition with no tools), queues a background compute job, and triggers harvest if needed.

`UpgradeHandler` is extended into `RecomputeHandler` using a policy-composed architecture with two independent paths:

- **Upgrade path** (`upgradePolicy`): evaluates existing definitions against schema rules and queues upgrade work when stale. Unchanged from existing `DefinitionVersionCheck` / `DefinitionQueueUpgrader` behavior. Queue errors are swallowed — returning a stale definition that will be retried on the next request is acceptable.  This is the existing upgrade workflow.  
- **Compute path** (`computePolicy`): handles requests where no stored definition exists(`MissingDefinitionComputePolicy`)
  - `OnDemandComputePolicy` — computes synchronously (existing behavior, default).
  - `DelayedComputePolicy` — returns a placeholder immediately and enqueues background compute. Queue errors propagate to the caller — a missing definition with no queued work is worse than a failed request.

    Delayed queue processing uses missing-only semantics (`SkipUpgradePolicy`) so the background processor backfills missing definitions only, not stale ones. A shared in-memory cache is passed to both processing paths to prevent duplicate concurrent compute.

The provider is selected via `DEFINITION_UPGRADE_PROVIDER`: `onDemand` (default) or `delayed`. Backward-compatible aliases `versionCheck` and `upgradeQueue` are preserved.

Class names and environment variables related to upgrade are minimally changed and can be updated in a follow-up if needed.

### Test Cases

In addition to the unit tests and integration tests added, the following end to end test cases were completed.

| Step | Test                                                                    | Result |
|------|-------------------------------------------------------------------------|--------|
|      | **Pass 0: Default file-based dev setup, on-demand compute (Regression)**  |        |
| 0.1  | Default dev: full harvest + recompute (no harvest data)                 | PASS   |
| 0.2  | Default dev: fast recompute (harvest data exists)                       | PASS   |
| 0.3  | Default dev: existing definition returned as-is                         | PASS   |
| 0.4  | Default dev: force recompute works                                      | PASS   |
| 0.5  | Default dev: stale schema triggers blocking recompute                   | PASS   |
|      | **Pass 1: Azurite on-demand compute (Regression)**                        |        |
| 1.1  | On-demand: full harvest + recompute                           | PASS   |
| 1.2  | On-demand: fast recompute (harvest exists)                    | PASS   |
| 1.3  | On-demand: force recompute works                              | PASS   |
|      | **Pass 2: Azurite upgrade-queue mode (Regression)**                       |        |
| 2.1  | Queue mode: current definition returned as-is                           | PASS   |
| 2.2  | Queue mode: stale definition triggers upgrade queue                     | PASS   |
| 2.3  | Queue mode: upgrade processor recomputes                                | PASS   |
| 2.4  | Queue mode: stale def + queue unavailable → stale returned (not error)  | PASS   |
|      | **Pass 3: Azurite delayed compute path (new)**                           |        |
| 3.1  | Delayed: missing definition returns placeholder + message queued        | PASS   |
| 3.2  | Delayed: background processor computes real definition                  | PASS   |
| 3.3  | Delayed: full harvest + recompute via background                        | PASS   |
| 3.4  | Delayed: existing definition not re-queued                              | PASS   |
| 3.5  | Delayed: force=true bypasses delayed path                               | PASS   |
| 3.6  | Delayed: stale → upgrade queue, not compute queue                       | PASS   |
|      | **Pass 4: Operational resilience (Azurite, new)**                         |        |
| 4.1  | Operational: queue unavailable → request fails                          | PASS   |
| 4.2  | Operational: restart - queued message not lost                          | PASS   |
| 4.3  | Operational: duplicate requests - one compute runs                      | PASS   |


business/definitionService.js
```diff
@@ -122,11 +132,13 @@ class DefinitionService {
...
-    let result = await this.upgradeHandler.validate(existing)
+    let result = await this.recomputeHandler.validate(existing)
     if (result) {
       this._logDefinitionStatus(result, coordinates)
+    } else if (force) {
+      result = await this.computeAndStore(coordinates)
     } else {
-      result = force ? await this.computeAndStore(coordinates) : await this.computeStoreAndCurate(coordinates)
+      result = await this.recomputeHandler.compute(this, coordinates)
     }
```
JamieMagee: When `force=true`, this calls `computeAndStore`. When `force=false` and nothing's stored, it calls `recomputeHandler.compute`, which (in the on-demand case) calls `computeStoreAndCurate`. The difference is curations, `computeAndStore` skips them.

That was already the behavior before this PR, so it's not a regression. But now that the two paths are more visible side by side, I wanted to flag it. Is skipping curations on force-recompute intentional? If so, a one-line comment here explaining why would save future readers some head-scratching.

providers/upgrade/recomputeHandler.js
```diff
@@ -0,0 +1,140 @@
...
+  async setupProcessing(definitionService, logger, once) {
+    this._logger.debug('Setting up recompute handler processing', {
+      once: !!once,
+      upgradePolicy: getPolicyName(this._upgradePolicy),
+      computePolicy: getPolicyName(this._computePolicy)
+    })
+    await Promise.all([
+      this._upgradePolicy.setupProcessing?.(definitionService, logger, once, this._sharedCache),
+      this._computePolicy.setupProcessing?.(definitionService, logger, once, this._sharedCache)
+    ])
```
JamieMagee: `setupProcessing` awaits `Promise.all([upgradePolicy.setupProcessing(...), computePolicy.setupProcessing(...)])`. Both of these (when `once=false`) kick off a `setTimeout` polling loop and then resolve. So `setupProcessing` resolves after one batch from each queue, and the consumers keep running in the background independently.

That works, but it's a bit surprising. The resolved promise doesn't mean "processing is done," it means "processing has started."

tsconfig.json
```diff
@@ -180,6 +180,15 @@
...
     "providers/upgrade/defUpgradeQueueConfig.js",
+    "providers/upgrade/computePolicy.d.ts",
+    "providers/upgrade/skipUpgradePolicy.d.ts",
+    "providers/upgrade/skipUpgradePolicy.js",
+    "providers/upgrade/onDemandComputePolicy.d.ts",
+    "providers/upgrade/onDemandComputePolicy.js",
+    "providers/upgrade/delayedComputePolicy.d.ts",
+    "providers/upgrade/delayedComputePolicy.js",
+    "providers/upgrade/recomputeHandler.d.ts",
+    "providers/upgrade/recomputeHandler.js",
```
JamieMagee: :heart: Thank you!

providers/index.d.ts
```diff
@@ -44,16 +44,24 @@ export interface AuthProviders {
...
 }
 
 /** Provider configuration for upgrade service */
 export interface UpgradeServiceProviders {
-  versionCheck: ProviderFactory
-  upgradeQueue: ProviderFactory
+  onDemand: ProviderFactory
+  versionCheck: ProviderFactory //deprecated, to be removed in favor of onDemand
+  delayed: ProviderFactory
+  upgradeQueue: ProviderFactory //deprecated, to be removed in favor of delayed
```
JamieMagee: Can you use the `@deprecated` JSDoc here, and maybe mark them as `TODO` so they're easier to find?

JamieMagee (changes requested):
The upgrade/compute policy split is the right call here. I like that `defaultFactory` reproduces existing behavior without touching any of the new code paths, so `delayed` is purely opt-in.

I'd want the unbounded `while(lock)` spin in `process.js` addressed before merge. If a compute hangs, that loop stalls the queue processor forever. Also, the delayed path enqueues a new message for every request hitting the same missing coordinate. The consumer deduplicates, but at 7K-10K req/hr you're generating a lot of queue trash that gets dequeued, locked, checked, and thrown away. Even a quick `computeLock.get()` check before enqueuing would help.

providers/upgrade/process.js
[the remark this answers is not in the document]
qtomlinson: The cache ttl for the _upgradeLock is currently set to be 5 minutes, which sets the upper bound for the wait for definition compute.  This avoids a stuck compute from stalling the queue indefinitely.

JamieMagee (approved):

providers/upgrade/process.js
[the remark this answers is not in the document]
qtomlinson: one-shot spin-wait warning log after waiting >60s to surface hot-spot coordinates is added in https://github.com/clearlydefined/service/pull/1534

providers/upgrade/azureQueueConfig.js
```diff
@@ -16,13 +16,36 @@ const defaultOptions = {
...
+const computeDefaultOptions = {
+  connectionString:
+    config.get('DEFINITION_COMPUTE_QUEUE_CONNECTION_STRING') ||
+    config.get('DEFINITION_UPGRADE_QUEUE_CONNECTION_STRING') ||
+    config.get('HARVEST_AZBLOB_CONNECTION_STRING'),
+  queueName: config.get('DEFINITION_COMPUTE_QUEUE_NAME') || 'recompute',
+  dequeueOptions: {
+    numOfMessages:
+      config.get('DEFINITION_COMPUTE_DEQUEUE_BATCH_SIZE') || config.get('DEFINITION_UPGRADE_DEQUEUE_BATCH_SIZE') || 16,
+    visibilityTimeout: 10 * 60 // 10 min. The default value is 30 seconds.
```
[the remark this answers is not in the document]
qtomlinson: Aligning harvest queue visibility time out is done in https://github.com/clearlydefined/service/pull/1531

PR #1536, opened 2026-04-08
Convert lib/mockInsights to TypeScript

JamieMagee:
Merge `mockInsights.js` + `mockInsights.d.ts` into `mockInsights.ts`. First `.ts` source file.

Node 24 runs it via type stripping, no build. Net -96 lines.

Easier to review commit-by-commit: the first commit is just the rename, the second has the type changes.


qtomlinson (approved):
may want to update tsconfig.json include globs don't cover .ts files.

PR #1538, opened 2026-04-09
Convert lib/ to TypeScript

JamieMagee:
Merge all 15 `lib/*.js` + `lib/*.d.ts` pairs into `lib/*.ts` files.

Review commit-by-commit:
1. Pure renames (100% similarity, no content changes)
2. Import path updates (.js → .ts across 147 files)
3. Type merges + .d.ts deletions

One gotcha worth noting: TypeScript class field declarations (`revision?: string`) create properties with value `undefined` on every instance, even when never assigned. The old JS only set properties in the constructor, so unset fields didn't exist. This broke JSON schema validation for `EntityCoordinates` (the validator rejected the extra `revision` key). Fixed with `declare` fields, which tell TypeScript the shape without emitting runtime code.


lib/coordinatesMapper.ts
```diff
@@ -0,0 +1,86 @@
...
+    this.cache = cache
+  }
+
+  /**
+   * Maps coordinates using the appropriate provider-specific mapper.
+   */
+  async map(
+    coordinates: EntityCoordinatesSpec | undefined
+  ): Promise<EntityCoordinates | EntityCoordinatesSpec | undefined> {
+    if (!coordinates) {
```
qtomlinson, line 64: might be simpler to use: 
```
async map(
    coordinates: EntityCoordinates | undefined
  ): Promise<EntityCoordinates | undefined>
```

Then the cast can be removed at toNormalizedEntityCoordinates

lib/coordinatesMapper.ts
```diff
@@ -0,0 +1,86 @@
...
+      return undefined
+    }
+    // biome-ignore lint/style/noNonNullAssertion: Existing codebase pattern for accessing provider
+    // biome-ignore lint/suspicious/noNonNullAssertedOptionalChain: Existing codebase pattern for accessing provider
+    const mapper = this.mappers[coordinates?.provider!]
```
qtomlinson: coordinates?.provider, ? can be dropped after falsy coordinates check at line 66.

lib/utils.ts
```diff
@@ -23,14 +18,138 @@ import extend from 'extend'
...
+    unknown: number
+  }
+  discovered?: {
+    expressions?: string[]
+    unknown: number
+  }
+}
+
+/** Full definition object */
+export interface Definition {
```
qtomlinson, line 110: A better home for these definition related types is probably in the business layer.  Something to consider for future refactoring effort.

qtomlinson (approved):

lib/utils.ts
```diff
@@ -375,68 +442,53 @@ function _mergeFiles(base, proposed, override) {
...
-      )
+  const result = _mergeExcept(base, proposed, ['license', 'attributions', 'facets', 'hashes', 'natures']) as FileEntry
+  const overrideStrategy: <T>(proposed: T, mergeStrategy?: (p: T) => T) => T = override
+    ? <T>(proposed: T): T => proposed
+    : <T>(proposed: T, mergeStrategy?: (p: T) => T): T => mergeStrategy!(proposed)
   setIfValue(
     result,
     'license',
-    overrideStrategy(proposed.license, /** @param {string | undefined} p */ p => SPDX.merge(p, base.license, 'AND'))
+    overrideStrategy(proposed.license, (p: string | undefined) => SPDX.merge(p!, base.license!, 'AND'))
```
qtomlinson: SPDX.merge can handle falsy values, is this assert necessary? This is used in merging tool results.  It is possible one tool detected a license for a file while another tool did not detect a license for the same file.

lib/utils.ts
```diff
@@ -375,68 +442,53 @@ function _mergeFiles(base, proposed, override) {
...
-  /** @type {<T>(proposed: T, mergeStrategy?: (p: T) => T) => T} */
-  const overrideStrategy = override
-    ? /** @type {<T>(proposed: T) => T} */ (proposed => proposed)
-    : /** @type {<T>(proposed: T, mergeStrategy: (p: T) => T) => T} */ (
-        (proposed, mergeStrategy) => mergeStrategy(proposed)
-      )
+  const result = _mergeExcept(base, proposed, ['license', 'attributions', 'facets', 'hashes', 'natures']) as FileEntry
+  const overrideStrategy: <T>(proposed: T, mergeStrategy?: (p: T) => T) => T = override
+    ? <T>(proposed: T): T => proposed
+    : <T>(proposed: T, mergeStrategy?: (p: T) => T): T => mergeStrategy!(proposed)
```
qtomlinson, line 459: considering mergeStrategy!, should we make mergeStrategy not optional?

PR #1540, opened 2026-04-10
Convert schemas/, providers/logging/, providers/caching/ to TypeScript

JamieMagee:
Merge 6 `.js` + `.d.ts` pairs into `.ts`: `schemas/validator`, `providers/logging/logger`, `providers/logging/winstonConfig`, `providers/caching/memory`, `providers/caching/redis`, `providers/caching/redisConfig`.


providers/caching/memory.ts
```diff
@@ -0,0 +1,46 @@
...
+  async done(): Promise<void> {}
+
+  /** Retrieves an item from the cache */
+  get(item: string): any {
+    return this.cache.get(item)
+  }
+
+  /** Stores an item in the cache */
+  set(item: string, value: any, ttlSeconds: number | null = null): void {
+    const expiration = 1000 * (ttlSeconds || this.defaultTtlSeconds || 0)
```
qtomlinson: `memory-cache`'s `put()` throws `Error: Cache timeout must be a positive number` if `time` is `0` or negative. The factory always passes `{ defaultTtlSeconds: 60 * 60 }` today, so this won't be hit in practice. `memory-cache` already handles "no expiry" natively by passing `undefined`. Suggest:

```ts
set(item: string, value: any, ttlSeconds: number | null = null): void {
  const expiration = ttlSeconds || this.defaultTtlSeconds
  this.cache.put(item, value, expiration ? expiration * 1000 : undefined)
}
```

schemas/validator.ts
```diff
@@ -21,23 +21,23 @@ const versionlessCoordinatesSchema = require('./versionless-coordinates-1.0.json
...
-const ajv = new Ajv(ajvOptions)
+const ajv = new Ajv.default(ajvOptions)
 
 // Add formats and error messages support
-addFormats(ajv)
+;(addFormats as unknown as (ajv: InstanceType<typeof Ajv.default>) => void)(ajv)
 if (allErrorsEnabled) {
-  ajvErrors(ajv)
+  ;(ajvErrors as unknown as (ajv: InstanceType<typeof Ajv.default>) => void)(ajv)
 }
```
qtomlinson, line 43: Consider extracting a type alias to reduce repetition and improve readability:

```ts
type AjvPlugin = (ajv: InstanceType<typeof Ajv.default>) => void

;(addFormats as unknown as AjvPlugin)(ajv)
if (allErrorsEnabled) {
  ;(ajvErrors as unknown as AjvPlugin)(ajv)
}
```

qtomlinson (approved):

providers/caching/memory.ts
JamieMagee: Done, used your suggestion.

schemas/validator.ts
JamieMagee: Can't use `as unknown as` here — `as` casts aren't erasable syntax, so Node's type stripping leaves a broken expression at runtime. Used `@ts-expect-error` instead. Not pretty, but it's the only option that works without a build step. The ajv-formats and ajv-errors packages are callable at runtime but their type definitions don't declare a default call signature.

PR #1541, opened 2026-04-10
Convert providers/queueing/, providers/search/ to TypeScript

JamieMagee:
Merge 6 `.js` + `.d.ts` pairs into `.ts`: `providers/queueing/azureStorageQueue`, `providers/queueing/memoryQueue`, `providers/search/abstractSearch`, `providers/search/azureConfig`, `providers/search/azureSearch`, `providers/search/memory`.

Stacked on #1540.


providers/queueing/memoryQueue.ts
```diff
@@ -1,32 +1,36 @@
...
 
-class MemoryQueue {
-  /**
-   * @param {import('./memoryQueue').MemoryQueueOptions} options
-   */
-  constructor(options) {
+class MemoryQueue implements IQueue {
+  declare options: MemoryQueueOptions
+  declare logger: Logger
+  declare decoder: ((text: string) => string) | undefined
```
qtomlinson: The type of decoder is correct, which means we should be either checking for null or assert in _parseData().  
Another option is to move the default decoder from the factory to the constructor (as a fallback) and then we can remove the `undefined` here in typing and the factory becomes a thin wrapper.

qtomlinson (approved):

PR #1548, opened 2026-04-11
Convert providers/harvest/ to TypeScript

JamieMagee:
Merge 9 `.js` + `.d.ts` pairs into `.ts` in `providers/harvest/` (incl. throttling).


qtomlinson (commented):
**Follow-up suggestion: tighten `HarvestEntry.coordinates` back to `EntityCoordinates`**

The widening to `EntityCoordinates | string` is currently needed to accommodate `routes/harvest.js:139`:

```js
return { ...entry, coordinates: normalized.toString() }
```

As a follow-up, consider passing the object directly instead:

```js
return { ...entry, coordinates: normalized }
```

This would allow `HarvestEntry.coordinates` to be kept as `EntityCoordinates` only. The `POST /harvest` route and its unit tests would need to be reviewed and adjusted accordingly.

qtomlinson (approved):

PR #1554, opened 2026-04-14
Finish TypeScript migration, enable verbatimModuleSyntax

JamieMagee:
Converts the last .js files: complex routes (auth, curations, definitions, harvest, webhook), all 7 business services, app.ts, and bin/config.ts. After that, enables verbatimModuleSyntax, standardizes .d.ts import extensions, drops proxyquire (tests already use esmock), and fixes 25 pre-existing test build errors.

13 commits following the same rename/imports/types pattern. Review commit by commit.


qtomlinson:
**`business/aggregator.ts:63`**

Minor non-blocking suggestion: consider adding `implements IAggregationService` to the class declaration to make the contract with `definitionService.ts` explicit at compile time. An alias avoids the name clash:

```ts
import type { AggregationService as IAggregationService } from './definitionService.ts'

class AggregationService implements IAggregationService {
```

Consistent with the pattern used elsewhere in this migration (e.g., `OnDemandComputePolicy implements MissingDefinitionComputePolicy`).

---

**`routes/definitions.ts:63`**

Minor non-blocking note: `!!request.query.force` treats `?force=false` as `true` since `'false'` is a truthy string. This is a pre-existing bug (the original JS passed the raw string, with the same effect), but the migration is a nice opportunity to align with `definitions-1.0.0.ts` and use `=== 'true'` instead.

qtomlinson (approved):

PR #1582, opened 2026-05-13
fix(logging): prevent verbose/debug logs from leaking to Application Insights

yashkohli88:
## Summary
- Custom `ApplicationInsightsTransport` replaces Winston `data` event listener
- Transport respects logger level config — verbose/debug blocked when `level: 'info'`
- Errors with stack → `trackException`, without → `trackTrace` with Error severity
- Console format includes timestamp, AI transport doesn't duplicate it

## Problem
Logger forwarding verbose/debug events to Application Insights even when configured as `info`:
- **Configured level:** `info`
- **Unexpected behavior:** verbose/debug entries still appear in telemetry
- **Result:** telemetry volume increases, signal-to-noise degrades

## Why
Winston `logger.on('data')` fired for ALL logs regardless of configured level. Debug/verbose leaked to Application Insights even when logger set to info/warn.

## Related
Similar fix for crawler:
- Issue: https://github.com/clearlydefined/crawler/issues/739
- PR: https://github.com/clearlydefined/crawler/pull/740

## Test coverage
- Verifies debug/verbose blocked at info level
- Confirms error routing (stack → exception, plain → trace)
- Validates severity mapping (warn → Warning)


providers/logging/winstonConfig.ts
```diff
@@ -103,6 +105,47 @@ function buildProperties(info: Record<string, any>): Record<string, any> {
...
+      }
+    } else {
+      this.aiClient.trackTrace({
+        message: info.message,
+        severity: mapLevel(info.level),
+        properties
+      })
+    }
+
+    callback()
```
JamieMagee: If anything in this block throws, `callback()` never runs. That can leave Winston waiting on the transport for a log write that already failed. Can we wrap the AI tracking work in `try/finally` and call `callback()` from `finally`?

JamieMagee (approved):

